PR #11, opened 2016-10-26
download all templates on package install to prevent issues in multip…

rwblair:
…roc environments of multiple calls to get template functions cloberring each other


chrisfilo:
+1

PR #13, opened 2016-11-15
move reports infrastructure to niworkflows

shoshber:


shoshber:
Thanks! I've updated the pr. I don't have write permissions for niworkflows though, so I can't merge

PR #15, opened 2016-11-17
[ENH] Add the new automated reporting

oesteban:
to the spatial normalization interface

example: http://htmlpreview.github.io/?https://raw.githubusercontent.com/poldracklab/niworkflows/enh/RegistrationReports/examples/viz-report.html


niworkflows/anat/mni.py
```diff
@@ -107,26 +116,35 @@ def _run_interface(self, runtime):
                 self._results.update(interface_result.outputs.get())
                 NIWORKFLOWS_LOG.info(
                     'Successful spatial normalization (retry #%d).', self.retry)
+
+                if self.inputs.generate_report:
+                    try:
+                        self._generate_report()
+                    except:
+                        pass
+
```
shoshber: what is this for? Subclasses of ReportCapableInterface call self._generate_report() automatically (unless I'm missing a code path here?)

niworkflows/anat/mni.py
```diff
@@ -138,8 +156,43 @@ def _config_ants(self, ants_settings):
...
-        norm.inputs.fixed_image_mask = op.join(mni_template,
-                                               '%dmm_brainmask.nii.gz' % resolution)
-        return norm
+        self.norm.inputs.fixed_image_mask = op.join(
+            mni_template, '%dmm_brainmask.nii.gz' % resolution)
+
+
+    def _generate_report(self):
+        """ Generates the visual report """
+        NIWORKFLOWS_LOG.debug('Generating visual report')
```
shoshber: 👍 to the logging

niworkflows/viz/overlay_3d_report.tpl
```diff
@@ -1,34 +1,29 @@
...
+    <h4>{{title}}</h4>
     <div class='image_container'>
         <div class='base_image'>{{base_image}}</div>
         <div class='overlay_image'>{{overlay_image}}</div>
     </div>
-    <h5>Inputs</h5>
-    <p><pre>{{inputs}}</pre></p>
-    <h5>Outputs</h5>
-    <p><pre>{{outputs}}</pre></p>
-    <p> Generated with overlay_3d_report.tpl </p>
```
shoshber: I added these to aid us while we debug. Can we leave this for now?

niworkflows/anat/mni.py
```diff
@@ -138,8 +156,43 @@ def _config_ants(self, ants_settings):
...
+                svgs.append('<div id="{0}-{1}" class="{0}">{2}</div>'.format(
+                    div_id, mode, report.as_svg(image)))
+                image.close()
+            return '\n'.join(svgs)
+
+        plot_params = {'cut_coords': 3}
+        report.save_html(
+            template='overlay_3d_report.tpl',
+            report_file_name=self._results['html_report'],
+            unique_string='antsRegistration',
```
shoshber: `unique_string` needs to be even more unique, in case this interface runs multiple times in a workflow. It doesn't matter right now, but it will head off future problems when we concatenate the reports. I would suggest `'ants' + uuid.uuid4()` (must start with alphabetic chars, bc it will be a div id).

Speaking of div_id, I don't see the motivation for the `div_id` variable in `_plot_xyz`. It doesn't seem to add any descriptive power to the final html (except for specifying the direction of the cuts, which since we have no guarantee as to what x, y, and z represent irl, doesn't seem useful). Also, because we will be concatenating this report with other reports from unknown sources, we can't specify any id's unless they contain a unique string.

I should add an html validator to save_html that enforces these rules so they aren't just in my head.

niworkflows/anat/mni.py
```diff
@@ -8,15 +8,20 @@
...
 from nipype.interfaces.base import (traits, isdefined, BaseInterface, BaseInterfaceInputSpec,
                                     File, InputMultiPath)
 
 from niworkflows.data import getters
+from niworkflows.common import report
 from niworkflows import __packagename__, NIWORKFLOWS_LOG
 
-MAX_RETRIES = 3
-
+DEFAULT_CUTS = {
```
oesteban: This is good for the MNI atlas. It is an atlas precisely because we accurately know where is everything _a priory_.

I agree with you that these defaults would not work, for instance, for showing the result of a segmentation on a brain, since you don't know the coordinates for that case.

niworkflows/anat/mni.py
oesteban: unique_string: Yeah, I was reluctant to add another dependency to the code. I would prefer to pick up the identifier of the nipype interface or so.

div_id for plots: this is an attempt to be able to easily select subplots with css. When you integrate the code in a larger html file, things don't work as expected (not your fault). In the long term, my idea was (still is) to have only one svg file, that composes the 3 panels of the background and the 3 panels of the foreground. Since svg is html compatible, the css tricks work inside this unique svg with the 6 panels.

So this divs are mocking the inner svg elements of that future panel.

Does it make sense?

niworkflows/anat/mni.py
shoshber: What about the difference between MNI 2mm and MNI 1mm?

niworkflows/anat/mni.py
```diff
@@ -41,19 +46,28 @@ class RobustMNINormalizationInputSpec(BaseInterfaceInputSpec):
     settings = traits.List(File(exists=True), desc='pass on the list of settings files')
     template_resolution = traits.Enum(1, 2, mandatory=True, usedefault=True,
                                       desc='template resolution')
+    generate_report = traits.Bool(
+        False, usedefault=True, desc='enable report generation')
+    html_report = File('viz-report.html', usedefault=True, desc='visual report file')
```
shoshber: (update: @rwblair volunteered to do this instead)

niworkflows/viz/overlay_3d_report.tpl
oesteban: Yeah, I will generate another tpl file for this interface. I am testing this in mriqc, so I don't need (want) that information.

niworkflows/anat/mni.py
shoshber: Ah, I see. And I presume we are confident that the reference image will almost always be MNI?

niworkflows/anat/mni.py
shoshber: > unique_string: Yeah, I was reluctant to add another dependency to the code. I would prefer to pick up the identifier of the nipype interface or so.

I had thought about this, but I don't think the unique node name is accessible from inside `_generate_reports()`. Can you think of a way to get it? uuid is in python's standard library and it's not a very big dependency so I think it's reasonable to use it.

re: ids, check out my pr here: https://github.com/poldracklab/niworkflows/pull/19

oesteban: I don't remember on top of my head if it is easy to get a hash of the interface from itself. Anyways, I think you are right, uuid is part of the standard library. So I finally adopted it: https://github.com/poldracklab/niworkflows/blob/85ce576d8f01e6ca9e204e4af6224147b5707809/niworkflows/anat/mni.py#L209.

That said, with the standalone svgs I got it working with duplicated IDs, no big deal.

niworkflows/anat/mni.py
oesteban: I'm doing this in #22

PR #19, opened 2016-11-17
ENH More valid HTML

shoshber:
Hardened against concatenation/non-uniqueness.


oesteban:
I think checking the HTML is great, but would not freak out about it. We are worried about correct HTML as long as the browsers render our reports correctly.

Therefore, I would merge this as soon as possible, and put on hold the HTML checking. I'm sure Zack will give us great feedback. In the end, we want these reports to render well in two use cases:
  1. standalone - you open the report files individually, for consumers using local installs of fmriprep. Here we can even recommend a browser and forget of everything else.
  2. inside the platform - we'll need Zack's input, and we may start seeing conflicts and worry about browsers. For the conflicts, I think here it is legit to use an ```iframe```. Regarding browsers, let's push this forward to the time we open the platform.

shoshber:
I disagree--the stuff that the HTML is validating is a central part of the current design for the reports generation. I believe we discussed this last week with all four of us in @rwblair 's and my office. The HTML checking will not prevent consumers from opening the reports as stand-alone files.

I believe--though I'm not 100% sure--that using `iframe` will require the full report to consist of multiple html files. A large design consideration is to make the full report a single html file @chrisfilo .

chrisfilo:
This looks like a misunderstanding. This validator checks for id conflicts to avoid problems when multiple reportles will be merged into a single HTML file (if I understand this correctly).

Seems like a useful feature, so lets finish this up (tests are failing), merge it and move on to finishing up BET, FLIRT, and ANTs reportles as well as implementing the report aggregator.

shoshber:
I just need to write CSS validation and refactor a bit, then this is done. Shouldn't take more than two days.

niworkflows/common/report.py
```diff
@@ -2,9 +2,12 @@
 
 from __future__ import absolute_import, division, print_function
 from io import open
-
+import uuid
```
chrisfilo: Where is this import used?

shoshber:
xml

shoshber:
Based on discussion, I reverted the last few commits for CSS validation. If tests pass this is ready to merge.

chrisfilo:
Please don't merge it until you make sure that all of the report capable interfaces pass this validation. Fix them if necessary.

I want to avoid the situation when FMRIPREP will stop working due to validation errors after merging this PR.

shoshber:
@oesteban please write tests for the interfaces you are responsible for so Chris's concern can be more easily met, now and in the future.

PR #20, opened 2016-11-18
[WIP] All interfaces from fmriprep

rwblair:
Also made base inputspec for all reportcapable interfaces to use. Working on getting tests to past still.

Putting this PR out there so others can see what changes are being made,


niworkflows/__init__.py
```diff
@@ -19,3 +20,5 @@
 
 NIWORKFLOWS_LOG = logging.getLogger(__packagename__)
 NIWORKFLOWS_LOG.setLevel(logging.INFO)
+
+matplotlib.use('Agg')
```
shoshber: what is this for? Can you add a comment to explain it?

niworkflows/common/report.py
```diff
@@ -8,6 +8,13 @@
 import jinja2
 from pkg_resources import resource_filename as pkgrf
 
+from nipype.interfaces.base import File, traits
+
+class ReportCapableInputSpec(object):
+    generate_report = traits.Bool(
+        desc="Set to true to enable report generation for node"
+    )
+
```
shoshber: Oscar added these to his input spec, which I quite like:

`
generate_report = traits.Bool(
        False, usedefault=True, desc='enable report generation')
html_report = File('viz-report.html', usedefault=True, desc='visual report file')
`

What do you think about adding these to `ReportCapableInputSpec`?

For consistency's sake I would prefer to call it `ReportCapableInterfaceInputSpec`. But I'll leave it up to you since that would be a monstrously long class name.

niworkflows/tests/test_niworkflows.py
```diff
@@ -1,110 +0,0 @@
-from __future__ import absolute_import, division, print_function
```
shoshber: I will hold a funeral for the death of these poor tests :'(

niworkflows/common/report.py
rwblair: When I tried to multiclass with the ReportCapableInputSpec I got this error:
traits.trait_errors.TraitError: Cannot set the undefined 'generate_report' attribute of a 'FLIRTInputSpecRPT' object.

I couldn't figure out why the attribute wasn't being carried over into the flirt input spec so I reverted the changes.

Once this gets merged I'll open up a new PR focused on just that feature and see if the error can't be resolved there.

niworkflows/__init__.py
rwblair: This sets the matplotlib backend. Without it I was not able to use matplotlib on circleci or on my mac. The legacy image generation code from Craig actually had this sprinkled in each of the plotting functions. On circle the error message generated had to with the fact that the version of python being used did not have the _tkinter module, but I don't have the exact text of it at hand unfortunately.

PR #21, opened 2016-11-18
[ENH] Add the new automated reporting (standalone svg file)

oesteban:
This PR supersedes the previous effort to generate granular and standalone reports to check registration results, and thus close #15.


shoshber:
I left some comments and responses on PR #15 before I saw that you had opened this one--can you take a look?

niworkflows/anat/mni.py
```diff
@@ -164,35 +170,115 @@ def _config_ants(self, ants_settings):
...
+            # Add styles for the flicker animation
+            with open(out_file, 'rb') as f:
+                svg = f.read().split('\n')
+
+            # Replace header (get rid of height, width and viewBox) and add custom stylesheet
+            # svg[1] = ('<svg xmlns:xlink="http://www.w3.org/1999/xlink" '
+            #           'xmlns="http://www.w3.org/2000/svg" version="1.1" class="flickering" '
+            #           'viewBox="0 0 %0.2f %0.2f">' % tuple(totalsize))
+            svg.insert(2, """\
+  <style type="text/css">
```
shoshber: let's try to keep style tags outside of svgs for now, until we establish a way of validating that they won't mess with each other. OR if you absolutely insist on putting it in here, then use the unique_string like this: `@keyframes flickerAnimation{{unique_string}}` and I will write a CSS validator as well.

niworkflows/anat/mni.py
```diff
@@ -164,35 +170,115 @@ def _config_ants(self, ants_settings):
...
                 plot_params['display_mode'] = mode
-                plot_params['cut_coords'] = cuts
-                image = plot_anat(anat_img, **plot_params)
-                svgs.append('<div id="{0}-{1}" class="{0}">{2}</div>'.format(
-                    div_id, mode, report.as_svg(image)))
-                image.close()
-            return '\n'.join(svgs)
-
-        plot_params = {'cut_coords': 3}
-        report.save_html(
```
shoshber: why aren't you using save_html?

oesteban: Because I don't need it. I think the visual reports should be as agnostic as possible about where they will end up included. With that in mind, a standalone svg is much better than decorating it with html code.

Example use case: if I want to directly import one artifact of this reports as a figure in a latex paper. If you don't produce svg you will not be able to do this directly: you'll need to remove html decorations.

niworkflows/anat/mni.py
oesteban: Ok, I'll add the unique string. But yeah, these svgs can be opened themselves and they correctly show the flickering trick since the CSS are embeded, I don't see why the styling should go aside.

niworkflows/anat/mni.py
```diff
@@ -48,7 +54,7 @@ class RobustMNINormalizationInputSpec(BaseInterfaceInputSpec):
                                       desc='template resolution')
     generate_report = traits.Bool(
         False, usedefault=True, desc='enable report generation')
-    html_report = File('viz-report.html', usedefault=True, desc='visual report file')
+    html_report = File('viz-report.svg', usedefault=True, desc='visual report file')
```
oesteban: Nope, this is the deal: the generated svg is standalone and works fine without anything else. I comment on why this is interesting below.

BTW, I would suggest to change the ```html_report``` by ```out_report``` which is more nipype-like or ```report_file``` which is also somewhat nipype-like and does not convey that the output is html always.

niworkflows/anat/mni.py
```diff
@@ -120,8 +126,8 @@ def _run_interface(self, runtime):
                 if self.inputs.generate_report:
                     try:
                         self._generate_report()
-                    except:
-                        pass
+                    except Exception as reason:
+                        NIWORKFLOWS_LOG.warn('Report was not generated, reason: %s', reason)
```
oesteban: Will do

niworkflows/anat/mni.py
```diff
@@ -164,35 +170,115 @@ def _config_ants(self, ants_settings):
...
+                      order=('z', 'x', 'y'), cuts=None):
             """
-
+            Plots the foreground and background views
+            Default order is: axial, coronal, sagittal
             """
-            NIWORKFLOWS_LOG.info('Plotting %s', str(anat_img))
-            svgs = []
+            if plot_params is None:
+                plot_params = {}
```
oesteban: I don't have a preference. If this will make you happier, bless it ... :D

niworkflows/anat/mni.py
shoshber: I prefer the name `html_report` because it creates a strong norm for the files being html files, because we will be writing an aggregator that assumes the reports are html snippets.

oesteban: WDYT @chrisfilo ?

niworkflows/anat/mni.py
shoshber: Currently, save_html provides validation so we know that the html is valid for its currently-planned use-case: the final report aggregator. As I commented below, I'll make the validation stand-alone from `save_html` so you can use it here.

BTW human-readable reports should have some semantic information about what the image is based on, the title, the original files, etc. Are those included in the SVG? I don't see it but I'm not familiar with SVGs or the `lxml` module so I could be wrong.

oesteban:
This is soon to be superseded by #22

shoshber:
ok, that's fine. But I'm wondering why you keep making new PRs? It makes it hard to follow the continuity of the diffs.

oesteban:
I will rebase this on #22 when merged.

This one will focus on the MNI Spatial normalization. #22 is more about the system.

I can't remember why I closed #15.

shoshber:
got it

niworkflows/anat/mni.py
shoshber: @rwblair

PR #22, opened 2016-11-22
[ENH] RegistrationReportCapable interfaces

oesteban:
Creates an abstract mixin for registration interfaces, where two images
(the reference and the moving) are to be presented with the flickering
animation.

This first commit covers the following:

- Split the visualization utilities from common/report.py to viz/utils.py
- Some house-keeping on the base class:
  - Change the main report output name to ```out_report``` to be agnostic to the output format
  - Fix the super classes of the ReportCapableInput/OutputSpecs to be traited (they won't work otherwise)
  - Simplify _run_interface: avoid using a try...except to wrap the super()._run_interface (less error tolerant)
  - Add logging
  - Pass exit code to _generate_error_report


shoshber:
> Change the main report output name to out_report to be agnostic to the output format
I've already registered the reasons for `html_report` vs. a more agnostic `out_report`. Let's let others weigh in rather than making unilateral decisions. Discussion here: https://github.com/poldracklab/niworkflows/pull/21#discussion_r88761066

> Fix the super classes of the ReportCapableInput/OutputSpecs to be traited (they won't work otherwise)
@rwblair is this ok? I recall you saying that it is safest for mixins to only subclass `object`

> Simplify _run_interface: avoid using a try...except to wrap the super()._run_interface (less error tolerant)
I would hold off on this, because it was originally supposed to be overriding `Interface.run()` instead of `Interface._run_interface()`. I still think we should override `.run()`, and if we do that, a lot of the complicated error handling and `pass`ing on `NotImplementedError`, etc will go away by itself. Currently it does not appear to work when overriding `run()` but given the architecture of nipype it *should* work, so I plan to dig into this even more.

The rest looks good.

oesteban:
Ok, this is passing. There is one test on BET  I don't understand, it expects a warning but I don't see the point. The test on FLIRT is disabled until I write that wrapper.

With all of that fixed, the WIP label will go away.

Regarding the discussion of ```run()``` vs ```_run_interface()```: I agree that run may workaround the NotImplementedError. However this is what we have right now and the reporting is blocking the project. So I would address the suggestion before PR'ing these reports into nipype. For that PR everything should be well backed up.

shoshber:
re:BET have you merged in the more recent code?

oesteban:
I think I started off master

shoshber:
try merging it in, that might fix the BET test

oesteban:
I would prefer to understand what this test is looking for before trying merging anything ... (https://github.com/oesteban/niworkflows/blob/a2bd6c7ae3a079c41a2bd00eefb4ad7da90a5ef5/niworkflows/tests/test_reports.py#L63)

Is it expecting a warning when ```bet.inputs.mask == False```? What does _Can't generate a report if there are no nifti outputs_ mean? Isn't that done by the generated error report?

shoshber:
Depending on the inputs, BET nodes may not output any nifti files. I don't know what the use-case is for that (maybe dry runs?), but it is valid! If there are no nifti files, we can't generate a report (or at least, I haven't thought of an easy way). I don't want to raise an exception, since it *is* valid, but I don't want to let it pass by without comment, because they did request for a report ... therefore, warning.

BTW, I looked at the code, and the test is failing because you modified it here (https://github.com/oesteban/niworkflows/commit/bf6d693dbd42330195f4b37b438e177597b23944#diff-8ce7a459bdf6bcfce8fa5e54ca308939L46) and here (https://github.com/oesteban/niworkflows/commit/a2bd6c7ae3a079c41a2bd00eefb4ad7da90a5ef5). I think the inputs should be
`generate_report=True, outline=False, mask=False, no_output=True`

chrisfilo:
We should always force mask to True if generate_report==True. Therefore there should be no need for this [warning](https://github.com/poldracklab/niworkflows/blob/master/niworkflows/common/report_interfaces.py#L69).

I found a couple other problems with the BET interface - I list them in a separate issue.

chrisfilo:
I listed everything here: https://github.com/poldracklab/niworkflows/issues/24 Basically as soon as we force the mask file to be written down (which is perfectly sensible if the user requested a report) everything becomes much more simple.

shoshber:
Outstanding issues pre-merge:
* `out_report` / `html_report`
* ReportCapableInput/OutputSpecs subclassed from `object` / `TraitedSpec`

oesteban:
Let's discuss the outstanding issues tomorrow in the meeting.

As regards the missing parts of this PR:
- https://34-63891310-gh.circle-artifacts.com/0/home/ubuntu/scratch/testFLIRT.svg
- https://34-63891310-gh.circle-artifacts.com/0/home/ubuntu/scratch/testBET_000.svg - this one is html code (I'll fix the extension), thus it needs CSS to show the divs correctly aligned.

I'll remove the WIP tag after fixing that extension.

oesteban:
Extension fixed:
- https://40-63891310-gh.circle-artifacts.com/0/home/ubuntu/scratch/testBET_000.html
- https://40-63891310-gh.circle-artifacts.com/0/home/ubuntu/scratch/testFLIRT.svg

niworkflows/common/report.py
```diff
@@ -1,106 +1,130 @@
...
-    searchpath = pkgrf('niworkflows', '/')
-    env = jinja2.Environment(
-        loader=jinja2.FileSystemLoader(searchpath=searchpath),
-        trim_blocks=True, lstrip_blocks=True
-    )
-    report_tpl = env.get_template('viz/' + template)
-    kwargs['unique_string'] = unique_string
-    report_render = report_tpl.render(kwargs)
+class RegistrationRCInputSpec(ReportCapableInputSpec):
+    out_report = File(
```
rwblair: hashing?

PR #36, opened 2016-11-29
ENH BET and segmentation reports

shoshber:


shoshber:
Tests work locally, it is failing on Circle because this bugfix https://github.com/nipy/nipype/pull/1723 hasn't been merged in yet. I'm going to say it's ready to merge.

chrisfilo:
Could you either set basename in the tests so the tests would pass and we could see the example reports or provide them some other way?

shoshber:
I set it to run shoshber/nipype, hopefully that works

niworkflows/viz/utils.py
```diff
@@ -56,7 +59,7 @@ def as_svg(image):
 
 def cuts_from_bbox(mask_nii, cuts=3):
     from nibabel.affines import apply_affine
-    imnii = nb.load(mask_nii)
+    imnii = mask_nii
```
oesteban: ??

rwblair: A loaded image is now being passed to cuts_from_bbox Line 128:
`mask_nii = nb.load(mask_nii) if masked else nlimage.threshold_img(mask_nii, 1e-3)`

shoshber: `mask_nii`, how I ended up using it, is an already-loaded Nifti1

chrisfilo:
test are passing (yay!), but example BET and FAST reports do not seem to be included among the artifacts...

niworkflows/viz/utils.py
oesteban: I understand, but then please update the whole function to avoid that double referencing, right? 
--
Could we have a nilearn-like behavior, and make this function allow both file paths and nifti images as input?

chrisfilo:
LGTM

PR #49, opened 2016-11-30
RTM/ENH ANTS brain extraction report

shoshber:


chrisfilo:
FYI: I triggered a rebuild for the 91e99e7 commit. I don't think the `-ti` flag was the problem. Let's see what happens.

shoshber:
If I understand it properly it'll time out again. I think it's because the docker command isn't outputting anything and I added a test to run ANTS which takes longer than 10 minutes. Looking around the internet, `-ti` is for running it interactively, so I thought no harm in trying!

shoshber:
Buuut it's still not outputting anything so it didn't do anything.

shoshber:
I increased the timeout for now (in the interests of time) and made an issue for figuring it out later: https://github.com/poldracklab/niworkflows/issues/55

PR #48, opened 2016-11-30
[WIP] Fmriprep intergration fixes

rwblair:
Kicking of circle for these changes.

PR #50, opened 2016-11-30
[FIX, RTM] Integration of RobustMNINormalization

oesteban:
Replaces #32, which I mistakenly requested for merging with a branch different from master.


chrisfilo:
If I'm not mistaken FMRIPREP is using RobustMNINormalization as the main way to coregister T1s with the MNI template (https://github.com/poldracklab/fmriprep/blob/00530d6dacdf8ec01d312f72a02537762dd7b87c/fmriprep/workflows/anatomical.py#L65). I previously thought that it was using the ants.Registration interface directly for which we already have reportlets. This means we'll need this PR merged to get any reporting on T1-MNI coregistration. Am I right?

oesteban:
@chrisfilo totally right. This was actually working well before. Now this is kind of blocked, awaiting #48.

rwblair:
#48 has been merged

oesteban:
This is RTM -> https://28-74174878-gh.circle-artifacts.com/0/home/ubuntu/scratch/testRobustMNINormalizationRPT.svg

niworkflows/interfaces/registration.py
```diff
@@ -24,24 +46,26 @@ class FLIRTRPT(nrc.RegistrationRC, fsl.FLIRT):
...
+# class ANTSRegistrationOutputSpecRPT(nrc.ReportCapableOutputSpec,
+#                                     ants.registration.RegistrationOutputSpec):
+#     pass
 
-class ANTSRegistrationRPT(nrc.RegistrationRC, ants.Registration):
-    input_spec = ANTSRegistrationInputSpecRPT
-    output_spec = ANTSRegistrationOutputSpecRPT
+# class ANTSRegistrationRPT(nrc.RegistrationRC, ants.Registration):
+#     input_spec = ANTSRegistrationInputSpecRPT
+#     output_spec = ANTSRegistrationOutputSpecRPT
```
shoshber: What is the reason for commenting these out?

niworkflows/tests/test_reports.py
```diff
@@ -46,6 +43,20 @@ def test_known_file_out(self, flirt=FLIRTRPT):
...
+                generate_report=True, moving_image=moving, testing=True)
+            res = ants_rpt.run()
+            out_report = res.outputs.out_report
+
+            if os.getenv('SAVE_CIRCLE_ARTIFACTS', False) == "1":
+                copy(out_report, os.path.join('/scratch', 'testRobustMNINormalizationRPT.svg'))
+
+            self.assertTrue(os.path.isfile(out_report), 'HTML report exists at {}'
+                            .format(out_report))
+
```
shoshber: I think you can refactor this to use `_smoke_test_report`.

niworkflows/interfaces/registration.py
oesteban: That was there before but the actual ANTS interface is not yet implemented. I will address this in a PR for #54 .

PR #51, opened 2016-12-01
Fmriprep intergration fixes patch

oesteban:


niworkflows/viz/validators.py
```diff
@@ -23,18 +27,17 @@ def validate(self, css):
     def validate_no_fixed_position(self, rule):
         ''' checks counter names and position values '''
         if rule.at_keyword is not None:
-            declarations = parser.parse_declaration_list(rule.body)
-        else: # not an at-rule
+            declarations = self.parser.parse_declaration_list(rule.body)
```
oesteban: Please use some linting before merging

PR #53, opened 2016-12-01
Fmriprep intergration fixes patch 2

oesteban:


chrisfilo:
(I moved the conversation here, since it pertains to the changes in this PR)

So if I understand correctly the TestHTMLValidator was not detected, by nose and actually never run until this change? And it turns out it does not pass?

In the interest of time on other things blocked on this I would recommend skipping this tests and opening an issue about it. This can be fixed in a separate PR (I expect it will be easy for @shoshber to fix it whenever there is enough time).

shoshber:
??? I have no idea what is going on here. `test_validators` has been in master, with the TestValidators class, since this commit: https://github.com/poldracklab/niworkflows/commit/b46b3f3adae6a468009cd26f7976936581416191

shoshber:
It looks like the tests @oesteban moved into `test_validators.py` are from a very old revision (which is why they don't pass). I'm guessing someone made a merge error a while back--easy to do when people are moving/making new files.

shoshber:
we can get rid of `common/test_report`. It is left over from when multiple people had set up different test dirs.

PR #58, opened 2016-12-02
[WIP] specify nipype version to be on the safe side

chrisfilo:


chrisfilo:
Could not figure it out. This will not be important when we release.

PR #64, opened 2016-12-03
[RTM] plot masked images if used as an option in coregistration

### LINKED ISSUE #61: RobustMNINormalizationRPT should mask images before plotting
If moving and or fixed masks were set as inputs the reportlet should present masked version of images (so the voxels not included in optimization would not distract the user like in this example: https://35-66973489-gh.circle-artifacts.com/0/home/ubuntu/ds005/scratch/work/workflow_enumerator/01/t1w_preprocessing/T1_2_MNI_Registration/report.svg)

chrisfilo:
Fixes #61

### DIFF
--- a/niworkflows/common/report.py
+++ b/niworkflows/common/report.py
@@ -7,9 +7,11 @@ import os
 from sys import version_info
 from abc import abstractmethod
 from io import open
+import nibabel as nb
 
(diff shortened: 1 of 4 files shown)


chrisfilo:
https://120-63891310-gh.circle-artifacts.com/0/home/ubuntu/scratch/testRobustMNINormalizationRPT_masked.svg here's an example when only fixed image was masked, but it should work when there are masks provided for both images as well.

PR #68, opened 2016-12-05
[RTM] use only fixed image mask for reports

chrisfilo:
Fixes https://github.com/poldracklab/fmriprep/issues/222

### DIFF
--- a/niworkflows/common/report.py
+++ b/niworkflows/common/report.py
@@ -113,7 +113,6 @@ class RegistrationRC(ReportCapableInterface):
         self._fixed_image = None
         self._moving_image = None
         self._fixed_image_mask = None
-        self._moving_image_mask = None
         super(RegistrationRC, self).__init__(**inputs)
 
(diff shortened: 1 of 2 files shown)


chrisfilo:
This is how it looks like now:
https://126-63891310-gh.circle-artifacts.com/0/home/ubuntu/scratch/testRobustMNINormalizationRPT_masked.svg

This is how it used to looks if we mask only the fixed image:
https://125-63891310-gh.circle-artifacts.com/0/home/ubuntu/scratch/testRobustMNINormalizationRPT_masked.svg

This is how it used to look like without masking:
https://125-63891310-gh.circle-artifacts.com/0/home/ubuntu/scratch/testANTSRegistrationRPT.svg

WDYT?

oesteban:
I would use the fixed mask for both images (reference and moving), in this case.

chrisfilo:
Agreed.

PR #70, opened 2016-12-05
[WIP] dilate the mask prior to the robust registration

chrisfilo:


chrisfilo:
Dilating the masks made the results worse. It seem they need to be very tight.

PR #77, opened 2016-12-07
WIP/ENH plot outline of FLIRT input wm_seg

shoshber:
In the artifacts on circle, it shows the outline of the brain mask. This is because it is more convenient in testing to use already-available files.

Here is what it looks like with a white matter outline: https://github.com/shoshber/pastebin/blob/master/reports/wm_mask.svg

### DIFF
--- a/niworkflows/common/report.py
+++ b/niworkflows/common/report.py
@@ -126,7 +126,10 @@ class RegistrationRC(ReportCapableInterface):
 
         fixed_image_nii = load_img(self._fixed_image)
         moving_image_nii = load_img(self._moving_image)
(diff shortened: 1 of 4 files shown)


oesteban:
Looking at the error report, you might need to add a ```self._contour = None``` to the ```__init__``` of the RegistrationRC class.

or use ```getattr(self, '_contour', None)``` instead.

I would like first better, but it is up to the coder.

The report is really good-looking :+1:

shoshber:
circle passed here: https://circleci.com/gh/shoshber/niworkflows/5#artifacts/containers/0

chrisfilo:
thank you

PR #78, opened 2016-12-07
[ENH,RTM] Finish up ApplyXFM implementation

oesteban:
It was half-implemented but commented out.


shoshber:
I wrote this on slack but I think you missed it: "shouldn't ApplyXFMRPT just inherit from FLIRTRPT? both their `_post_run_hook`s look the same.
or is this a goofy mixin thing?"

PR #88, opened 2016-12-10
[RTM] aCompCor reportlets

### LINKED ISSUE #86: Add aCompCor reportlet
Analogous to #85

chrisfilo:
Closes #86


niworkflows/common/report.py
```diff
@@ -56,17 +56,14 @@ def _run_interface(self, runtime):
         # check exit code and act consequently
         NIWORKFLOWS_LOG.debug('Running report generation code')
```
shoshber: This changes `_run_interface`'s behavior a bit. Previously, if generating a report raised an error, _run_interface would attempt to generate an error report. Now that is not the case. That's fine by me but just want to make sure that it is intentional.

PR #93, opened 2016-12-13
[RTM] add n4 bias field correction to afni skullstrip workflow, include it …

rwblair:
…in the output


rwblair:
Still need to point nipype at the correct commit.

chrisfilo:
You can point to your fork while this is WIP - so the test would pass when you remove N4 from the main workflow.

PR #96, opened 2016-12-14
[ENH,RTM] Fieldmap-correction RC interfaces

oesteban:
- Adds the FUGUE interface
- Rename labels in output reportlet
- Add option to set a reference WM contour

### DIFF
--- a/niworkflows/interfaces/registration.py
+++ b/niworkflows/interfaces/registration.py
@@ -9,7 +9,7 @@ from nipype.interfaces import fsl, ants
 from niworkflows.anat import mni
 import niworkflows.common.report as nrc
 from niworkflows import NIWORKFLOWS_LOG
-from nipype.interfaces.base import isdefined
+from nipype.interfaces.base import isdefined, File
 from nilearn.image import index_img
 
 class RobustMNINormalizationInputSpecRPT(
@@ -75,7 +75,8 @@ class ANTSApplyTransformsRPT(nrc.RegistrationRC, ants.ApplyTransforms):
 
 class ApplyTOPUPInputSpecRPT(nrc.RegistrationRCInputSpec,
                              fsl.epi.ApplyTOPUPInputSpec):
-    pass
+    wm_seg = File(argstr='-wmseg %s',
+                  desc='reference white matter segmentation mask')
 
 
 class ApplyTOPUPOutputSpecRPT(nrc.ReportCapableOutputSpec,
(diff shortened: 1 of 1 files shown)


niworkflows/interfaces/registration.py
```diff
@@ -88,14 +89,40 @@ class ApplyTOPUPRPT(nrc.RegistrationRC, fsl.ApplyTOPUP):
     output_spec = ApplyTOPUPOutputSpecRPT
 
     def _post_run_hook(self, runtime):
-        self._fixed_image_label = "before"
-        self._moving_image_label = "after"
-        self._fixed_image = index_img(self.inputs.in_files[0], 0)
-        self._moving_image = index_img(self.aggregate_outputs().out_corrected, 0)
-        NIWORKFLOWS_LOG.info('Report - setting before (%s) and after (%s) images',
+        self._fixed_image_label = "corrected"
+        self._moving_image_label = "warped"
```
chrisfilo: I find this selection of words a bit confusing. Afterall, you apply a non-linear warp filed tot correct the image so technically both images are "warped". That's why I chose 'before' and 'after' because it's so unambiguous. I leave the final decision to your discretion though.

niworkflows/interfaces/registration.py
```diff
@@ -88,14 +89,40 @@ class ApplyTOPUPRPT(nrc.RegistrationRC, fsl.ApplyTOPUP):
...
+        self._fixed_image = index_img(self.aggregate_outputs().out_corrected, 0)
+        self._moving_image = index_img(self.inputs.in_files[0], 0)
+        self._contour = self.inputs.wm_seg if isdefined(self.inputs.wm_seg) else None
```
chrisfilo: Does ApplyTOPUP even have a `wm_seg` input?

rwblair: http://nipype.readthedocs.io/en/latest/interfaces/generated/nipype.interfaces.fsl.epi.html#applytopup

Does not look like it from the docs.

oesteban: No, I added that trait in the report capable input spec.

niworkflows/interfaces/registration.py
oesteban: I have a different approach, somehow my brain links more quickly 'corrected' to "this is important" and warped to "this is the reference". Before/after requires the knowledge about what you are actually doing (I agree that's very present if you look at the final report, with the title and descriptions). Probably, the addition of the WM contour will highlight the corrected vs. warped concepts.

That said, I don't have a particular interest in the naming: I will roll back if the previous argument does not convince you.

PR #105, opened 2017-01-12
Have registration SVGs start out paused, set to running on hover. sty…

rwblair:
…le change in report.py.

The note to hover to play animation might live in fmriprep or it might live here, I haven't decided yet.


rwblair:
Not about playing animation will have to go in fmriprep template since this code is only generating the svg itself.

PR #108, opened 2017-01-12
automatic deployment to pypi

chrisfilo:


chrisfilo:
I need to merge this to be able to test it

PR #109, opened 2017-01-12
[RTM] blank the values of height and width in the svg etree object

rwblair:


effigies:
When this gets re-opened, check the artifacts. I was having trouble viewing images when my commits were rebased on top of 297c814, both locally and on CircleCI.

PR #115, opened 2017-01-18
ReconAll report

### LINKED ISSUE #102: Add recon_all reportlet
Inspired by https://surfer.nmr.mgh.harvard.edu/fswiki/QATools, but simpler (without the skullstrippping or 3D plots). Limited to outlines of pial and white matter surfaces. Similar to the segmentation plot, but with thinner lines and potentailly more slices.

effigies:
As discussed in #102, shows white/pial surfaces overlaid on masked T1.

`SurfaceSegmentationRC` is derived more from `RegistrationRC` with the "moving" image removed, as FreeSurfer doesn't have quite the equivalents of FAST's 3-class output files... `aseg.mgz` could be used, but it has many subcortical divisions, which would need to be mapped into white, gray, and possibly CSF.

As recon-all is very long to run, report interfaces are extended to allow mock runs, which assumes necessary files have already been created. At present, I believe this requires that the necessary files can be derived from the InputSpec, as the OutputSpec shouldn't be properly instantiated.

Closes #102.

### DIFF
--- a/niworkflows/common/report.py
+++ b/niworkflows/common/report.py
@@ -23,6 +23,8 @@ class ReportCapableInputSpec(BaseInterfaceInputSpec):
         False, usedefault=True, desc="Set to true to enable report generation for node")
     out_report = File(
         'report.html', usedefault=True, desc='filename for the visual report')
+    mock_run = traits.Bool(False, usedefault=True,
+                           desc="Set to true to skip running the base interface")
 
 
 class ReportCapableOutputSpec(TraitedSpec):
@@ -40,11 +42,14 @@ class ReportCapableInterface(BaseInterface):
     def _run_interface(self, runtime):
         """ delegates to base interface run method, then attempts to generate reports """
 
-        try:
-            runtime = super(
-                ReportCapableInterface, self)._run_interface(runtime)
-        except NotImplementedError:
-            pass  # the interface is derived from BaseInterface
+        if not self.inputs.mock_run:
+            try:
+                runtime = super(
+                    ReportCapableInterface, self)._run_interface(runtime)
+            except NotImplementedError:
+                pass  # the interface is derived from BaseInterface
+        else:
+            runtime.returncode = 0
 
         # leave early if there's nothing to do
         if not self.inputs.generate_report:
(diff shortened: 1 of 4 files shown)


effigies:
Hmmm. Only showing sagittal and coronal slices. This isn't quite ready.



Suspect it's because I removed the "Fixed" label.

niworkflows/viz/utils.py
```diff
@@ -259,7 +259,7 @@ def compose_view(bg_svgs, fg_svgs, ref=0, out_file='report.svg'):
     # Read all svg files and get roots
     svgs = [svgt.fromfile(f) for f in bg_svgs + fg_svgs]
     roots = [f.getroot() for f in svgs]
-    nsvgs = len(svgs) // 2
+    nsvgs = len(bg_svgs)
```
effigies: This was the main issue. With only `bg_svgs`, `nsvgs == 1`, so on the first round through line 286, `yoffset` was set to 0.

Because `len(fg_svgs) in (0, len(bg_svgs)`, this shouldn't change the two-layer functioning.

niworkflows/viz/utils.py
```diff
@@ -287,24 +287,27 @@ def compose_view(bg_svgs, fg_svgs, ref=0, out_file='report.svg'):
...
-        svgt.GroupElement(roots[:3], {'class': 'background-svg'}),
-        svgt.GroupElement(roots[3:], {'class': 'foreground-svg'})
-    ]
+    if fg_svgs:
+        newroots = [
+            svgt.GroupElement(roots[:3], {'class': 'background-svg'}),
+            svgt.GroupElement(roots[3:], {'class': 'foreground-svg'})
+        ]
+    else:
+        newroots = roots
```
effigies: Two layers only applies when `fg_svgs` is present.

niworkflows/common/report.py
```diff
@@ -23,6 +23,8 @@ class ReportCapableInputSpec(BaseInterfaceInputSpec):
         False, usedefault=True, desc="Set to true to enable report generation for node")
     out_report = File(
         'report.html', usedefault=True, desc='filename for the visual report')
+    mock_run = traits.Bool(False, usedefault=True,
```
oesteban: Could this be a private ```ReportCapableInterface``` member? I think it makes more sense that way.

- this is just a detail, this could perfectly be merged without this change.

effigies:
The latest commit keeps the calling style the same, but removes the documentation implicit in the inputspec. Let me know if there's another, preferred way to do this.

oesteban:
Mmm, maybe a good point to ask for @chrisfilo's opinion.

I meant that ```mock_run``` shouldn't be an input, is more like a state variable of the interface itself, don't you think?. With the new property (which is exactly what I was suggesting), you can always access to the variable:

```
interface.mock_run = True  # Using interface directly
-- or --
node.interface.mock_run = True  # Within a workflow
```

The only thing missing would be a setter for the interfaces.

oesteban:
I mean, this is relevant iff we finally merge the reports into nipype. For using our own, it really doesn't matter.

effigies:
It's no big deal to me, however you want to do it. So you want private variable `_mock_run` and getter/setter `mock_run` methods?

oesteban:
I think so, but it is no big deal to me either :+1:

effigies:
Cool. Passing.

PR #117, opened 2017-02-10
[RTM] Typo fixed in nipype

effigies:
nipy/nipype#1784 has been merged, so we can track nipype/master. Remove version constraint from `info.py`, to avoid overwriting pinned copy.

Precondition for poldracklab/fmriprep#347.

### DIFF
--- a/niworkflows/info.py
+++ b/niworkflows/info.py
@@ -49,7 +49,7 @@ CLASSIFIERS = [
     'Programming Language :: Python :: 3.5',
 ]
 
-REQUIRES = ['nipype>=0.13.0rc1', 'future', 'nilearn>=0.2.6', 'sklearn', 'pandas', 'matplotlib',
+REQUIRES = ['nipype', 'future', 'nilearn>=0.2.6', 'sklearn', 'pandas', 'matplotlib',
             'jinja2', 'svgutils', 'tinycss']
 SETUP_REQUIRES = []
 REQUIRES += SETUP_REQUIRES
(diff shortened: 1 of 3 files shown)


oesteban:
Not sure if this will work with the pypi distribution of fmriprep. That will look into the info.py package and fall back to the latest full release of nipype (0.12). In that environment, requiring nipype>=0.13.0-rc2 is necessary.

effigies:
Until we're ready to release, this is for people working off of master. If there's a better way to do it (e.g. a branch in poldracklab/niworkflows that's not master, that's fine with me.

Or am I missing something? Dealing with pins and versions is new to me.

oesteban:
Basically, we have two ways to declare dependencies in python packages (and therefore, you can pin packages to specific versions):

1 - Via pip and requirements.txt
2 - Via setuptools and the setup.py (in our case delegated to info.py).

When you run `pip install fmriprep[install-opts]` you are using option 2. It will fetch the package and run `python setup.py install` on it, and therefore take as requirement whatever it is declared as dependency in `setup.py` (our `info.py`). In theory, you could point here to git repos to install any branch of them using the `link_dependencies` option. In practice, this does not work very well. It could well be due to the shady way setuptools manages the link_dependencies or that nipype versioning is not 100% compatible with setuptools, I don't know :(. But the fact is we never got it to work well.

For development environments, then 1) is more appropriate. You *need* to run `pip install -r requirements.txt` (here's when pinning happens, and it is quite effective - whatever you have in the requirements file is installed). Then you can do `pip install -e .[install-opts]` to install the package in developer mode. If you don't run the `pip install -r` it will fall back to option 2).

I wish I could provide clearer descriptions, but this field is also very blurry to me.

chrisfilo:
This is just a temporary measure to help working with the bleeding edge versions of nipype. Before releasing we should revert back to specifying minimal version of nipype in info.py. Let's not worry about pypi until then.

oesteban:
Alright then!

PR #120, opened 2017-02-14
[RFC] Set default initialization selections for BBRegisterRPT

effigies:
FS6 has a new default initialization method for `bbregister` (`mri_coreg`). I've added an updated inputspec to Nipype with nipy/nipype#1811, but, as discussed Friday, I'm trying to avoid depending on unreleased Nipype features.

I'm proposing to use `--init-coreg` in FSv6.0, and `--init-fsl` in FSv5.3. By using a modified `init` trait in both cases, this allows `BBRegisterRPT` to be called without an explicit `init` parameter, in either case, which means the check doesn't need to be made at workflow creation.

What do people think about (a) the initialization choices; (b) the method for setting them?


chrisfilo:
It has been a long-standing nipype policy to avoid setting defaults unless necessary. This way parameters used in an analysis are more explicit when reading workflow code.

PR #121, opened 2017-02-15
[WIP] Adding constrained cost-function masking to T1 registration

danlurie:
Key difference is the use of a lesion mask to exclude regions from being evaluated during `antsRegistration`.


niworkflows/anat/mni.py
```diff
@@ -17,28 +17,43 @@
...
         File(exists=True), mandatory=True, desc='image to apply transformation to')
+    # Set an input file to be used as the reference image (instead of the default template).
     reference_image = InputMultiPath(
         File(exists=True), desc='override the reference image')
+    # Set the input file to be used as the moving mask.
     moving_mask = File(exists=True, desc='moving image mask')
+    # Set the input file to be used as the reference mask.
     reference_mask = File(exists=True, desc='reference image mask')
+    # Set the input file to be used as the lesion mask.
+    lesion_mask = File(exists=True, desc='lesion mask image')
```
chrisfilo: Could you specify which image the lesion mask corresponds to (moving or reference)?

chrisfilo:
Love the comment to code ratio!

danlurie:
I'll take a look at this during Brainhack. By "without making any changes to inputs", I assume you mean no lesion mask was provided? I haven't done much testing on that use case yet, as my first priority was to get the CCFM working at all when a mask **is** provided. Good chance there's something broken in the code I wrote to handle fallback situations where there is no mask.

chrisfilo:
Yes - this was without lesion mask being provided.

danlurie:
My local regression tests are passing, so I guess now we wait to see if Circle gets angry about anything. In the meantime, where in the docs should I add info about the lesion masking features?

effigies:
We aren't currently publishing the niworkflows docs, but I think the best place to document this behavior is in the `RobustMNINormalization` docstring, with both some description and an/some example(s), as in [nipype proper](https://github.com/nipy/nipype/blob/4b587d6bc297d202148fa2a7f9ab4de50ff49bb7/nipype/interfaces/ants/registration.py#L1096-L1113).

And then in the fmriprep docs, perhaps a new sub-heading ("Lesion masking" or "Preprocessing datasets with lesions"?) under [T1w/T2w preprocessing](https://fmriprep.readthedocs.io/en/0.6.0/workflows.html#t1w-t2w-preprocessing).

niworkflows/anat/mni.py
```diff
@@ -214,14 +342,102 @@ def _validate_results(self):
...
+    # Set all non-mask voxels in the input file to zero.
     data = in_nii.get_data()
     data[mask_nii.get_data() == 0] = 0
+    # Save the new masked image.
     new_nii = nb.Nifti1Image(data, in_nii.affine, in_nii.header)
     new_nii.to_filename(new_name)
     return os.path.abspath(new_name)
 
+
+def create_cfm(in_file, out_path, lesion_mask, global_mask=True):
```
oesteban: I would suggest the following to make this function more "nipype -ly":

1. Change function signature to make out_path optional: 
    ```Python
    def create_cfm(in_file, lesion_mask=None, global_mask=True, out_path=None):
        ...
    ```

2. Then generate the new path automatically using nipype (preferably at the beginning of the function body):
    ```Python
    if out_path is None:
        out_path = fname_presuffix(in_file, suffix='_lesionmask',  # suffix can be _cfm if you like it better
                                   newpath=os.getcwd())
    else:
        out_path = os.path.abspath(out_path) 
    ```

    This will require the following at the top:
    ```Python
    import os
    ```

3. Use `out_path` whenever possible, once defined that way. A nice side effect of this suggestion is that you always use `out_path` and don't split with that saving and returning a possibly modified value. There's little chance for an error there, but better be consistent by default, right?

4. When you call `create_cfm` above, let nipype generate the name (i.e. do not set `out_path`).

niworkflows/anat/mni.py
```diff
@@ -214,14 +342,102 @@ def _validate_results(self):
...
+
+        # Subtract the lesion mask from the secondary mask.
+        cfm_data = in_data - lm_data
+        cfm_data[cfm_data < 0] = 0
+
+        # Create the cost function mask image from the subtraction.
+        cfm_img = nb.Nifti1Image(cfm_data, in_img.affine, in_img.header)
+    else:
+        # Confirm that global masking is enabled.
+        assert (global_mask is True), "If no lesion mask is provided, global_mask must be True"
```
oesteban: Why don't preemptively raise an Error as I suggested in my PR? -  this would go in the beginning when checking inputs.

niworkflows/anat/mni.py
```diff
@@ -214,14 +342,102 @@ def _validate_results(self):
...
+
+        # Create the cost function mask from the global mask.
+        cfm_img = nb.Nifti1Image(in_data, in_img.affine, in_img.header)
+
+    # Save the cost function mask.
+    cfm_img.to_filename(out_path)
```
oesteban: Please add before the `to_filename` call this:

```Python
cfm_img.set_data_dtype(np.uint8)
```
... to make sure the written file is unsigned char.

niworkflows/anat/mni.py
```diff
@@ -214,14 +342,102 @@ def _validate_results(self):
...
+        in_data = np.ones_like(in_data, dtype=np.uint8)
+
+    # If a lesion mask was provided, combine it with the secondary mask.
+    if lesion_mask is not None:
+        # Reorient the lesion mask and get the data.
+        lm_img = nb.as_closest_canonical(nb.load(lesion_mask))
+        lm_data = lm_img.get_data()
+
+        # Subtract the lesion mask from the secondary mask.
+        cfm_data = in_data - lm_data
```
oesteban: If you rewrite `in_data` you can safely generate the nifti image from it in line 440, just before exiting. If you do that, lines 431 and 432 can go away.

niworkflows/anat/mni.py
```diff
@@ -214,14 +342,102 @@ def _validate_results(self):
...
+        cfm_data[cfm_data < 0] = 0
+
+        # Create the cost function mask image from the subtraction.
+        cfm_img = nb.Nifti1Image(cfm_data, in_img.affine, in_img.header)
+    else:
+        # Confirm that global masking is enabled.
+        assert (global_mask is True), "If no lesion mask is provided, global_mask must be True"
+
+        # Create the cost function mask from the global mask.
+        cfm_img = nb.Nifti1Image(in_data, in_img.affine, in_img.header)
```
oesteban: This can be done for all execution paths (see my PR), since you are checking all possible options. That would also make the logic easier if you also check for the correctness the way I suggested (because that `else:` case would disappear).

(requires applying first the comment on line 428)

niworkflows/anat/mni.py
```diff
@@ -145,44 +184,133 @@ def _get_ants_args(self):
...
+                        op.join(mni_template, '%dmm_%s.nii.gz' % (resolution, self.inputs.reference)),
+                        op.join(mni_template, '%dmm_brainmask.nii.gz' % resolution), "fixed_masked.nii.gz")
+
+                # If a lesion mask is provided...
+                if isdefined(self.inputs.lesion_mask):
+                    # Create a cost function mask with the form: [global mask]
+                    # Use this as the fixed mask.
+                    args['fixed_image_mask'] = create_cfm(
+                        op.join(mni_template, '%dmm_brainmask.nii.gz' % resolution),
+                        "fixed_cfm.nii.gz",
```
oesteban: This can be removed if the way `out_path` is generated is updated.

niworkflows/anat/mni.py
```diff
@@ -145,44 +184,133 @@ def _get_ants_args(self):
...
+                        lesion_mask=None,
```
oesteban: These two default parameters can be removed

niworkflows/anat/mni.py
```diff
@@ -145,44 +184,133 @@ def _get_ants_args(self):
...
                     args['fixed_image_mask'] = self.inputs.reference_mask
+
+            # If no reference mask is provided...
+            # But a lesion mask *IS* provided ...
+            elif isdefined(self.inputs.lesion_mask):
+                # Create a cost function mask with the form: [global mask]
+                # Use this as the fixed mask
+                args['fixed_image_mask'] = create_cfm(
+                    self.inputs.reference_image,
+                    "fixed_cfm.nii.gz",
```
oesteban: This can be removed if `out_path` is generated as suggested.

niworkflows/anat/mni.py
```diff
@@ -145,44 +184,133 @@ def _get_ants_args(self):
...
+                    lesion_mask=None,
```
oesteban: Default arguments `lesion_mask` and `global_mask` can be omitted

niworkflows/anat/mni.py
```diff
@@ -145,44 +184,133 @@ def _get_ants_args(self):
...
+                        self.inputs.reference_mask,
                         "fixed_masked.nii.gz")
+
+                    # If a lesion mask is also provided...
+                    if isdefined(self.inputs.lesion_mask):
+                        # Create a cost function mask with the form: [global mask]
+                        # Use this as the fixed mask.
+                        args['fixed_image_mask'] = create_cfm(
+                            self.inputs.reference_mask,
+                            "fixed_cfm.nii.gz",
```
oesteban: See comments below

niworkflows/anat/mni.py
```diff
@@ -145,44 +184,133 @@ def _get_ants_args(self):
...
+
+            # If a reference mask is provided...
             if isdefined(self.inputs.reference_mask):
+                # If explicit masking is enabled...
                 if self.inputs.explicit_masking:
+                    # Mask the reference image.
+                    # Do not use a fixed mask during registration.
                     args['fixed_image'] = mask(
                         self.inputs.reference_image,
-                        self.inputs.mreference_mask,
```
oesteban: Good catch!

niworkflows/anat/mni.py
```diff
@@ -145,44 +184,133 @@ def _get_ants_args(self):
...
+                    global_mask=self.inputs.explicit_masking)
+        
+        # If no moving mask is provided...
+        # But a lesion mask *IS* provided...
+        elif isdefined(self.inputs.lesion_mask):
+            # Create a cost function mask with the form: [global mask - lesion mask]
+            # Use this as the moving mask.
+            args['moving_image_mask'] = create_cfm(
+                self.inputs.moving_image,
+                "moving_cfm.nii.gz",
```
oesteban: See comments below

niworkflows/anat/mni.py
```diff
@@ -145,44 +184,133 @@ def _get_ants_args(self):
...
 
+            # If a lesion mask is also provided...
+            if isdefined(self.inputs.lesion_mask):
+                # Create a cost function mask with the form:
+                # [global mask - lesion mask] (if explicit masking is enabled)
+                # [moving mask - lesion mask] (if explicit masking is disabled)
+                # Use this as the moving mask.
+                args['moving_image_mask'] = create_cfm(
+                    self.inputs.moving_mask,
+                    "moving_cfm.nii.gz",
```
oesteban: See comments below

niworkflows/anat/mni.py
```diff
@@ -214,14 +342,102 @@ def _validate_results(self):
...
+    import os
+    import numpy as np
+    import nibabel as nb
+    
+    # Load the input image
+    in_img = nb.load(in_file)
+    in_data = in_img.get_data()
+
+    # If we want a global mask, create one based on the input image.
+    if global_mask is True:
```
oesteban: If you preemptively check for consistency of `global_mask` and `lesion_mask` (see comment on line 435), then you can make this more pythonic:

```Python
if global_mask:
```

oesteban:
One more comment I missed before:

We could consider adding two additional outputs to this interface: `out_fixed_mask` and `out_moving_mask`. These two would be defined only if they are modified and they are not all ones (so the only case is when a `lesion_mask` has been set).

oesteban:
Hi @danlurie, do you think we can push this forward during next week? It'd be great if we finish up with integrating your code in fmriprep.

danlurie:
Hey Oscar,

I’m planning on trying to wrap this up this weekend. Will post an update
probably on Sunday or Monday.

Dan

On August 18, 2017 at 12:13:24 PM, Oscar Esteban (notifications@github.com)
wrote:

Hi @danlurie <https://github.com/danlurie>, do you think we can push this
forward during next week? It'd be great if we finish up with integrating
your code in fmriprep.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/poldracklab/niworkflows/pull/121#issuecomment-323437973>,
or mute the thread

oesteban:
Awesome!

effigies:
Hey @danlurie I merged in the latest niworkflows. If you want to check 9e426921314d0ec3e55353d2ca035b4fc42671fb to make sure I didn't disrupt changes you made, that'd be great.

PR #128, opened 2017-03-01
BBRegisterRPT: Use inputs.subjects_dir to find structurals

effigies:
The reporting in `BBRegisterRPT` ignores `subjects_dir` when passed through nipype.

Also prefers `brainmask.mgz` to `T1.mgz`, to more closely resemble the output of `FLIRTRPT`.

### DIFF
--- a/niworkflows/interfaces/registration.py
+++ b/niworkflows/interfaces/registration.py
@@ -167,15 +167,11 @@ class BBRegisterRPT(nrc.RegistrationRC, freesurfer.BBRegister):
     output_spec = BBRegisterOutputSpecRPT
 
     def _post_run_hook(self, runtime):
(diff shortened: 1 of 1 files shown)


chrisfilo:
I don't understand the comment about FLIRTRPT - it plots whatever it got on input. Which is the full brain (unskullstripped) volume:

image

effigies:
In which case, I've forgotten why I switched it.

chrisfilo:
It's not a very important issue - I'm happy to go either way.

effigies:
At least [here](https://128-78558509-gh.circle-artifacts.com/0/home/ubuntu/ds005/out/fmriprep/sub-01.html) (no fieldmaps), BBR is done against the masked T1.

chrisfilo:
Indeed you are right (I was looking at old reports).

PR #133, opened 2017-03-08
Compress surface segmentation reports

### LINKED ISSUE #132: Compression is not currently attempted for SurfaceSegmentationRC

effigies:
Should be an easy one. I'll check the artifacts and merge if the ReconAll output is compressed.

Fixes #132.


effigies:
The output was previously compressed, since `svg_compress` is set to `compress=auto` by default.

But it's now still compressed and can be manipulated via the `compress_report` option.

chrisfilo:
Ah yes - you are right.

PR #136, opened 2017-03-09
[WIP] initial commit to add ants to fsl conversion for motion parameters

rwblair:


oesteban:
Looking at the code, why don't we submit this directly to nipype? WDYT @chrisfilo?

rwblair:
welp looks like I have some similar code in my pull request to nipype. Let me test that code and see if its any good...

oesteban:
Yeah, I think this is exactly the same we are pulling into nipype, right? Feel free to reopen if I'm wrong.

PR #146, opened 2017-04-01
[RTM] Colection of minor visual improvements

chrisfilo:
Closes https://github.com/poldracklab/fmriprep/issues/259

### DIFF
--- a/niworkflows/interfaces/segmentation.py
+++ b/niworkflows/interfaces/segmentation.py
@@ -36,7 +36,7 @@ class FASTRPT(nrc.SegmentationRC,
         overlaid '''
         self._anat_file = self.inputs.in_files[0],
         self._mask_file = self.aggregate_outputs().tissue_class_map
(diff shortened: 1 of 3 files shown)


niworkflows/interfaces/segmentation.py
```diff
@@ -36,7 +36,7 @@ def _post_run_hook(self, runtime):
         overlaid '''
         self._anat_file = self.inputs.in_files[0],
         self._mask_file = self.aggregate_outputs().tissue_class_map
-        self._seg_files = self.aggregate_outputs().tissue_class_files
+        self._seg_files = self.aggregate_outputs().tissue_class_files[1:]
```
oesteban: What class is discarded here? (just to double-check)

chrisfilo: This is the CSF class which in combination with other classes was showing skullstripping mask. I've added a comment explaining this.

PR #148, opened 2017-04-09
Moved ref image estimator from FMRIPREP

chrisfilo:
I'm planning to use it in MRIQC


oesteban:
I would start a new module under interfaces for these various tools, but I don't mind merging as is. Whatever you prefer.

chrisfilo:
I hope this is temporary and I will move this to nipype soon.

PR #151, opened 2017-04-20
FIX: LooseVersion bug in interfaces/registration.py

danlurie:
When attempting to check the FreeSurfer version, `LooseVersion` was called on `fs.preprocess.FSversion`, which returns a `LooseVersion` object by itself. This resulted in LooseVersion throwing an error (because it expects a string).

I've fixed this and confirmed that the version-check logic works as expected.

(There's also a bunch of detailed comments on `/anat/mni.py` which I wrote for myself when I was first learning how the interface was structured. I can remove those from the PR if that's better.)


effigies:
Ugh, sorry, this is my fault. Can you switch to freesurfer.Info.looseversion()? I was trying not to break apps that depended on FSVersion, and broke our own. I'll submit a patch to nipype, to restore the old behavior, but we shouldn't depend on it.

PR #152, opened 2017-04-20
Move to fs.Info.looseversion() to check FreeSurfer version

effigies:
This PR moves to a dependency on an updated version check. This would have broken things for people updating nipype since ~8am EDT today. This would not affect any Docker builds, due to the earlier requirement. FMRIPREP should be moved to this same nipype commit ASAP once this is merged.

Closes #151.


effigies:
No intermediate changes to nipype master. Safe to merge.

PR #153, opened 2017-04-23
[RTM] Add SimpleInterface subclass of BaseInterface

effigies:
This exists as both [`mriqc.interfaces.base.MRIQCBaseInterface`](https://github.com/poldracklab/mriqc/blob/master/mriqc/interfaces/base.py#L11) and [`fmriprep.interfaces.bids.SimpleInterface`](https://github.com/poldracklab/fmriprep/blob/master/fmriprep/interfaces/bids.py#L38).


effigies:
poldracklab/fmriprep#470 passing. Assuming we want to deduplicate and use consistent naming across projects, this is good to go. Let me know what you think.

oesteban:
:+1:

PR #155, opened 2017-04-29
[ENH] More flexible settings for RobustMNIRegistration

oesteban:
This is useful per se, and also for https://github.com/poldracklab/mriqc/issues/450


oesteban:
Quick look at this? That will allow me start testing stuff in mriqc. Thanks!

effigies:
On the plane to SF. Going to lose internet shortly. I'd just pin against your branch for the moment. Sorry.



Sent on my Samsung Galaxy S® 5


-------- Original message --------

chrisfilo:
Were changes to `t1-mni_registration_testing_000.json` intended to be part of this PR?

oesteban:
Yes, the input `initial_moving_transform_com` now is set to 1 when no input initialization transform is given.

niworkflows/data/t1-mni_registration_testing_000.json
```diff
@@ -1,24 +1,23 @@
...
-    "convergence_window_size": [15, 5],
+    "convergence_window_size": [10, 5],
     "metric": ["Mattes", "Mattes"],
     "metric_weight": [1, 1],
     "radius_or_number_of_bins": [56, 56],
-    "transforms": ["Rigid", "Affine"],
-    "transform_parameters": [[1.0], [1.0]],
-    "number_of_iterations": [[20], [15]],
-    "initial_moving_transform_com": true,
+    "transforms": ["Affine", "BSplineSyN"],
```
chrisfilo: Was this intended to be part of this PR? Seems unrelated to initial transform setting.

oesteban: Yes,  I added the BSplineSyN for two reasons:
  - Produce a more accurate result with testing settings
  - Produce more realistic outputs in testing settings, since now the transform will be required to be collapsed (or whatever name it has in ANTs) in the same way we are doing in production settings.
  - I could add a flag to this interface (something like `only_affine`) to perform only affine registrations (adding the corresponding settings)

WDYT?

chrisfilo: What's the run time difference?

BTW Please try keeping PRs focused on single issues if possible. Combining multiple features in one PR makes reviewing harder and creates unnecessary bottlenecks.

PR #156, opened 2017-04-30
[RTM] updating nipype and fixing API changes

chrisfilo:


niworkflows/interfaces/masks.py
```diff
@@ -189,11 +189,11 @@ def _post_run_hook(self, runtime):
         ''' generates a report showing slices from each axis '''
 
         self._anat_file = self.inputs.realigned_file
-        self._mask_file = self.aggregate_outputs().high_variance_mask
-        self._seg_files = [self.aggregate_outputs().high_variance_mask]
+        self._mask_file = self.aggregate_outputs().high_variance_masks
+        self._seg_files = [self.aggregate_outputs().high_variance_masks]
```
effigies: I think this should be:

```Python
self._mask_file = self.aggregate_outputs().high_variance_masks[0]
self._seg_files = self.aggregate_outputs().high_variance_masks
```

niworkflows/interfaces/masks.py
```diff
@@ -189,11 +189,11 @@ def _post_run_hook(self, runtime):
...
         self._masked = False
         self._report_title = 'tCompCor - high variance voxels'
 
         NIWORKFLOWS_LOG.info('Generating report for tCompCor. file "%s", mask "%s"',
                              self.inputs.realigned_file,
-                             self.aggregate_outputs().high_variance_mask)
\ No newline at end of file
+                             self.aggregate_outputs().high_variance_masks)
```
effigies: `high_variance_masks[0]`

niworkflows/interfaces/masks.py
chrisfilo: The problem is that `high_variance_masks` is an OutputMultiPath so it converts single element lists to just the inner element. We could alternatively run a function on it to convert it back to a list.

effigies: Oh okay. And we're sure in these cases that it will always be a single path?

chrisfilo: But in case of multiple masks this reportlet does not work well anyway (it ignores all but the first mask). I can add an assertion for such case.

effigies: Yeah, maybe just throw an informative exception so that if it ever comes up we won't have to spend any time figuring it out.

PR #157, opened 2017-05-01
[ENH] Add a new family of "faster" registration settings

oesteban:
Defines a new flag "faster" to perform lesser levels of registration.
Combined with subsampled images, this should result in a significant
reduction of registration run-time.


oesteban:
(still working on this, feel free to review but do not merge please)

oesteban:
This is ready for review. It is needed for the new mriqc registration. It'd be great to release pinning to the official niworkflows repo instead of mine.

effigies:
This looks fine, but the deprecated 99.99.99 isn't great. Can that be resolved by making a release?

oesteban:
Yeah, that's the issue I found with versions. The problem is that, if we make a release, it will fail for development environments because we are using this 99.99.99 when it is not a proper release.

effigies:
What about moving to <last release> + 0.0.1, followed by "-dev"? The Circle machinery can handle this easily, so there's no need to do a before-release update. Just one, after release.

oesteban:
@effigies, my blessings to that -- but that should go in a different PR right?

BTW, in mriqc I've been playing around the package [versioneer](https://github.com/warner/python-versioneer):
```
$ mriqc -v
mriqc v0.9.4-rc2+1.gdd8e4a9.dirty
```
Which means: 1 commit after tag 0.9.4-rc2 (commit sha starts with gdd8e4a9) and the _dirty_ bit indicating that I have changes not committed.

effigies:
Sure. But we should have this sorted before the next release. And given that we're having a deprecation, we should probably release soon.

PR #158, opened 2017-05-02
ENH: Expose get_dataset and a template name-resource map

effigies:
This is intended to enable poldracklab/fmriprep#498. It enables selecting a data grabber by template name as used in fmriprep.


effigies:
poldracklab/fmriprep#498 is testing clean. Any concerns with how I've done this?

PR #161, opened 2017-05-15
[RTM] Added simple interface for visualising masks.

chrisfilo:


effigies:
Could smoke test this pretty quickly by hooking up its inputs to the outputs of [BETRPT](https://github.com/poldracklab/niworkflows/blob/master/niworkflows/tests/test_reports.py#L135-L145).

chrisfilo:
https://79-75443898-gh.circle-artifacts.com/0/home/ubuntu/scratch/py3/testSimpleShowMaskRPT.html

effigies:
Is that what that mask is supposed to look like?

chrisfilo:
Yeah it's a broad mask used to speed up ANTs brain extraction.

PR #164, opened 2017-05-17
Dev version: Use last release + 0.0.1 + '-dev'

### LINKED ISSUE #163: Fix development versioning
Replace '99.99.99' by '0.0.1-dev' format.

effigies:
This version will resolve to `< 0.0.7`, which allows us to use meaningful deprecation versions, as discussed in #157.

This approach maintains the ability for Circle to replace the version on release, as well as allows meaningful version comparisons for git repositories. The additional work of updating the dev version number occurs *after* the release, so there's no risk of this resulting in a need for a quick release cycle because we screwed up the version number.

@oesteban mentioned a new system he's using in MRIQC. I'm okay with switching to that if that seems like a better option.

Fixes #163

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,1 +1,1 @@
--e git+https://github.com/nipy/nipype.git@cda3fdc3be231f2c9d7bdbdbadee472a0bad9e92#egg=nipype
+-e git+https://github.com/nipy/nipype.git@c13f967506aa426c63ad4131cf29b655f742649c#egg=nipype
--- a/niworkflows/anat/mni.py
+++ b/niworkflows/anat/mni.py
@@ -8,6 +8,7 @@ from os import path as op
 import shutil
 import pkg_resources as pkgr
 from multiprocessing import cpu_count
+from distutils.version import LooseVersion
 
 from nipype.interfaces.ants.registration import Registration, RegistrationOutputSpec
 from nipype.interfaces.ants.resampling import ApplyTransforms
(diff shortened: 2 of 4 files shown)


effigies:
It's been a while since a release, anyway. If this goes in, I'd say let's do a 0.0.7 release, to make sure I'm right about how this will work.

If it does work, I'd say let's move to either this system or MRIQC's system for FMRIPREP, just so we don't have three different systems going.

oesteban:
I think for FMRIPREP it may make sense a more elaborate versioning. But versioneer could be an overkill.

Let's replicate this simple solution into FMRIPREP and if versioneer does not raise any problems for mriqc then we can switch FMRIPREP knowing it will work well. 

+1000 to a new release of niworkflows.

effigies:
Looks like deprecation tests against the nipype version, not the niworkflows version. I'm going to submit a PR that makes the version they check against interface-specific (nipype by default).

effigies:
Relevant line: https://github.com/nipy/nipype/blob/f39e222499cca6647d732e9164a75339f1f3a6a7/nipype/interfaces/base.py#L447

oesteban:
Good catch!

requirements.txt
```diff
@@ -1 +1 @@
--e git+https://github.com/nipy/nipype.git@f66a5a6f13a151517cd47b43d998c1fee840c55c#egg=nipype
+-e git+https://github.com/effigies/nipype.git@6f3083adf8dd230d2568ea3a31080adc4665a0d5#egg=nipype
```
oesteban: Should we wait until the PR is merged into nipype?

effigies: Yeah. There's no good way to do deprecations until then.

effigies:
This should be good to go, pending a final test.

PR #165, opened 2017-05-17
[RTM] added assertion for testing normalization validity

chrisfilo:


niworkflows/anat/mni.py
```diff
@@ -173,6 +173,42 @@ def _config_ants(self, ants_settings):
...
+    def _get_resolution(self):
+        resolution = self.inputs.template_resolution
+        if self.inputs.testing:
+            resolution = 2
+        return resolution
+
+    def _get_template(self):
+        get_template = getattr(getters, 'get_{}'.format(self.inputs.template))
+        mni_template = get_template()
+        return mni_template
```
effigies: Is there a reason we create the function string, when that function just calls `get_dataset(template)`?

```Python
def _get_template(self):
    return getters.get_dataset(self.inputs.template)
```

PR #166, opened 2017-05-18
[ENH] Give access to num_threads of N4BiasFieldCorrection

oesteban:
Defaults to 1 (keep current behavior). N4BiasFieldCorrection is part
of ANTs and the number of threads are controlled in the same way you'd
do with antsRegistration.


niworkflows/anat/skullstrip.py
```diff
@@ -23,7 +23,8 @@ def afni_wf(name='AFNISkullStripWorkflow'):
     outputnode = pe.Node(niu.IdentityInterface(
         fields=['bias_corrected', 'out_file', 'out_mask', 'bias_image']), name='outputnode')
 
-    inu_n4 = pe.Node(ants.N4BiasFieldCorrection(dimension=3, save_bias=True),
+    inu_n4 = pe.Node(ants.N4BiasFieldCorrection(
+        dimension=3, save_bias=True, num_threads=n4_nthreads), num_threads=n4_nthreads,
                      name='CorrectINU')
```
effigies: `Node` takes an `n_procs` parameter, not `num_threads`. Also PEP8 fix for white space:

```Python
inu_n4 = pe.Node(
    ants.N4BiasFieldCorrection(dimension=3, save_bias=True, num_threads=n4_nthreads),
    n_procs=n4_threads,
    name='inu_n4')
```

And should we be copying the header, as in [fmriprep](https://github.com/poldracklab/fmriprep/blob/master/fmriprep/workflows/util.py#L64-L79)? Maybe it would make sense to move that here, and thread in the `n4_threads`?

niworkflows/anat/skullstrip.py
```diff
@@ -26,8 +28,12 @@ def afni_wf(name='AFNISkullStripWorkflow', unifize=False):
     outputnode = pe.Node(niu.IdentityInterface(
         fields=['bias_corrected', 'out_file', 'out_mask', 'bias_image']), name='outputnode')
 
-    inu_n4 = pe.Node(ants.N4BiasFieldCorrection(dimension=3, save_bias=True),
-                     name='CorrectINU')
+    inu_n4 = pe.Node(
+        ants.N4BiasFieldCorrection(dimension=3, save_bias=True, num_threads=n4_nthreads),
+        n_procs=n4_nthreads,
+        name='inu_n4')
+    orig_hdr = pe.Node(CopyHeader(), name='orig_hdr')
```
effigies: `orig_hdr` has no `in_file` input.

effigies:
Very prompt fix.

niworkflows/anat/skullstrip.py
oesteban: yes, just noticed (https://github.com/poldracklab/niworkflows/pull/166/commits/f57c315dd5b73edaaf43bd515e2197395de59def)

effigies:
We could avoid issues like the above by pulling the whole `init_n4bias_wf` workflow from fmriprep.

oesteban:
That would create a circular dependency between both. IMHO, the permanent solution to this would be to include the feature (with a proper boolean switch) in the `N4BiasFieldCorrection` nipype interface.

chrisfilo:
I read this as "moving" `init_n4bias_wf` to niworkflows not importing it.

oesteban:
Oh, ok, my bad.

oesteban:
Let's make it easier: https://github.com/nipy/nipype/pull/2034

PR #168, opened 2017-05-18
[ENH] Copy CopyHeader interface from fmriprep

oesteban:
It will be useful in mriqc as well (in fact it is probably replicated there)

### DIFF
--- a/niworkflows/interfaces/utils.py
+++ b/niworkflows/interfaces/utils.py
@@ -7,10 +7,9 @@ Utilities
 
 """
 
-import os.path as op
 import nibabel as nb
+from nipype.utils.filemanip import fname_presuffix
 from nipype.interfaces.base import File, OutputMultiPath, BaseInterfaceInputSpec, TraitedSpec
-
 from .base import SimpleInterface
 
 class CopyHeaderInputSpec(BaseInterfaceInputSpec):
@@ -25,17 +24,13 @@ class CopyHeader(SimpleInterface):
     output_spec = CopyHeaderOutputSpec
 
     def _run_interface(self, runtime):
+        in_img = nb.load(self.inputs.hdr_file)
+        out_img = nb.load(self.inputs.in_file, mmap=False)
+        new_img = out_img.__class__(out_img.get_data(), in_img.affine, in_img.header)
+        new_img.set_data_dtype(out_img.get_data_dtype())
 
-        hdr = nb.load(self.inputs.hdr_file).get_header().copy()
-        aff = nb.load(self.inputs.hdr_file).get_affine()
-        data = nb.load(self.inputs.in_file).get_data()
-
-        fname, ext = op.splitext(op.basename(self.inputs.in_file))
-        if ext == '.gz':
-            fname, ext2 = op.splitext(fname)
-            ext = ext2 + ext
-
-        out_name = op.abspath('{}_fixhdr{}'.format(fname, ext))
-        nb.Nifti1Image(data, aff, hdr).to_filename(out_name)
+        out_name = fname_presuffix(self.inputs.in_file,
+                                   suffix='_fixhdr', newpath='.')
+        new_img.to_filename(out_name)
         self._results['out_file'] = out_name
         return runtime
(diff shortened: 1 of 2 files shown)


niworkflows/interfaces/utils.py
```diff
@@ -0,0 +1,41 @@
...
+    out_file = OutputMultiPath(File(exists=True, desc='written file path'))
+
+class CopyHeader(SimpleInterface):
+    input_spec = CopyHeaderInputSpec
+    output_spec = CopyHeaderOutputSpec
+
+    def _run_interface(self, runtime):
+
+        hdr = nb.load(self.inputs.hdr_file).get_header().copy()
+        aff = nb.load(self.inputs.hdr_file).get_affine()
```
effigies: Since this could be a `.nii.gz`, you might not be `mmap`ing, so should just read once.

```Python
hdr_img = nb.load(self.inputs.hdr_file)
```

niworkflows/interfaces/utils.py
```diff
@@ -0,0 +1,41 @@
...
+        data = nb.load(self.inputs.in_file).get_data()
+
+        fname, ext = op.splitext(op.basename(self.inputs.in_file))
+        if ext == '.gz':
+            fname, ext2 = op.splitext(fname)
+            ext = ext2 + ext
+
+        out_name = op.abspath('{}_fixhdr{}'.format(fname, ext))
+        nb.Nifti1Image(data, aff, hdr).to_filename(out_name)
```
effigies: Down here. you can just do:

```Python
nb.Nifti1Image(data, hdr_img.affine, hdr_img.header).to_filename(out_name)
```

Are we worried about setting the on-disk `dtype` to match the `in_file`?

niworkflows/interfaces/utils.py
```diff
@@ -0,0 +1,41 @@
...
+        hdr = nb.load(self.inputs.hdr_file).get_header().copy()
+        aff = nb.load(self.inputs.hdr_file).get_affine()
+        data = nb.load(self.inputs.in_file).get_data()
+
+        fname, ext = op.splitext(op.basename(self.inputs.in_file))
+        if ext == '.gz':
+            fname, ext2 = op.splitext(fname)
+            ext = ext2 + ext
+
+        out_name = op.abspath('{}_fixhdr{}'.format(fname, ext))
```
effigies: Setting the `out_name` can be done trivially:

```Python
out_name = nipype.utils.filemanip.fname_presuffix(self.inputs.in_file, suffix='_fixhdr', newpath='.')
```

niworkflows/interfaces/utils.py
```diff
@@ -0,0 +1,36 @@
...
+class CopyHeaderOutputSpec(TraitedSpec):
+    out_file = OutputMultiPath(File(exists=True, desc='written file path'))
+
+class CopyHeader(SimpleInterface):
+    input_spec = CopyHeaderInputSpec
+    output_spec = CopyHeaderOutputSpec
+
+    def _run_interface(self, runtime):
+        in_img = nb.load(self.inputs.hdr_file)
+        out_img = nb.load(self.inputs.in_file, mmap=False)
```
effigies: This is fine, but since we're not overwriting the file, `mmap=False` isn't necessary. (In practice, its value doesn't matter since we're always using `.nii.gz`, which precludes `mmap`ing.)

niworkflows/interfaces/utils.py
```diff
@@ -0,0 +1,36 @@
...
+from nipype.utils.filemanip import fname_presuffix
+from nipype.interfaces.base import File, OutputMultiPath, BaseInterfaceInputSpec, TraitedSpec
+from .base import SimpleInterface
+
+class CopyHeaderInputSpec(BaseInterfaceInputSpec):
+    in_file = File(exists=True, mandatory=True, desc='the file we get the data from')
+    hdr_file = File(exists=True, mandatory=True, desc='the file we get the header from')
+
+class CopyHeaderOutputSpec(TraitedSpec):
+    out_file = OutputMultiPath(File(exists=True, desc='written file path'))
```
effigies: I think this should just be a `File`, not an `OutputMultiPath`, unless I'm missing something?

PR #169, opened 2017-05-19
[RTM] initialize the normalization with AffineInitializer

chrisfilo:


niworkflows/anat/mni.py
```diff
@@ -96,6 +97,11 @@ def _run_interface(self, runtime):
 
             self._config_ants(ants_settings)
 
+            if not isdefined(self.inputs.initial_moving_transform):
```
oesteban: One comment comes up to my mind:

AffineInitializer can be run just once, regardless the parameter set that is then hooked up, if this call was done before https://github.com/chrisfilo/niworkflows/blob/579741a350739fbb8f75fd0dff4e4fc7fb1fd3e7/niworkflows/anat/mni.py#L95.

niworkflows/anat/mni.py
```diff
@@ -131,14 +137,10 @@ def _config_ants(self, ants_settings):
             num_threads=self.inputs.num_threads,
             from_file=ants_settings,
             terminal_output='file',
-            write_composite_transform=True
+            write_composite_transform=True,
+            initial_moving_transform=self.inputs.initial_moving_transform
         )
 
-        if isdefined(self.inputs.initial_moving_transform):
```
oesteban: why don't something like:

```
if init_transform is None and isdefined(...)
```
where `init_transform` is the result of the call to `AffineInitializer.run()` if it was actually run.

chrisfilo:
I moved it outside of the loop, but overridden `self.inputs` for simplicity. Let me know if this works for you.

PR #171, opened 2017-05-25
[FIX] Remove initial_moving_transform_com EPI to MNI settings

oesteban:
Necessary since #169 was merged.


oesteban:
Since this is a minor change necessary for releasing mriqc, I'll merge now.

PR #174, opened 2017-05-31
ENH: Import nipype as submodule

effigies:
First pass at resolving poldracklab/fmriprep#541.

This approach makes a submodule in the root directory and symlinks `niworkflows/nipype` to `../nipype/nipype`. From a little bit of research, the alternative seemed to involve keeping an actual copy of `nipype/nipype` in the source tree, using either `git subtree` or `git filter-branch`. I'm not entirely clear on the details of these approaches, so it's possible that this isn't actually the simplest method.

### DIFF
--- a/niworkflows/info.py
+++ b/niworkflows/info.py
@@ -10,6 +10,11 @@ as well as for open-source software distribution.
 """
 from __future__ import absolute_import, division, print_function
 import datetime
+from os import path as op
+import runpy
+
+nipype_info = runpy.run_path(op.join(op.abspath(op.dirname(__file__)),
+                                     'nipype', 'info.py'))
 
 __version__ = '0.0.8-dev'
 __packagename__ = 'niworkflows'
@@ -39,8 +44,9 @@ CLASSIFIERS = [
     'Programming Language :: Python :: 3.5',
 ]
 
-REQUIRES = ['future', 'nilearn>=0.2.6', 'sklearn', 'pandas', 'matplotlib',
-            'jinja2', 'svgutils', 'tinycss', 'packaging']
+REQUIRES = nipype_info['REQUIRES'] + [
+    'nilearn>=0.2.6', 'sklearn', 'pandas', 'matplotlib', 'jinja2', 'svgutils',
+    'tinycss']
 SETUP_REQUIRES = []
 REQUIRES += SETUP_REQUIRES
 
--- a/niworkflows/info.py
+++ b/niworkflows/info.py
@@ -39,7 +39,7 @@ CLASSIFIERS = [
     'Programming Language :: Python :: 3.5',
 ]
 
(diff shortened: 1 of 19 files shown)


niworkflows/info.py
```diff
@@ -39,7 +39,7 @@
     'Programming Language :: Python :: 3.5',
 ]
 
-REQUIRES = ['nipype', 'future', 'nilearn>=0.2.6', 'sklearn', 'pandas', 'matplotlib',
```
chrisfilo: Do we need to add nipypes dependencies here?

effigies:
Stuck on preventing Circle from running nipype tests.

effigies:
I pinned an MRIQC branch against this one, and we're running into [issues](https://circleci.com/gh/effigies/mriqc/2) where we can't mix `niworkflows.nipype` and `nipype`.

```Python
  File "/usr/local/miniconda/bin/mriqc", line 11, in <module>
    load_entry_point('mriqc===09f08a30af98dd6d73364422eedb4c7265ba93a8', 'console_scripts', 'mriqc')()
  File "/usr/local/miniconda/lib/python3.6/site-packages/mriqc/bin/mriqc_run.py", line 280, in main
    wf_list.append(build_workflow(dataset[mod], mod, settings=settings))
  File "/usr/local/miniconda/lib/python3.6/site-packages/mriqc/workflows/core.py", line 24, in build_workflow
    workflow = anat_qc_workflow(dataset, mod='T1w', settings=settings)
  File "/usr/local/miniconda/lib/python3.6/site-packages/mriqc/workflows/anatomical.py", line 94, in anat_qc_workflow
    to_ras = pe.Node(ConformImage(check_dtype=False), name='conform')
  File "/usr/local/miniconda/lib/python3.6/site-packages/nipype/pipeline/engine/nodes.py", line 159, in __init__
    raise IOError('interface must be an instance of an Interface')
OSError: interface must be an instance of an Interface
```

It could be we don't care; if you use `niworkflows`, you commit to using `niworkflows.nipype`. So we would just need to update MRIQC before we make it depend on an updated niworkflows.

But if we want interoperability, this could be a dealbreaker. `isinstance` is going to be specific to each class hierarchy, even if they're equivalent.

chrisfilo:
I don't see why we would want to mix.

effigies:
Not that we would want to mix, but if other people wanted to incorporate some of our interfaces into existing nipype pipelines without porting all of their code to `niworkflows.nipype`. But I think it makes more sense to focus on upstreaming interfaces that are generally useful outside of our projects.

I'm okay with breaking mixing, as long as we know that's what we're doing. This is ready to go.

chrisfilo:
Great! Could you open a PR on fmriprep with using this branch so we would see how well this new approach integrates?

effigies:
poldracklab/fmriprep#549

PR #176, opened 2017-06-06
FIX: Pass runtime to aggregate_outputs

effigies:
`FLIRT(save_log=True)` needs access to `runtime.stdout`. It seems like a good idea to pass `runtime` in all cases.


oesteban:
This is safe merging, right?

effigies:
Yes.

PR #187, opened 2017-07-22
[ENH] Do not enforce float precision for ANTs

oesteban:
This PR allows using ANTs' default (double precision) and also adds a `float` input in `RobustMNINormalization` to allow for changing it.

### DIFF
--- a/niworkflows/anat/mni.py
+++ b/niworkflows/anat/mni.py
@@ -57,6 +57,7 @@ class RobustMNINormalizationInputSpec(BaseInterfaceInputSpec):
                                         "Requires reliable and accurate masks."
                                         "See https://sourceforge.net/p/advants/discussion/840261/thread/27216e69/#c7ba")
     initial_moving_transform = File(exists=True, desc='transform for initialization')
+    float = traits.Bool(False, usedefault=True, desc='use single precision calculations')
 
 
 class RobustMNINormalization(BaseInterface):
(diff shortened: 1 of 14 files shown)


oesteban:
@effigies @chrisfilo any idea against merging this in? (I'll submit a PR to fmriprep to set it back to float=True as it was before)

effigies:
I'm not a huge fan of "float" as a variable name, since it's a keyword, but since that seems to be the pattern in the ANTs interfaces, it'd be more confusing to change it and have a mismatch. So I'm okay with this as is.

If it's not too urgent to get this in so you can move forward with MRIQC, can you pin to the latest nipype master, after nipy/nipype#2129 went in, and push an fmriprep PR off this branch, just to make sure it's not breaking before the merge?

If it needs to go in quickly, that's fine, too.

oesteban:
@effigies how do I pin nipype to 57150052449a2f5ddead061b4ac2ba918161092c, is it done in the `.gitmodules` file?

effigies:
Should add this to docs somewhere...

Update `.gitmodules` to point to the right repo (no change here, but if you're pinning to a feature branch you might need to do your own repo). Then (adapted from https://github.com/poldracklab/fmriprep/pull/539#issuecomment-309210055):

```Shell
git submodule init
git submodule update
git -C nipype fetch origin
git -C nipype checkout <ref>
git add .gitmodule nipype
git commit -m 'PIN: Update to nipype/master'
```

PR #192, opened 2017-08-02
[RTM] take TR units into account

### LINKED ISSUE #191: Melodic report ignores TR units
https://github.com/poldracklab/niworkflows/blob/7594df0f4a7bb054e51e2d1f31586017c3cade81/niworkflows/viz/utils.py#L505

Those can be seconds or msec. There is a field for it in the nifti header.

chrisfilo:
Closes #191.

### DIFF
--- a/niworkflows/viz/utils.py
+++ b/niworkflows/viz/utils.py
@@ -504,8 +504,17 @@ def plot_melodic_components(melodic_dir, in_file, tr=None,
     if not tr:
         tr = in_nii.header.get_zooms()[3]
         units = in_nii.header.get_xyzt_units()
-        if units and units[-1] == 'msec':
-            tr = tr/1000.0
+        if units:
+            if units[-1] == 'msec':
+                tr = tr / 1000.0
+            elif units[-1] == 'usec':
+                tr = tr / 1000000.0
+            elif units[-1] != 'sec':
+                NIWORKFLOWS_LOG.warn('Unknown repetition time units '
+                                     'specified - assuming seconds')
+        else:
+            NIWORKFLOWS_LOG.warn(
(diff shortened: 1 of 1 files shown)


niworkflows/viz/utils.py
```diff
@@ -503,6 +503,9 @@ def plot_melodic_components(melodic_dir, in_file, tr=None,
     in_nii = nb.load(in_file)
     if not tr:
         tr = in_nii.header.get_zooms()[3]
+        units = in_nii.header.get_xyzt_units()
+        if units and units[-1] == 'msec':
+            tr = tr/1000.0
```
effigies: Is it actually possible for `units` to evaluate to `False`? Do we want to handle `'usec'` units, for completeness? And should we warn on unknown units? It's almost always seconds, but the warning might help us track down the issue, if it's causing one.

So I might rewrite as:

```Python
_, t_unit = in_nii.header.get_xyzt_units()
if t_unit == 'unknown':
    warnings.warn('Unknown t units, assuming seconds.')
    t_unit = 'sec'
tr /= 10.0 ** {'sec': 0, 'msec': 3, 'usec': 6}[t_unit]
```

Would also be okay to just put spaces around `/`.

effigies:
Build timed out. Can you rebuild?

chrisfilo:
Tried couple of times. Not sure why is it timing out.

effigies:
Really not sure. The commit you've branched off of didn't have problems. It might have been a temporary circle bug. I've pushed an empty commit so it'll run under my account and I can try to debug, if it keeps happening. If https://circleci.com/gh/effigies/niworkflows/146 passes, I'll go ahead and merge, here.

PR #203, opened 2017-10-12
General cleanup

oesteban:


niworkflows/viz/validators.py
```diff
@@ -18,7 +21,7 @@ def validate(self, css):
         stylesheet = self.parser.parse_stylesheet(css)
         for rule in stylesheet.rules:
             self.validate_no_fixed_position(rule)
-        if not stylesheet.errors is None and len(stylesheet.errors) > 0:
+        if stylesheet.errors is not None and len(stylesheet.errors) > 0:
```
effigies: This seems to be equivalent to:

```Python
if stylesheet.errors:
```

effigies:
Good to merge?

oesteban:
yess, kind of urgent actually. MRIQC is failing.

Have a look to the move I've made: mni from niworkflows.anat to
niworkflows.interfaces

effigies:
Looks reasonable. Do you want to build MRIQC/FMRIPREP against this branch before merge? (I'm okay merging ASAP.)

oesteban:
I know MRIQC works because I've been checking locally. I'd say we should go
ahead this time.

PR #204, opened 2017-10-13
ENH: Add contours to MRICoregRPT report

effigies:
See https://github.com/poldracklab/fmriprep/issues/758.

### DIFF
--- a/niworkflows/interfaces/registration.py
+++ b/niworkflows/interfaces/registration.py
@@ -226,13 +226,13 @@ class MRICoregRPT(nrc.RegistrationRC, fs.MRICoreg):
 
     def _post_run_hook(self, runtime):
         outputs = self.aggregate_outputs(runtime=runtime)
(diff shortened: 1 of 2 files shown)


effigies:
@oesteban Look reasonable?

niworkflows/interfaces/registration.py
```diff
@@ -243,6 +245,8 @@ def _post_run_hook(self, runtime):
 
         self._fixed_image = target_file
         self._moving_image = res.outputs.transformed_file
+        if mri_dir is not None:
+            self._contour = os.path.join(mri_dir, 'ribbon.mgz')
```
oesteban: Do our reports work with mgz files? Is there a way to add a contour to the test we have right now?

effigies: Yes, it's just a nibabel load. See `BBRegisterRPT`.

PR #205, opened 2017-10-23
[ENH] Optional crop-to-FoV added into GenerateSamplingReference

oesteban:
Should address poldracklab/fmriprep#722

Adds a `fov_mask` input to `GenerateSamplingReference`. This mask corresponds to the brain mask calculated from the T1w, and should not contain the brain stem.

The mask is resampled into target space, and then the bounding box is calculated. Finally, the FoV is adjusted to that bounding box.


oesteban:
This should be ready to review and merge 👍

chrisfilo:
I would recommend testing it on one of the OpenfMRI limited FOV datasets to
see check that it behaves correctly in extreme cases.

Best,
Chris

PR #207, opened 2017-10-26
[FIX] Increase tolerance in GenerateSamplingReference

oesteban:
Addresses poldracklab/fmriprep#800

### DIFF
--- a/niworkflows/interfaces/utils.py
+++ b/niworkflows/interfaces/utils.py
@@ -202,8 +202,17 @@ def _gen_reference(fixed_image, moving_image, fov_mask=None, out_file=None,
         # out of the FoV.
         fixednii = nb.load(fixed_image)
         masknii = nb.load(fov_mask)
-        assert np.allclose(masknii.affine, fixednii.affine)
-        assert np.all(masknii.shape[:3] == fixednii.shape[:3])
+
+        if np.all(masknii.shape[:3] != fixednii.shape[:3]):
+            raise RuntimeError(
+                'Fixed image and mask do not have the same dimensions.')
+
+        if not np.allclose(masknii.affine, fixednii.affine, atol=1e-5):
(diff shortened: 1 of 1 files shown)


niworkflows/interfaces/utils.py
```diff
@@ -202,8 +202,17 @@ def _gen_reference(fixed_image, moving_image, fov_mask=None, out_file=None,
         # out of the FoV.
         fixednii = nb.load(fixed_image)
         masknii = nb.load(fov_mask)
-        assert np.allclose(masknii.affine, fixednii.affine)
-        assert np.all(masknii.shape[:3] == fixednii.shape[:3])
+
+        if np.all(masknii.shape[:3] != fixednii.shape[:3]):
```
effigies: Just a quick note: `img.shape[:3]` is a tuple, whose equality is all-or-nothing. So this is just `masknii.shape[:3] != fixednii.shape[:3]`.

Not worth cleaning up right now, but perhaps in passing at some point.

niworkflows/interfaces/utils.py
```diff
@@ -202,8 +202,17 @@ def _gen_reference(fixed_image, moving_image, fov_mask=None, out_file=None,
...
+            raise RuntimeError(
+                'Fixed image and mask do not have the same dimensions.')
+
+        if not np.allclose(masknii.affine, fixednii.affine, atol=1e-5):
+            raise RuntimeError(
+                'Fixed image and mask have different affines')
+
+        # Mask affine should be
+        masknii.affine = fixednii.affine
```
effigies: This, on the other hand, should raise an error. Are you not having issues depending on this behavior? `img.affine` is a read-only property.

oesteban:
Yes, I fixed it directly in master. Resetting the mask affine wasn't even necessary

PR #209, opened 2017-11-16
[ENH] Lower priority of "Affines do not match" warning

oesteban:
See poldracklab/fmriprep#837


effigies:
I think it might be good to phrase the warning a bit less definitively, too. Possibly indicating that users should be sure to verify outputs align as expected.

PR #210, opened 2017-12-01
[ENH] Generate svgs only

oesteban:
Make it easier for FMRIPREP to strip svgs from the reports.
HTML files will be used for reportlets (like the summary) that
can be completelly inlined. Otherwise, expect standalone svgs.


chrisfilo:
Could you open WIP PRs for MRIQC and FMRIPREP using this branch so we would see how this is working?

oesteban:
There already (poldracklab/fmriprep#867):

- https://435-53608443-gh.circle-artifacts.com/0/home/ubuntu/ds054/out/fmriprep/sub-100185.html
- https://435-53608443-gh.circle-artifacts.com/1/home/ubuntu/ds005/out/fmriprep/sub-01.html

chrisfilo:
So does this help with large reports that previously were slow to load or crashed browsers?

oesteban:
The load time is definitely shorter. I can't claim though that this solves the problem until I see a huge dataset like MyConnectome using this. But that would be an issue of fmriprep, not in niworkflows.

This PR modifies niworkflows:
  - to make it more consistent (ends the dichotomy of registration vs. segmentation reports or svg vs. html respectively),
  - to reduce the codebase and simplify functionality (we can stop caring about html validity, etc)
  - to simplify the tests of report-capable segmentations by leveraging py.tests, as we previously did with tests if registration interfaces.

chrisfilo:
Yes it does seem like a major refactor. A bit too much for me to dig through right now, but if it works and does not break MRIQC and FMRIPREP you should be good to go.

PR #211, opened 2017-12-02
add new ROIsPlot interface, with test

oesteban:


oesteban:
Just adds one new report like this (generated by the corresponding test): https://215-74174878-gh.circle-artifacts.com/0/home/ubuntu/scratch/py3/testROIsPlot.svg

oesteban:
This is working, just adds a new type of plots: https://223-74174878-gh.circle-artifacts.com/0/home/ubuntu/scratch/py3/testROIsPlot.svg

There are some changes to the general way Segmentation interfaces work, but I've checked they are producing the expected results.

niworkflows/tests/test_segmentation.py
```diff
@@ -28,6 +29,31 @@ def test_BETRPT(moving):
...
+    """ the BET report capable test """
+    import nibabel as nb
+    import numpy as np
+
+    labels = nb.load(os.path.join(oasis_dir, 'T_template0_glm_4labelsJointFusion.nii.gz'))
+    data = labels.get_data()
+    out_files = []
+    ldata = np.zeros_like(data)
+    for i, l in enumerate([1, 3, 4, 2]):
+        ldata[data == l] = 1
```
effigies: Shouldn't `ldata` be defined in the loop? Otherwise each subsequent ROI contains the previous one.

niworkflows/viz/utils.py
```diff
@@ -87,12 +88,10 @@ def _run(args, input=None, stdout=None, stderr=None, shell=False, check=False,
...
 
 def robust_set_limits(data, plot_params):
-    vmin = np.percentile(data, 15)
-    if plot_params.get('vmin', None) is None:
-        plot_params['vmin'] = vmin
-    if plot_params.get('vmax', None) is None:
-        plot_params['vmax'] = np.percentile(data[data > vmin], 99.8)
-
+    plot_params['vmin'] = plot_params.get(
+        'vmin', np.percentile(data, 15))
```
chrisfilo: Wouldn't this change how a lot of figures are plotter? Have you checked if there are no side effects?

niworkflows/tests/test_segmentation.py
oesteban: That's to make the masks nested. The second includes the first and so on so forth.

oesteban:
1 - The colors are assigned "smartly". You provide colors, but if there are more contours than colors, then a palette from seaborn is assigned. The original idea is that you can set a red contour to the brainmask (first contour) and then let seaborn add nice colors. The example you linked is with `colors=['g']` only. For testing purposes, I set this: https://github.com/poldracklab/niworkflows/pull/211/files#diff-422d820d69635000670ff27c02411928R52. I can roll back the test to the original situation by just setting `colors=['g']`, but since this is not hard-coded, I think it does not make much of a difference.

2 - no more report_title. Updated in all interfaces. This is a result of transitioning to SVG, let's let fmriprep set this in the reports configuration.

niworkflows/viz/utils.py
oesteban: I haven't seen any problem, but the logic of this code should be equivalent to the former.

niworkflows/tests/test_segmentation.py
oesteban: And the masks are nested to avoid overlapping contours...

effigies: Okay.

PR #214, opened 2017-12-14
[RTM] ENH: Use nearest for coreg visualization

effigies:
This will add an `mri_vol2vol` call for each BOLD series in FMRIPREP, but it should produce better coregistration images without smoothing, which can make assessing the alignment of gray-white boundaries difficult.


effigies:
See https://github.com/poldracklab/fmriprep/pull/889#issuecomment-351824756.

PR #220, opened 2018-02-07
Make retry information more clear and fix a bug failing any retry

chrisfilo:


chrisfilo:
We should niworkflows release after this is merged as well as mriqc and fmriprep.

niworkflows/interfaces/mni.py
```diff
@@ -76,7 +76,7 @@ def _list_outputs(self):
 
     def __init__(self, **inputs):
         self.norm = None
-        self.retry = 0
+        self.retry = 1
```
effigies: Where is the number of retries defined? Does this mean we do 1 fewer iterations, or is it purely a logging change?

chrisfilo: Each retry comes with its own settings and the number of retries is defined by the set of distinct settings to try.

chrisfilo:
Previous version of the code would try to use uninitialized variable leading to errors such as https://github.com/poldracklab/mriqc/issues/684

niworkflows/interfaces/mni.py
effigies: Okay, I'm seeing the settings here: https://github.com/poldracklab/niworkflows/tree/master/niworkflows/data

But this `retry` attribute is just a counter for logging, and has nothing to do with indexing those files, right?

PR #221, opened 2018-02-10
Adding SanitizeImage

chrisfilo:
This is a port (with modifications) of ValidateImage from FMRIPREP. Modifications to the ValidateImage interface are best viewed using this link https://github.com/poldracklab/niworkflows/compare/132558efe2013588fb2135c464405b16234dd84e...chrisfilo:a9e0d4a4769ddb252d9f99fee6d1930d35a672bf

### DIFF
--- a/.circle/pytests.sh
+++ b/.circle/pytests.sh
@@ -11,8 +11,7 @@ set -x         # Print command traces before executing command.
 code=0
 
 if [ "${CIRCLE_NODE_TOTAL:-1}" == "1" ]; then
-    docker run -ti --env SAVE_CIRCLE_ARTIFACTS="/scratch" -v ${SCRATCH}/py3:/scratch -w /scratch --entrypoint=/usr/local/miniconda/bin/py.test niworkflows:py3 /root/niworkflows -n ${CIRCLE_NPROCS:-4} -v --junit-xml=/scratch/pytest.xml && \
-    docker run -ti --env SAVE_CIRCLE_ARTIFACTS="/scratch" -v ${SCRATCH}/py2:/scratch -w /scratch --entrypoint=/usr/local/miniconda/bin/py.test niworkflows:py2 /root/niworkflows -n ${CIRCLE_NPROCS:-4} -v --junit-xml=/scratch/pytest.xml
+    docker run -ti --env SAVE_CIRCLE_ARTIFACTS="/scratch" -v ${SCRATCH}/py3:/scratch -w /scratch --entrypoint=/usr/local/miniconda/bin/py.test niworkflows:py3 /root/niworkflows -n ${CIRCLE_NPROCS:-4} -v --junit-xml=/scratch/pytest.xml
     code=$(( $code + $? ))
 else
     case ${CIRCLE_NODE_INDEX} in
@@ -20,15 +19,10 @@ else
         docker run --rm=false -ti --env SAVE_CIRCLE_ARTIFACTS="/scratch" --env CIRCLE_NPROCS=4 -v ${SCRATCH}/py3:/scratch -w /scratch --entrypoint=/usr/local/miniconda/bin/py.test niworkflows:py3 /root/niworkflows  -n 2 -v --junit-xml=/scratch/pytest.xml
         code=$(( $code + $? ))
         ;;
-    1)
-        docker run --rm=false -ti --env SAVE_CIRCLE_ARTIFACTS="/scratch" --env CIRCLE_NPROCS=4 -v ${SCRATCH}/py2:/scratch -w /scratch --entrypoint=/usr/local/miniconda/bin/py.test niworkflows:py2 /root/niworkflows  -n 2 -v --junit-xml=/scratch/pytest.xml
-        code=$(( $code + $? ))
-        ;;
     esac
 fi
 
 # Place xml files where appropriate
 cp $SCRATCH/py3/*.xml ${CIRCLE_TEST_REPORTS}/py3/  || true
-cp $SCRATCH/py2/*.xml ${CIRCLE_TEST_REPORTS}/py2/  || true
 
 exit $code
--- a/.circle/pytests.sh
+++ b/.circle/pytests.sh
@@ -10,17 +10,8 @@ set -x         # Print command traces before executing command.
 
 code=0
 
-if [ "${CIRCLE_NODE_TOTAL:-1}" == "1" ]; then
-    docker run -ti --env SAVE_CIRCLE_ARTIFACTS="/scratch" -v ${SCRATCH}/py3:/scratch -w /scratch --entrypoint=/usr/local/miniconda/bin/py.test niworkflows:py3 /root/niworkflows -n ${CIRCLE_NPROCS:-4} -v --junit-xml=/scratch/pytest.xml
-    code=$(( $code + $? ))
-else
-    case ${CIRCLE_NODE_INDEX} in
-    0)
-        docker run --rm=false -ti --env SAVE_CIRCLE_ARTIFACTS="/scratch" --env CIRCLE_NPROCS=4 -v ${SCRATCH}/py3:/scratch -w /scratch --entrypoint=/usr/local/miniconda/bin/py.test niworkflows:py3 /root/niworkflows  -n 2 -v --junit-xml=/scratch/pytest.xml
-        code=$(( $code + $? ))
-        ;;
-    esac
-fi
+docker run -ti --env SAVE_CIRCLE_ARTIFACTS="/scratch" -v ${SCRATCH}/py3:/scratch -w /scratch --entrypoint=/usr/local/miniconda/bin/py.test niworkflows:py3 /root/niworkflows -n ${CIRCLE_NPROCS:-4} -v --junit-xml=/scratch/pytest.xml
+code=$(( $code + $? ))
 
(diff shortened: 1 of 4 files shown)


chrisfilo:
Sanitizer in action https://github.com/poldracklab/mriqc/pull/687

chrisfilo:
Fails due to python 2 compatibility. Do we still support python 2?

oesteban:
This is the issue - [`textwrap.indent` was included in python 3.3](https://docs.python.org/3/library/textwrap.html#textwrap.indent):

```
/root/niworkflows/niworkflows/interfaces/utils.py:15: in <module>
    from textwrap import indent
E   ImportError: cannot import name indent
```

I think we can drop python 2. We were waiting for nipype 2.0 to drop it in niworkflows, but I'd say we should stop that effort. We could also shorten the 2h long builds if we don't test python 2 anymore.

.circle/pytests.sh
```diff
@@ -11,24 +11,18 @@ set -x         # Print command traces before executing command.
...
     0)
         docker run --rm=false -ti --env SAVE_CIRCLE_ARTIFACTS="/scratch" --env CIRCLE_NPROCS=4 -v ${SCRATCH}/py3:/scratch -w /scratch --entrypoint=/usr/local/miniconda/bin/py.test niworkflows:py3 /root/niworkflows  -n 2 -v --junit-xml=/scratch/pytest.xml
         code=$(( $code + $? ))
         ;;
-    1)
-        docker run --rm=false -ti --env SAVE_CIRCLE_ARTIFACTS="/scratch" --env CIRCLE_NPROCS=4 -v ${SCRATCH}/py2:/scratch -w /scratch --entrypoint=/usr/local/miniconda/bin/py.test niworkflows:py2 /root/niworkflows  -n 2 -v --junit-xml=/scratch/pytest.xml
-        code=$(( $code + $? ))
-        ;;
     esac
 fi
```
effigies: Looks like the whole if/else structure can be dropped now, and CircleCI reduced to 1x parallelism?

niworkflows/interfaces/utils.py
```diff
@@ -256,3 +257,150 @@ def _gen_reference(fixed_image, moving_image, fov_mask=None, out_file=None,
...
+    The qform and sform matrices have been set to a default, LAS-oriented affine.
+    Analyses of this dataset MAY BE INVALID.
+</p>
+"""
+
+        if (self.inputs.force_float32 and np.dtype(img.get_data_dtype()).itemsize > 4) or self.inputs.n_volumes_to_discard:
+            in_data = img.get_data()
+
+            if (self.inputs.force_float32 and np.dtype(img.get_data_dtype()).itemsize > 4):
+                in_data = in_data.astype(np.float32)
```
effigies: How about replacing the previous four lines with:


```Python
in_data = img.get_fdata(dtype=np.float32) if self.inputs.force_float32 else img.dataobj
```

For large datasets, this could save us some memory.

This bumps our nibabel dependency up to 2.2.0, but we're going to need that if we want to start doing CIFTI outputs, anyway.

niworkflows/interfaces/utils.py
```diff
@@ -256,3 +257,150 @@ def _gen_reference(fixed_image, moving_image, fov_mask=None, out_file=None,
...
+
+            img = nb.Nifti1Image(in_data[:, :, :, self.inputs.n_volumes_to_discard:],
+                                     img.affine,
+                                     img.header)
```
effigies: Align with open paren.

niworkflows/interfaces/utils.py
```diff
@@ -256,3 +257,150 @@ def _gen_reference(fixed_image, moving_image, fov_mask=None, out_file=None,
...
+            warning_txt = 'WARNING - Missing orientation information'
+            description = """\
+<p class="elem-desc">
+    Orientation information could not be retrieved from the image header.
+    The qform and sform matrices have been set to a default, LAS-oriented affine.
+    Analyses of this dataset MAY BE INVALID.
+</p>
+"""
+
+        if (self.inputs.force_float32 and np.dtype(img.get_data_dtype()).itemsize > 4) or self.inputs.n_volumes_to_discard:
```
effigies: I don't fully understand the `np.dtype(img.get_data_dtype()).itemsize > 4`. So if we have `uint16` inputs and `force_float32`, we will have `uint16` outputs? That seems counterintuitive.

niworkflows/interfaces/utils.py
```diff
@@ -256,3 +257,150 @@ def _gen_reference(fixed_image, moving_image, fov_mask=None, out_file=None,
...
+                                     img.affine,
+                                     img.header)
+            save_file = True
+
+        # Store new file and report
+        if save_file:
+            snippet = '<h3 class="elem-title">%s</h3>\n%s\n' % (
+            warning_txt, description)
+            with open(out_report, 'w') as fobj:
+                fobj.write(indent(snippet, '\t' * 3))
```
effigies: If you don't have a warning_txt and description (i.e. the file was fine, but you wanted to coerce to float32), this will fail.

niworkflows/interfaces/utils.py
chrisfilo: `dataobj` is cool when you don't need to read all of the data, but since we need save the whole file would there be any gains?

effigies: Sure, you can use `get_data()`. The only advantage to `dataobj` here is that if you're dropping some volumes (which is the case when you aren't forcing float), you don't have to load them into memory first. It's going to be a pretty small savings, so I don't feel strongly here.

The main point was the `get_fdata` which lets nibabel do the work of coercing the type while keeping memory usage under control.

```Python
# Loads full file and fills cache
in_data = img.get_data()
# Makes new array, old one still around until `img` is garbage collected
in_data = in_data.astype(np.float32)
```

niworkflows/interfaces/utils.py
chrisfilo: Your interpretation is correct - as described in the input description. I left it this way because this is the way things have been done in the sanitizer function in MRIQC this interface is intended to replace.

niworkflows/interfaces/utils.py
chrisfilo: Your interpretation is correct - as described in the input description. I left it this way because this is the way things have been done in the sanitizer function in MRIQC this interface is intended to replace.

niworkflows/interfaces/utils.py
chrisfilo: nice catch!

niworkflows/interfaces/utils.py
oesteban: Yep, this does not make much sense. We should probably correct this (do not check `itemsize`, basically).

chrisfilo: The idea is that you only want to cast when the dtype is larger than 32 bit. Basically we don't want to cast int16 to float32 (which is what will happen if you remove this check). This is how MRIQC has been operating so far (see https://github.com/poldracklab/mriqc/blob/1ba709a2fa31e71e14d249b8db2fd87d5b349f53/mriqc/utils/misc.py#L55). Maybe this code can make things more clear:
```
In [14]: np.array(23,dtype=np.int16).dtype.itemsize
Out[14]: 2

In [15]: np.array(23,dtype=np.float32).dtype.itemsize
Out[15]: 4

In [16]: np.array(23,dtype=np.float64).dtype.itemsize
Out[16]: 8
```

effigies: I see now that it's in the description, so that's good. But maybe it'd be better to rename the variable to be a little more congruent with its function.

Some options:

* `max_32bit`
* `disable_64bit`
* `allow_float64` (inverting logic)
* `allow_64bit` (inverting logic)

I don't think any of these are great, but I think they'll reduce the double-take factor.

chrisfilo:
I would not disagree with this.

Best,
Chris

.circle/pytests.sh
```diff
@@ -10,25 +10,10 @@ set -x         # Print command traces before executing command.
...
-        code=$(( $code + $? ))
-        ;;
-    1)
-        docker run --rm=false -ti --env SAVE_CIRCLE_ARTIFACTS="/scratch" --env CIRCLE_NPROCS=4 -v ${SCRATCH}/py2:/scratch -w /scratch --entrypoint=/usr/local/miniconda/bin/py.test niworkflows:py2 /root/niworkflows  -n 2 -v --junit-xml=/scratch/pytest.xml
-        code=$(( $code + $? ))
-        ;;
-    esac
-fi
+docker run -ti --env SAVE_CIRCLE_ARTIFACTS="/scratch" -v ${SCRATCH}/py3:/scratch -w /scratch --entrypoint=/usr/local/miniconda/bin/py.test niworkflows:py3 /root/niworkflows -n ${CIRCLE_NPROCS:-4} -v --junit-xml=/scratch/pytest.xml
+code=$(( $code + $? ))
```
effigies: I find the whole exit code thing here a little strange. We have `set -e` in the preamble, so if `$?` is non-zero, we won't even get to this line. I think we can drop the whole `code` variable altogether.

@oesteban I think you wrote this script. Am I missing something?

oesteban: Yes, that's right. Anyways it makes less sense now.

PR #224, opened 2018-02-14
[TST] Migrate to CircleCI 2.0

oesteban:


oesteban:
This is ready to merge. The most relevant change is that fmriprep's docker image is not pulled anymore. The runtime is around the same it was before with 1.0 (1h 30min).

PR #226, opened 2018-02-22
Remove extensions from the nifti header

chrisfilo:
Necessary to fix https://github.com/poldracklab/mriqc/issues/693


effigies:
Looks fine. Only concern is (if we use this in fmriprep) if the AFNI headers should be preserved to include in the output time series. In MRIQC it's irrelevant, so this is fine in that context.

chrisfilo:
I don't think they actually are right now due to all the splitting and applying transformation using ants. This might have to be done explicitly.

effigies:
Fair point. Let's drop it here then, and we can recover through a separate path, later, if desired. Might be worth having someone who uses AFNI chime in with whether restoring the headers to the preprocessed BOLD series would be useful to them.

PR #229, opened 2018-03-14
[ENH] Migrate the summary plot to niworkflows

oesteban:
Updated look and feel (basically removing distracting elements). The colorbar of the carpetplot is:

blue=cortical regions
orange = deep gray matter
green = cerebellum
red = white and csf

newfmriprepsummary


oesteban:
Further iterations over this will include reading in a confounds file (instead of passing independent text files) and allowing for a list of labels to select which confounds are being plotted. The idea behind this is fixing poldracklab/fmriprep#174 and poldracklab/fmriprep#877 in one shot.

PR #230, opened 2018-03-22
[ENH] fMRI summary plot to take ``_confounds.tsv``

oesteban:
Also updates the wrapper interface for MRIQC to keep working
Preparing for poldracklab/fmriprep#174 and poldracklab/fmriprep#877

### DIFF
--- a/niworkflows/interfaces/plotting.py
+++ b/niworkflows/interfaces/plotting.py
@@ -6,6 +6,7 @@ Visualization tools
 
 """
 import numpy as np
+import pandas as pd
 
 from ..nipype.utils.filemanip import fname_presuffix
 from ..nipype.interfaces.base import (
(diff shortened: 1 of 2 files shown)


oesteban:
This is how they look now:

newfmrisummary

effigies:
That does look nice, and I think they give a good gist for people familiar with the metrics, but I think it would be good also to add a moderately detailed "How to read" section in the MRIQC/fMRIPrep docs.

That's not an issue for this PR, but just a heads up for the apps that use this.

oesteban:
Working already on that howto. Agreed on that this plot should be explained there (https://1387-54756129-gh.circle-artifacts.com/1/home/ubuntu/scratch/docs/interpreting.html)

PR #232, opened 2018-05-03
[FIX] ValueError: invalid literal for int()

oesteban:
While creating the ICA-AROMA reports.


niworkflows/viz/utils.py
```diff
@@ -549,7 +550,7 @@ def plot_melodic_components(melodic_dir, in_file, tr=None,
         row = int(i / 2)
         l_row = row * 2
 
-        if noise_components_file:
+        if noise_components:
```
effigies: `noise_components.size > 0`

PR #233, opened 2018-05-04
[ENH] Constrained cost-function masking for T1-MNI registration

effigies:
Starting by manually rebasing #121 onto current master

Closes #121.

### DIFF
--- a/niworkflows/interfaces/mni.py
+++ b/niworkflows/interfaces/mni.py
@@ -441,6 +441,8 @@ def create_cfm(in_file, lesion_mask=None, global_mask=True, out_path=None):
         data = np.ones(in_img.shape, dtype=np.uint8)
     else:
         data = in_img.get_data()
+        if set(np.unique(data)) - {0, 1}:
+            raise ValueError("`global_mask` must be true if `in_file` is not a binary mask")
 
     # If a lesion mask was provided, combine it with the secondary mask.
     if lesion_mask is not None:
--- a/niworkflows/interfaces/mni.py
+++ b/niworkflows/interfaces/mni.py
@@ -225,8 +225,7 @@ class RobustMNINormalization(BaseInterface):
                 # Use this as the moving mask.
                 args['moving_image_mask'] = create_cfm(
                     self.inputs.moving_mask,
-                    "moving_cfm.nii.gz",
-                    self.inputs.lesion_mask,
+                    lesion_mask=self.inputs.lesion_mask,
                     global_mask=self.inputs.explicit_masking)
 
         # If no moving mask is provided...
@@ -236,8 +235,7 @@ class RobustMNINormalization(BaseInterface):
             # Use this as the moving mask.
             args['moving_image_mask'] = create_cfm(
                 self.inputs.moving_image,
-                "moving_cfm.nii.gz",
-                self.inputs.lesion_mask,
+                lesion_mask=self.inputs.lesion_mask,
                 global_mask=True)
 
         """
@@ -265,7 +263,6 @@ class RobustMNINormalization(BaseInterface):
                         # Use this as the fixed mask.
                         args['fixed_image_mask'] = create_cfm(
                             self.inputs.reference_mask,
-                            "fixed_cfm.nii.gz",
                             lesion_mask=None,
                             global_mask=True)
 
@@ -283,7 +280,6 @@ class RobustMNINormalization(BaseInterface):
                 # Use this as the fixed mask
                 args['fixed_image_mask'] = create_cfm(
                     self.inputs.reference_image,
-                    "fixed_cfm.nii.gz",
                     lesion_mask=None,
                     global_mask=True)
 
@@ -314,7 +310,6 @@ class RobustMNINormalization(BaseInterface):
                     # Use this as the fixed mask.
                     args['fixed_image_mask'] = create_cfm(
                         op.join(mni_template, '%dmm_brainmask.nii.gz' % resolution),
-                        "fixed_cfm.nii.gz",
                         lesion_mask=None,
                         global_mask=True)
 
@@ -396,7 +391,7 @@ def mask(in_file, mask_file, new_name):
     return os.path.abspath(new_name)
 
 
-def create_cfm(in_file, out_path, lesion_mask, global_mask=True):
+def create_cfm(in_file, lesion_mask=None, global_mask=True, out_path=None):
     """
     Create a mask to constrain registration.
 
@@ -425,6 +420,12 @@ def create_cfm(in_file, out_path, lesion_mask, global_mask=True):
     import os
     import numpy as np
     import nibabel as nb
+    from nipype.utils.filemanip import fname_presuffix
+
+    if out_path is None:
+        out_path = fname_presuffix(in_file, suffix='_cfm', newpath=os.getcwd())
+    else:
+        out_path = os.path.abspath(out_path)
 
     # Load the input image
     in_img = nb.load(in_file)
@@ -458,7 +459,7 @@ def create_cfm(in_file, out_path, lesion_mask, global_mask=True):
     cfm_img.set_data_dtype(np.uint8)
     cfm_img.to_filename(out_path)
 
-    return os.path.abspath(out_path)
+    return out_path
 
 
 def _write_outputs(runtime, out_fname=None):
--- a/niworkflows/interfaces/mni.py
+++ b/niworkflows/interfaces/mni.py
@@ -344,14 +344,14 @@ class RobustMNINormalization(BaseInterface):
             # Set the template resolution.
             resolution = self.inputs.template_resolution
 
+            _template_fmt = op.join(mni_template, '%dmm_%s.nii.gz')
             # If explicit masking is enabled...
             if self.inputs.explicit_masking:
                 # Mask the template image with the template mask.
                 # Do not use a fixed mask during registration.
-                _template_fmt = op.join(mni_template, '%dmm_%s.nii.gz')
                 args['fixed_image'] = mask(
                         _template_fmt % (resolution, self.inputs.reference),
-                        _template_fmt % (resolution, 'brainmask.nii.gz'),
+                        _template_fmt % (resolution, 'brainmask'),
                         "fixed_masked.nii.gz")
 
                 # If a lesion mask is provided...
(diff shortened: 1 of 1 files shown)


effigies:
9d3c25c addresses https://github.com/poldracklab/niworkflows/pull/121#discussion_r131733989.

effigies:
Last three commits address @oesteban's other comments.

@danlurie and I decided to leave some explicit restating of defaults (`lesion_mask=None, global_mask=True`) for the sake of reading.

niworkflows/interfaces/mni.py
```diff
@@ -231,18 +354,113 @@ def _validate_results(self):
...
+
+    # Load the input image
+    in_img = nb.load(in_file)
+
+    # If we want a global mask, create one based on the input image.
+    if global_mask:
+        # Create a mask of ones with the shape of the input image.
+        data = np.ones(in_img.shape, dtype=np.uint8)
+    else:
+        data = in_img.get_data()
```
effigies: I think this is assuming that the data in `in_img` is all 0 or 1. Should we make this:

```Python
data = np.asanyarray(in_img.get_data() > 0, dtype=np.uint8)
```

effigies:
Modulo the comment above, I think we're ready to move on to fmriprep integration.

@oesteban if you have a few minutes, do you want to make a quick look over this? Having worked on this PR before, you might see something I've missed.

niworkflows/interfaces/mni.py
```diff
@@ -231,18 +354,115 @@ def _validate_results(self):
...
 def mask(in_file, mask_file, new_name):
+    """
+    Apply a binary mask to an image.
+
+    Parameters
+    ----------
+    in_file : str
+        Path to a NIfTI file to mask
+    mask_file : str
+        Path to a binary mask
```
oesteban:
```
Path to a binary mask. Dimensions (including time) should match those of ``in_file``
```

niworkflows/interfaces/mni.py
```diff
@@ -231,18 +354,115 @@ def _validate_results(self):
...
     import nibabel as nb
     import os
-
+    # Load the input image
     in_nii = nb.load(in_file)
+    # Load the mask image
     mask_nii = nb.load(mask_file)
+    # Set all non-mask voxels in the input file to zero.
     data = in_nii.get_data()
     data[mask_nii.get_data() == 0] = 0
```
oesteban:
```
data[mask_nii.get_data() <= 0] = 0
```

niworkflows/interfaces/mni.py
```diff
@@ -231,18 +354,115 @@ def _validate_results(self):
...
+    ----------
+    in_file : str
+        Path to an existing image (usually a mask).
+        If global_mask = True, this is used as a size/dimension reference.
+    out_path : str
+        Path/filename for the new cost function mask.
+    lesion_mask : str, optional
+        Path to an existing binary lesion mask.
+    global_mask : bool
+        Create a whole-image mask (True) or limit to reference mask (False)
```
oesteban: what does "whole-image mask" mean? It is a bit unclear what the result will be if `global_mask=True`.

niworkflows/interfaces/mni.py
```diff
@@ -231,18 +354,115 @@ def _validate_results(self):
...
+    -------
+    str
+        Absolute path of the new cost function mask.
+
+    Notes
+    -----
+    in_file and lesion_mask must be in the same
+    image space and have the same dimensions
+    """
+    import os
```
oesteban: should these imports be added on top and removed from here?

niworkflows/interfaces/mni.py
```diff
@@ -231,18 +354,115 @@ def _validate_results(self):
...
+    if not global_mask and not lesion_mask:
+        NIWORKFLOWS_LOG.warning(
+            'No lesion mask was provided and global_mask not requested, '
+            'therefore the original mask will not be modified.')
+
+    # Load the input image
+    in_img = nb.load(in_file)
+
+    # If we want a global mask, create one based on the input image.
+    if global_mask:
```
oesteban:
```Python
data = np.ones(in_img.shape, dtype=np.uint8)  # Set a default whole-image mask

if not global_mask:  # Rewrite mask if not global
    data = np.asanyarray(in_img.get_data() > 0, dtype=np.uint8)
```

niworkflows/interfaces/mni.py
```diff
@@ -231,18 +354,115 @@ def _validate_results(self):
...
+        # Create a mask of ones with the shape of the input image.
+        data = np.ones(in_img.shape, dtype=np.uint8)
+    else:
+        data = in_img.get_data()
+        if set(np.unique(data)) - {0, 1}:
```
oesteban: Doing what @effigies suggested right above this is not necessary anymore.

niworkflows/interfaces/mni.py
oesteban: Why `asany`?

niworkflows/interfaces/mni.py
```diff
@@ -231,18 +354,115 @@ def _validate_results(self):
...
+        # Cost function mask will be created from subtraction
+    # Otherwise, CFM will be created from global mask
+
+    cfm_img = nb.Nifti1Image(data, in_img.affine, in_img.header)
+
+    # Save the cost function mask.
+    cfm_img.set_data_dtype(np.uint8)
+    cfm_img.to_filename(out_path)
+
+    return out_path
```
oesteban: At this point, if `global_mask=True` and `lesion_mask = None` then this mask will be all-ones. It'd be easier to debug if we returned `None` and didn't set a mask into ANTs. WDYT?

niworkflows/interfaces/mni.py
```diff
@@ -231,18 +354,115 @@ def _validate_results(self):
 
 
 def mask(in_file, mask_file, new_name):
```
oesteban: Better more explicit about the action:

```Python
def apply_mask(....
```

niworkflows/interfaces/mni.py
```diff
@@ -161,42 +197,129 @@ def _get_ants_args(self):
                 'write_composite_transform': True,
                 'initial_moving_transform': self.inputs.initial_moving_transform}
 
+        """
+        Moving image handling
```
oesteban: I'd love to see here one of those @effigies' truth tables. I feel we can make all these branches simpler.

| `inputs.moving_mask` | `inputs.explicit_masking` | `inputs.lesion_mask` | action                                                                                                                  |
|----------------------|---------------------------|----------------------|-------------------------------------------------------------------------------------------------------------------------|
| True                 | True                      | True                 | Update `moving_image` after applying `apply_mask` Set `moving_image_mask` applying `create_cfm` with `global_mask=True` |
| True                 | True                      | False                | Update `moving_image` after applying `apply_mask`                                                                       |
| True                 | False                     | True                 | Set `moving_image_mask` applying `create_cfm` with `global_mask=False`                                                  |
| True                 | False                     | False                | `args['moving_image_mask'] =self.inputs.moving_mask`                                                                    |
| False                | *                         | True                 | Set `moving_image_mask` applying `create_cfm` with `global_mask=True`                                                   |
| False                | *                         | False                | Nothing                                                                                                                 |

niworkflows/interfaces/mni.py
```diff
@@ -161,42 +197,129 @@ def _get_ants_args(self):
...
+        elif isdefined(self.inputs.lesion_mask):
+            # Create a cost function mask with the form: [global mask - lesion mask]
+            # Use this as the moving mask.
+            args['moving_image_mask'] = create_cfm(
+                self.inputs.moving_image,
+                lesion_mask=self.inputs.lesion_mask,
+                global_mask=True)
+
+        """
+        Reference image handling
```
oesteban: Truth table again, reuse function if underlying logic is the same.

niworkflows/interfaces/mni.py
```diff
@@ -161,42 +197,129 @@ def _get_ants_args(self):
...
+                # Mask the template image with the template mask.
+                # Do not use a fixed mask during registration.
+                _template_fmt = op.join(mni_template, '%dmm_%s.nii.gz')
+                args['fixed_image'] = mask(
+                        _template_fmt % (resolution, self.inputs.reference),
+                        _template_fmt % (resolution, 'brainmask.nii.gz'),
+                        "fixed_masked.nii.gz")
+
+                # If a lesion mask is provided...
+                if isdefined(self.inputs.lesion_mask):
```
oesteban: After we get the previous comments more clear, I'd dive into this.

niworkflows/interfaces/mni.py
effigies: `asanyarray` passes through arrays that don't need to change without copying. `np.array()` of a `np.memmap` will copy, but `np.asanyarray` will not unless it has to (e.g. the dtype doesn't match).

But anyway, we settled with emitting an error rather than silently interpreting a non-mask as a mask.

oesteban:
Sounds good

niworkflows/interfaces/mni.py
effigies: As a pre-existing funtion, I'm going to say this is out-of-scope for this PR. If we're going to bother renaming, we should probably just go ahead and use nilearn to handle masking.

niworkflows/interfaces/mni.py
effigies: I guess I don't see a value in allocating an array we'll have to throw out. I guess it's only 3D, so that's NBD, but how about we just use a ternary operator?

```Python
data = np.ones(in_img.shape, dtype=np.uint8) if global_mask else in_img.get_data()
```

niworkflows/interfaces/mni.py
effigies: See above comment on scope.

niworkflows/interfaces/mni.py
oesteban: We can just add the table into the comment for now and try to simplify in a future.

niworkflows/interfaces/mni.py
effigies: So just return [here](https://github.com/poldracklab/niworkflows/pull/233/files#diff-34ffe363a301e0c280a14f9fb630ac2fR433)? Can do that.

niworkflows/interfaces/mni.py
danlurie: In this case I was just following the style from the existing function ("mask"). I've no particular opinion of where the imports should live.

niworkflows/interfaces/mni.py
effigies: Masks have to come in pairs. When this is called in that way, it's because we need a mask with null-effect to partner with the intended mask.

niworkflows/interfaces/mni.py
danlurie: To be clear, are you talking here about adding a comment for a truth table? I'd be inclined to say that for right now this is a "nice to have" and that it might be worth doing a separate PR for the comments (so we can get the code merged today).

effigies:
Last two commits are no-effect, so if our tests work (running with and without a lesion mask), I think this is good to go.

@oesteban Let me know if any of the comments we rejected/postponed you think are important to deal with first.

oesteban:
Let's go as is and re-iterate in the future. This has been sitting here for too long, time to close up!

effigies:
@danlurie Are we good to go?

danlurie:
Great success! No errors in either run.
ezgif com-gif-maker

oesteban:
Outstanding job guys

PR #234, opened 2018-05-04
Bump dev version to 0.3.12-dev

effigies:
@oesteban Made this a PR just in case I've forgotten how we do this...


oesteban:
Actually, you just reminded me that I've been releasing without updating versions.

PR #236, opened 2018-05-06
RF: Normalization image/mask preparation

effigies:
This factors template normalization into three interfaces:

* Template default/override selection
* Explicit masking and lesion masking
* Normalization

I will propose a workflow in fmriprep to use these together.


oesteban:
This has been replaced by #238, I'll close it. Feel free to reopen is that was a mistake.

PR #255, opened 2018-10-12
[ENH] Add "fMRIPrep" template, with new boldref template

oesteban:


effigies:
This might be a good train for @ihashi3 to hop on. We would need to add the pediatric atlas here to use it in fMRIPrep. (Doesn't have to be this PR, but just to draw attention to how this is done.)

oesteban:
This should be merged after #256

oesteban:
Under test in https://github.com/poldracklab/fmriprep/pull/1321

PR #257, opened 2018-10-18
[RF] Updates for templateflow

oesteban:
To work with https://github.com/poldracklab/fmriprep/pull/1334


oesteban:
Any crazy stuff that I don't see?

niworkflows/data/__init__.py
```diff
@@ -8,20 +8,8 @@
...
-    get_ants_oasis_template,
-    get_ants_oasis_template_ras,
-    get_mni_epi,
-    get_mni152_nlin_sym_las,
-    get_mni152_nlin_sym_ras,
-    get_mni_icbm152_linear,
-    get_mni_icbm152_nlin_asym_09c,
-    get_oasis_dkt31_mni152,
-    get_hcp32k_files,
-    get_conte69_mesh,
```
effigies: Should we go through a deprecation cycle for these? Or is niworkflows more of a makes-no-promises sort of library?

niworkflows/interfaces/masks.py
```diff
@@ -259,20 +262,50 @@ def _generate_report(self):
...
+
+        nsegs = len(seg_files)
+        if nsegs == 1:
+            if not levels:  # Segmentation
+                levels = np.unique(np.round(
+                    nb.load(seg_files[0]).get_data()).astype(int))
+                levels = (levels[levels > 0] - 0.5).tolist()
+            nsegs = len(levels)
+        else:
+            if not levels:
```
effigies: `elif not levels:`?

niworkflows/interfaces/masks.py
```diff
@@ -259,20 +262,50 @@ def _generate_report(self):
...
+            colors = colors + color_palette("husl", missing)
+
+        sortedc = []
+        start = 0
+        for l in levels:
+            end = start + len(l)
+            sortedc.append(colors[start:end])
+            start = end
 
+        print(levels, sortedc)
```
effigies: Suspect this shouldn't be here.

niworkflows/viz/utils.py
```diff
@@ -283,28 +283,32 @@ def plot_segs(image_nii, seg_niis, out_file, bbox_nii=None, masked=False,
...
 def _plot_anat_with_contours(image, segs=None, compress='auto',
                              **plot_params):
-    assert segs is not None
-
+    nsegs = len(segs or [])
     plot_params = plot_params or {}
-    colors = plot_params.pop('colors', []) or []  # colors should not be None
-    nsegs = len(segs)
+    colors = plot_params.pop('colors', None) or []
+    levels = plot_params.pop('levels', None) or []
```
effigies: Suspect I'm missing why you changed it, but isn't this equivalent to:

```Python
colors = plot_params.pop('colors', [])
levels = plot_params.pop('levels', [])
```

niworkflows/data/__init__.py
oesteban: Given that this was a pretty unscalable approach and that fmriprep is the only client of niworkflows... I'd go with makes-no-promises. That will also push people to update fmriprep if they are not using containers, which is positive in the long term.

We should consider always setting a maximum version for niworkflows in fmriprep, though.

niworkflows/viz/utils.py
oesteban: Unfortunately not:
```Python
>>> plot_params = {'colors': None, 'levels': None}
>>> colors = plot_params.pop('colors', [])
... levels = plot_params.pop('levels', [])
>>> colors is None
True
>>> levels is None
True
```

Many times they are set and `None`, so I had to change to that ugly `or`. Maybe there's a better way of doing this I'm missing (e.g. cleaning the dictionary of `None`s before, however I'm not sure other kw may use None as a valid value or maybe replacing and then pop... I don't have a preference).

effigies: Got it. Maybe just a quick comment (e.g. "None is a valid parameter") so the next person to look at this won't think the same thing.

effigies:
Sounds good.

effigies:
Good to go? Looks like poldracklab/fmriprep#1334 is passing.

effigies:
(Or do we want to wait for the next release, to avoid too much churn?)

oesteban:
I'm fixing the colors of `ROIsPlot`. If I'm not done by the end of the day, let's push it to the next release.

oesteban:
Alright, this is finished. I've tested locally and the plots look the same as before the PR 👍.

Currently testing fmriprep here: https://circleci.com/workflow-run/d3fa2769-7013-4fce-83f0-23e9a32ae158

Right after merging this, I'll release 0.5.0 (I think the big change in templates deserves that minor bump)

PR #259, opened 2018-11-06
[FIX] Remove spurious nipype empty file

oesteban:
Removes this file, which seems added by mistake in https://github.com/poldracklab/niworkflows/commit/181e0680b24eb339c695c5b0fce035ea61a07950#diff-b67f391f860d7d8302fc023720679a77. Please let me know if I'm wrong and it actually serves a purpose @effigies @chrisfilo.


effigies:
Nope. Almost certainly a leftover symlink that got re-added after we went back to depending on nipype release versions.

chrisfilo:
I believe that's my bad. Sorry about that.

PR #263, opened 2018-12-07
[ENH] Upstream work from fMRIPrep (prep. sMRIPrep)

oesteban:
## Summary

This PR contains a batch of utilities in fmriprep that need to be upstreamed to niworkflows in order for smriprep to access them (without depending on fmriprep).

No improvements have been done (except for sorting out imports, cleaning up, stop using os.path at places, etc.)


niworkflows/interfaces/bids.py
```diff
@@ -0,0 +1,374 @@
...
+    SimpleInterface
+)
+from ..utils.bids import BIDS_NAME, get_metadata_for_nifti
+from ..utils.misc import splitext as _splitext, _copy_any
+
+LOGGER = logging.getLogger('nipype.interface')
+
+
+class FileNotFoundError(IOError):
+    pass
```
effigies: Is there any reason not to use the built-in `FileNotFoundError`?

niworkflows/interfaces/utils.py
```diff
@@ -414,3 +417,400 @@ def _run_interface(self, runtime):
...
+
+    def _run_interface(self, runtime):
+        out_mat = os.path.join(runtime.cwd, 'concat.mat')
+        in_list = [self.inputs.get()[name] for name in self._get_names(self._num_affines)]
+
+        out_xfm = _concat_xfms(in_list, invert=self.inputs.invert)
+        np.savetxt(out_mat, out_xfm, fmt=str('%.12g'))
+
+        self._results['out_mat'] = out_mat
+        return runtime
```
effigies: I don't think we use this interface anymore.

Issue #265, opened 2018-12-10
Remove ``SignalExtraction`` ?


oesteban:
It seems like we do not use the SignalExtraction from fmriprep anymore (recently moved into niworkflows codebase). Can we just delete this one?

effigies:
Actually, it looks like we never used the implementation in poldracklab/fmriprep#1170, even though it was faster and took less memory. Maybe we should switch to using it?

oesteban:
Hell yeah then

effigies:
Do you have an fmriprep branch open to test moving all of these components to niworkflows? We can just update there.

oesteban:
:+1:

PR #266, opened 2018-12-10
[TST] Add tests missed out in #263

oesteban:
Ref #265. Upstream missed out tests on ``SignalExtraction``.


oesteban:
Checked locally, they should pass (they were also passing with fmriprep within the container, so I'll take the risk)

PR #268, opened 2018-12-13
[FIX] Summary fMRIPlot chokes when confounds are all-nan

oesteban:
Allow the summary plot to go on if one confound is fully nans


oesteban:
Tested locally, I'm going ahead and merge to release quickly

PR #269, opened 2018-12-14
[HOTFIX] ``LiterateWorkflow`` returning empty desc

oesteban:
When migrated from fmriprep, I replaced networkx.topological_sort
with nipype.pipeline.engine.utiles.topological_sort, which does not
have exactly the same interface (basically it wraps the iterable
within a list).


oesteban:
Tested locally

PR #271, opened 2018-12-20
Conform names to updated TemplateFlow, add options conducive to small animal neuroimaging

rciric:
* Names conform to the BIDS standard as implemented in the current version of TemplateFlow.
* Use of Laplacian alignment as a criterion during the final (CC-based) registration stage is made optional.
* The `bspline_fitting_distance` is measured in mm. The default setting of 200 can become problematic when images have a smaller FoV, as is often the case in small animal neuroimaging. Thus, setting this value is also made available as an option. (I'm wondering if we shouldn't set this to a reasonable default based on the image affine rather than using 200.)


niworkflows/anat/ants.py
```diff
@@ -252,13 +252,20 @@ def init_brain_extraction_wf(name='brain_extraction_wf',
...
     # Set up spatial normalization
-    norm = pe.Node(Registration(
-        from_file=pkgr_fn(
-            'niworkflows.data',
-            'antsBrainExtraction_%s.json' % normalization_quality)),
-        name='norm',
-        n_procs=omp_nthreads,
-        mem_gb=mem_gb)
+    if use_laplacian:
+        norm = pe.Node(Registration(from_file = pkgr_fn(
```
oesteban: just as a stylistic suggestion, the if ... else can be more readable stripping the settings file name:

```Python
    settings_file = 'antsBrainExtraction_%s.json' if use_laplacian \
        else 'antsBrainExtractionNoLaplacian_%s.json'
    norm = pe.Node(Registration(from_file = pkgr_fn(
            'niworkflows.data', settings_file % normalization_quality)),
        mem_gb=mem_gb,
        n_procs=omp_nthreads,
        name='norm')
```

oesteban:
Good job!

PR #280, opened 2019-01-17
[ENH] Datalad+templateflow integration

oesteban:
- [x] Revise workflows to use `templateflow.api.get`.
- [x] Revise tests to use templates from templateflow.
- [x] Docker image: install datalad and templateflow, cache some templates + incidental improvements.
- [x] (incidental) Do not bake FS' license into the container anymore.

### DIFF
--- a/niworkflows/__about__.py
+++ b/niworkflows/__about__.py
@@ -64,6 +64,7 @@ REQUIRES = [
     'scipy',
     'seaborn',
     'svgutils',
+    'templateflow>=0.0.3',
 ]
 
 SETUP_REQUIRES = []
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -91,6 +91,11 @@ jobs:
               docker images
             fi
       - run:
+          name: Store FreeSurfer license file
+          command: |
+            mkdir -p /tmp/fslicense
+            printf "$FS_LICENSE_CONTENT" | base64 -d  >> /tmp/fslicense/license.txt
+      - run:
           name: Run unit tests
           no_output_timeout: 2h
           command: |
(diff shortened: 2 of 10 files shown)


niworkflows/__about__.py
```diff
@@ -64,6 +64,7 @@
     'scipy',
     'seaborn',
     'svgutils',
+    'templateflow>=0.0.3',
```
effigies: Since we import from `datalad`, I think we should make it an explicit dependency, even if it's covered by `templateflow`.

```suggestion
    'templateflow>=0.0.3',
    'datalad',
```

oesteban: oooh true. I'll change the import if you don't mind.

oesteban:
> Have we ever tested the skullstripping workflow?

@rciric has used it quite a bit, so I think I can safely say "yes" to this question.

> we should definitely pin fmriprep and mriqc to this branch and make sure they build and run before merging

Sure, I was also thinking of bumping versions to 0.6.0 or even 1.0.0. WDYT?

niworkflows/__about__.py
oesteban: Done

effigies:
I'm okay with bumping versions. But nothing uses the semantics of our versioning system (and we don't backport fixes from master to other minor version series) so it's not critical. It's mostly just let's try not to break things.

oesteban:
The reason why I propose 0.6.0 is this:

https://github.com/poldracklab/smriprep/blob/master/smriprep/__about__.py#L54

https://github.com/poldracklab/fmriprep/blob/master/fmriprep/__about__.py#L93 (<- although this one should be <0.6.0 as well)

https://github.com/poldracklab/mriqc/blob/master/mriqc/__about__.py#L60

In principle, tagging 0.6.0 will not break any of them.

oesteban:
What do you think about releasing 0.6.0 but keep accepting PR to the 0.5.x branch?

effigies:
Sure. That might end up being easier than having to co-release.

Let's start a `maint/0.5.x` branch that we'll PR into and release off of. Is it safe to use current `master`, or should we go back some commits first?

oesteban:
Totally safe to branch at this point of master, I've just opened `maint/0.5` :+1:

PR #282, opened 2019-01-18
MAINT: Drop more FreeSurfer components

effigies:
I think for testing niworkflows we only need `bin/` and possibly some `lib/`. Letting Circle decide if I'm right.

Meant to post this against Oscar's branch, but that got merged.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -36,7 +36,8 @@ RUN curl -sSL https://surfer.nmr.mgh.harvard.edu/pub/dist/freesurfer/6.0.1/frees
     --exclude='freesurfer/mni' \
     --exclude='freesurfer/subjects' \
     --exclude='freesurfer/tktools' \
-    --exclude='freesurfer/trctrain'
+    --exclude='freesurfer/trctrain' && \
+    mkdir -p /opt/freesurfer/subjects
 
 ENV FSL_DIR="/usr/share/fsl/5.0" \
     OS="Linux" \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -25,23 +25,17 @@ RUN apt-get update && \
...
-    --exclude='freesurfer/subjects/fsaverage4' \
-    --exclude='freesurfer/subjects/cvs_avg35' \
-    --exclude='freesurfer/subjects/cvs_avg35_inMNI152' \
-    --exclude='freesurfer/subjects/bert' \
-    --exclude='freesurfer/subjects/lh.EC_average' \
-    --exclude='freesurfer/subjects/rh.EC_average' \
-    --exclude='freesurfer/subjects/sample-*.mgz' \
-    --exclude='freesurfer/subjects/V1_average' \
+    --exclude='freesurfer/mni' \
+    --exclude='freesurfer/subjects' \
```
oesteban: you can't remove `freesurfer/subjects` in full, `fsaverage{5,6,}` have to be there

effigies: For niworkflows?

oesteban: In principle not, maybe you can get away with `mkdir -p /opt/freesurfer/subjects`.

effigies: Oh, I see. The failing test.

effigies:
Ah thanks for updating this.

effigies:
Oh, looks like we need `fsaverage`, still.

effigies:
Eh. This is more trouble than it's worth for a minor potential pickup in speed.

Issue #273, opened 2018-12-20
Enable CodeCov


effigies:
With so much from fMRIPrep coming in here and more projects intending to depend more on niworkflows, we should be verifying basic integrity with smoketests.

effigies:
Working: https://codecov.io/gh/poldracklab/niworkflows/branch/master

Fixed in #288, #290.

PR #285, opened 2019-01-19
[HOTFIX] Fix ``UnboundLocalError`` in utils.bids

oesteban:
Addresses the issue posted in https://neurostars.org/t/fmriprep-from-singularity-unboundlocalerror/3299?u=oesteban

### DIFF
--- a/niworkflows/utils/bids.py
+++ b/niworkflows/utils/bids.py
@@ -167,21 +167,17 @@ def collect_data(dataset, participant_label, task=None, echo=None):
     subj_data = {modality: [x.filename for x in layout.get(**query)]
                  for modality, query in queries.items()}
 
-    def _grp_echos(x):
-        if '_echo-' not in x:
-            return x
-        echo = re.search("_echo-\\d*", x).group(0)
-        return x.replace(echo, "_echo-?")
-
-    if subj_data["bold"]:
-        bold_sess = subj_data["bold"]
-
-        if any(['_echo-' in bold for bold in bold_sess]):
-            ses_uids = [list(bold) for _, bold in groupby(bold_sess, key=_grp_echos)]
-            ses_uids = [x[0] if len(x) == 1 else x for x in ses_uids]
-        else:
-            ses_uids = bold_sess
-
+    if subj_data['bold'] and any(['_echo-' in bold for bold in bold_sess]):
+        # Special case: multi-echo BOLD, grouping echos
+        def _grp_echos(x):
+            if '_echo-' not in x:
+                return x
+            echo = re.search("_echo-\\d*", x).group(0)
+            return x.replace(echo, "_echo-?")
+        
+        bold_sess = subj_data['bold']
+        ses_uids = [list(bold) for _, bold in groupby(bold_sess, key=_grp_echos)]
+        ses_uids = [x[0] if len(x) == 1 else x for x in ses_uids]
         subj_data.update({"bold": ses_uids})
 
     return subj_data, layout
--- a/niworkflows/utils/bids.py
+++ b/niworkflows/utils/bids.py
@@ -283,6 +283,30 @@ def group_multiecho(bold_sess):
       'sub-01_task-beh_echo-3_run-02_bold.nii.gz'],
       'sub-01_task-beh_run-03_bold.nii.gz']
 
+    Some tests from https://neurostars.org/t/fmriprep-from\
+-singularity-unboundlocalerror/3299/7
+
+    >>> bold_sess = ['sub-01_task-AudLoc_echo-1_bold.nii',
+    ...              'sub-01_task-AudLoc_echo-2_bold.nii',
+    ...              'sub-01_task-FJT_echo-1_bold.nii',
+    ...              'sub-01_task-FJT_echo-2_bold.nii',
+    ...              'sub-01_task-LDT_echo-1_bold.nii',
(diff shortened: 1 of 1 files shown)


oesteban:
Can you make sure this is correct, @emdupre?

emdupre:
One concern here, and likely how the problem arose: dual echo data is considered a special case in multi-echo, and you would not make an optimal combination for it or calculate a T2* map, etc. Requesting one change for that:

niworkflows/utils/bids.py
```diff
@@ -167,22 +167,18 @@ def collect_data(dataset, participant_label, task=None, echo=None):
...
+        # Special case: multi-echo BOLD, grouping echos
+        def _grp_echos(x):
+            if '_echo-' not in x:
+                return x
+            echo = re.search("_echo-\\d*", x).group(0)
+            return x.replace(echo, "_echo-?")
+
+        bold_sess = subj_data['bold']
+        ses_uids = [list(bold) for _, bold in groupby(bold_sess, key=_grp_echos)]
+        ses_uids = [x[0] if len(x) == 1 else x for x in ses_uids]
```
emdupre:
```suggestion
        ses_uids = [x[0] if len(x) <= 2 else x for x in ses_uids]
```

effigies: Is there any possibility of `x == []`?

emdupre: I don't believe so... If lacks an `echo` key-value pair you shouldn't hit this condition. Can you think of a case ?

effigies: Not obviously, but my brain isn't very engaged right now. Mostly just recognizing the pattern and asking you to check. :-)

oesteban: why would we want to pick only the first element if there are two echos? Maybe I lost something when transferring from fmriprep.

emdupre:
Actually, I modified some data locally to be dual echo and I'm still not encountering the reported problem (neither with the original code nor your branch).

I looked at the error again and I'm still not sure how you'd hit that condition unless you didn't have BOLD data, but from his call it looks like it was BIDS validated with recognized BOLD data... Curious to understand what happened, here. Either way, I think your re-write is safer. 

You are right, though, that they may want both files returned (depending on what they're doing with the dual echo). I'll make one more suggested change, sorry !

niworkflows/utils/bids.py
```diff
@@ -167,22 +167,18 @@ def collect_data(dataset, participant_label, task=None, echo=None):
...
+        # Special case: multi-echo BOLD, grouping echos
+        def _grp_echos(x):
+            if '_echo-' not in x:
+                return x
+            echo = re.search("_echo-\\d*", x).group(0)
+            return x.replace(echo, "_echo-?")
+
+        bold_sess = subj_data['bold']
+        ses_uids = [list(bold) for _, bold in groupby(bold_sess, key=_grp_echos)]
+        ses_uids = [x[0] if len(x) <= 2 else x for x in ses_uids]
```
emdupre:
```suggestion
                ses_uids = [x for ses in ses_uids for x in ses if len(ses) <= 2] + \
                   [ses for ses in ses_uids if len(ses) > 2 ]
```

emdupre:
Ok, I've updated this to just return dual echos as if they were single-echo sequences. I believe my original reasoning was that even as a dual-echo sequence, there are certain facets which are better considered in a multi-echo context (such as coregistration), but that brings up some larger issues that I don't think are directly related, here.

Of note: I ran the original code with the file tree reported in the error and can't reproduce the issue locally. Is anyone else able to ?

niworkflows/utils/bids.py
```diff
@@ -245,3 +231,71 @@ def get_metadata_for_nifti(in_file):
...
+      'sub-01_task-rest_echo-3_run-02_bold.nii.gz'],
+      'sub-01_task-rest_run-03_bold.nii.gz',
+      'sub-01_task-beh_echo-1_run-01_bold.nii.gz',
+      'sub-01_task-beh_echo-2_run-01_bold.nii.gz',
+     ['sub-01_task-beh_echo-1_run-02_bold.nii.gz',
+      'sub-01_task-beh_echo-2_run-02_bold.nii.gz',
+      'sub-01_task-beh_echo-3_run-02_bold.nii.gz'],
+      'sub-01_task-beh_run-03_bold.nii.gz']
+
+    """
```
effigies: I don't think we run doctests.

oesteban: I think I added them when I brought the code from fmriprep:

https://github.com/poldracklab/niworkflows/blob/master/.circleci/config.yml#L69-L126

effigies: Oh, you're right. It's in:

https://github.com/poldracklab/niworkflows/blob/f69a835c33f2c465cda57bd8533e2d4030657ba3/tox.ini#L12

effigies: Would be easier to be check with coverage turned on. :-)

oesteban: Definitely - #273

oesteban:
I'm going to go ahead. I'll happily accept any PR improving this implementation if necessary.

emdupre:
So sorry for the radio silence ! I'll look into this tomorrow, but from the quickest review it seems great :+1: I'll open a PR if I have any concerns :)

PR #287, opened 2019-01-21
[ENH] Confound enhancement

rciric:
This PR enables the following enhancements to the confound modelling workflow:

* The `SpikeRegressors` interface adds spike regressors to a confound model on the basis of specified criteria.
  * By default, these criteria are a framewise displacement of 0.2 mm and DVARS of 20, following the censoring recommendations in the [2014 Power](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC3849338/) paper.
  * Spike regressors can be used to exclude frames from any model fitting step.
  * Currently, spikes use column names `motion_outlier<index>` (to parallel `non_steady_state_outlier<index>`). I'm not sure what the BIDS standard name for a motion spike column is, and it doesn't appear to be specified in the outliers section in [RC1](https://docs.google.com/document/d/17ebopupQxuRwp7U7TFvS6BH03ALJOgGHufxK8ToAvyI/edit#heading=h.dbfk7w8jo2wp) or [BEP012](https://docs.google.com/document/d/16CvBwVMAs0IMhdoKmlmcm3W8254dQmNARo-7HhE-lJU).
* The `ExpandModel` interface can be used to add temporal derivative and exponential terms.
  * By default, a 36-parameter expansion is generated (quadratic terms/first derivatives of realignment parameters, mean WM and CSF time series, and mean global time series), following (for instance) [Satterthwaite 2013](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC3811142/).
  * The same interface can be also used to extract columns from a TSV file for use in future confound regression steps.


oesteban:
Hey, you want to copy from `poldracklab/niworkflows` the `FS_LICENSE_CONTENT` environment variable of Circle (or include yours after base64 encoding).

oesteban:
Do you think we could add some kind of tests for these interfaces?

niworkflows/interfaces/confounds.py
```diff
@@ -0,0 +1,548 @@
...
+        'a single variable (e.g., (x)^2, not x^2).')
+    output_file = traits.Str(desc='Output path')
+
+
+class ExpandModelOutputInterface(TraitedSpec):
+    confounds_file = File(exists=True, desc='Output confounds file')
+
+
+class ExpandModel(SimpleInterface):
+    """Expand a confound model according to a specified formula.
```
oesteban: Could you add here some standard formulas that people may find handy?

niworkflows/interfaces/confounds.py
```diff
@@ -0,0 +1,548 @@
...
+        self._results['confounds_file'] = out_file
+        return runtime
+
+
+class SpikeRegressorsInputInterface(BaseInterfaceInputSpec):
+    confounds_file = File(
+        exists=True, mandatory=True,
+        desc='TSV containing criterion time series (e.g., framewise '
+        'displacement, DVARS) to be used for creating spike regressors.')
+    criteria = traits.Dict(
```
oesteban: what about splitting this dict into two independent, Float inputs: `fd_thresh` and `dvars_thresh`?

niworkflows/interfaces/confounds.py
```diff
@@ -0,0 +1,548 @@
...
+        None, traits.Int, usedefault=True,
+        desc='Minimum number of contiguous volumes required to avoid '
+        'flagging as a spike')
+    concatenate = traits.Bool(
+        True, usedefault=True,
+        desc='Indicates whether to concatenate spikes to existing confounds '
+        'or return spikes only')
+    output_format = traits.Enum('spikes', 'mask', usedefault=True,
+                                desc='Format of output (spikes or mask)')
+    output_file = traits.Str(desc='Output path')
```
oesteban:
```suggestion
    output_file = File(desc='Output path')
```

niworkflows/interfaces/confounds.py
```diff
@@ -0,0 +1,548 @@
...
+                     output='spikes'):
+    """
+    Add spike regressors to a confound/nuisance matrix.
+
+    Parameters
+    ----------
+    data: pandas DataFrame object
+        A tabulation of observations from which spike regressors should be
+        estimated.
+    criteria: dict{str: ('>' or '<', float)}
```
oesteban: here it makes sense to keep the dictionary

PR #288, opened 2019-01-24
CI: Add coverage

effigies:
Codecov never works right until merged into master, but we should make sure that we can actually run the whole Circle job.


oesteban:
Should we merge into `maint/0.5.x` better?

effigies:
Might as well get going on master. Can cherry pick onto maint once we know what works.

effigies:
Test failure seems related to nedbat/coveragepy#750. Trying out an older version of `pytest-cov`.

PR #292, opened 2019-01-27
[ENH] AROMA plots - add warning for edge cases (none/all are noise)

jdkent:
per @oesteban's [comment](https://github.com/poldracklab/fmriprep/pull/1467#issuecomment-457018934), this pull request attempts to add a warning to the figure if all components are listed as either signal or noise.
references poldracklab/fmriprep#1467

However, I must be needing to improve my matplotlib since even with this addition the warning does not appear to be displayed in the fmriprep reports.

### DIFF
--- a/niworkflows/tests/test_viz.py
+++ b/niworkflows/tests/test_viz.py
@@ -21,6 +21,7 @@ def test_carpetplot():
         legend=True
     )
 
+
 def test_plot_melodic_components():
     """Test plotting melodic components"""
     import numpy as np
@@ -55,7 +56,7 @@ def test_plot_melodic_components():
     # noise_components
     noise_comps = np.array([1, 2])
     noise_comps_file = os.path.join(os.getcwd(), 'noise_ics.csv')
-    np.savetxt(noise_comps_file, noise_comps, 
+    np.savetxt(noise_comps_file, noise_comps,
                fmt='%i', delimiter=',')
     # in_file
     voxel_ts = np.random.rand(2, 2, 2, 10)
--- a/niworkflows/tests/test_viz.py
+++ b/niworkflows/tests/test_viz.py
@@ -20,3 +20,57 @@ def test_carpetplot():
         output_file=out_file,
         legend=True
     )
+
+def test_plot_melodic_components():
+    """Test plotting melodic components"""
+    import numpy as np
+    # save the artifacts
+    save_artifacts = os.getenv('SAVE_CIRCLE_ARTIFACTS', False)
+    if save_artifacts:
+        all_noise = os.path.join(save_artifacts, 'melodic_all_noise.svg')
+        no_noise = os.path.join(save_artifacts, 'melodic_no_noise.svg')
+    else:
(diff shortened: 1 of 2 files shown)


jdkent:
I appear to need a more recent freesurfer license for the tests to complete:
```
GNU libc version: 2.23
ERROR: Systems running GNU glibc version greater than 2.15
require a newly formatted license file (it's free). Please
download a new one from the following page:
http://surfer.nmr.mgh.harvard.edu/registration.html
```

effigies:
I'm rerunning with SSH. I suspect your license file is empty...

effigies:
Apparently I can't SSH into the tests for the PR you made. I opened #294. If that passes, could you go ahead and cherry-pick dd243283701604e377ede1797ccbb292d659f731 to make sure that your license file is the same as ours?

jdkent:
I'll try to SSH in as well.

oesteban:
You'll need to set the FS_LICENSE_CONTENT. Allow me 20min and will tell you
more

effigies:
It seems to be running under poldracklab/niworkflows, not jdkent/niworkflows. The problem is that we don't use our environment variables on PRs. This is a security measure, as anybody could create a PR where an environment variable containing e.g., PyPI passwords, was just printed to the screen.

jdkent:
Thanks @effigies! Then the solution would be to start building my forked niworkflows on circleci and make a `$FS_LICENSE_CONTENT` environment variable ([in circleci](https://circleci.com/docs/2.0/env-vars/#setting-an-environment-variable-in-a-project)) from a `cat license.txt | base64` command?

effigies:
That is a solution. I don't really like the idea that we're forcing all contributors to configure a CircleCI setup.

oesteban:
Sorry for the wait, I just sent you an email with handy links and hints.

oesteban:
I can see the warning here: https://469-92465059-gh.circle-artifacts.com/0/tmp/ds005/derivatives_partial/fmriprep/sub-01.html#task-mixedgamblestask_run-02

We should add some tests to the AROMA reportlets with pre-baked data. Do you need any pointers?

jdkent:
I think I understand, I have a couple of clarifying questions:
1. Is the AROMA reportlets referring to the  [ICA_AROMARPT](https://github.com/poldracklab/niworkflows/blob/963d080ce12d62cc12898b411580a68d4dfbdb36/niworkflows/interfaces/segmentation.py#L149) interface?

2. What cases do I need to test for the interface?
  - The input provided would have to run through ICA_AROMA and then through the report mechanism. I was hesitant towards this since from past experience ICA_AROMA is not necessarily deterministic in its labeling of signal/noise components. Constraining the number of components was my solution for trying to make all the components noise (but even that wasn't a sure thing as I had to switch from testing run-1 to run-2) and I'm not sure how to make all the components signal to test the opposite case.

3. Would the inputs to ICA-AROMA be stored in dropbox or some other storage mechanism?

Thanks for helping push this along!

PR #294, opened 2019-01-29
CI: MD5 license to verify contents

effigies:
The goal of this is to help debug situations like #292, which I suspect is a result of the license file not being properly made for PRs.


effigies:
In retrospect this was dumb. I have my own Circle setup. Figured out the issue anyway in https://github.com/poldracklab/niworkflows/pull/292#issuecomment-458634818.

PR #297, opened 2019-01-31
[ENH,WIP] Establish a unique, global ``BIDSLayout``

oesteban:


effigies:
So I'm super wary of global variables like this, and this could cause problems if anybody wants to use these functions/interfaces with multiple datasets. I recognize that we don't currently handle multiple datasets, but if these are at least in some sense meant to be candidates for upstreaming to nipype, I think this is asking for breakage.

Would it be easier to do all of this if we just made `collect_participants` take a `layout` parameter, and then fmriprep/mriqc could instantiate a layout once and pass it in?

For the things like parsing file names, it might be worth pinging Tal over in bids-standard/pybids#369 about APIs that shouldn't require crawling a directory to have access to.

oesteban:
I'm not a fan of super-global variables, but I think here it makes most sense. My original thought about this is that (like we do with the context in multiprocessing) fmriprep initializes the layout in the beginning. This way, it can be queried at any time throughout the workflow.

Similarly to the context of multiprocessing, the logic is made to allow setting it just once unless you are very aware that you want to initialize another layout.

But in general, I think the BIDSLayout is a good singleton for most of the use-cases (counter-example: meta-analyses?).

I've been considering whether making the bids_dir argument to `collect_*` optional. If not None, then create a new BIDSLayout, otherwise use the global. Let me know if you think that would address some of your concerns.

effigies:
Some of this I guess is what niworkflows is supposed to be, which is to me a bunch of pretty much isolable units. Adding a thread of logic that isn't tied together in an object just kind of breaks my mental model. In the end it's probably not a big deal, but just to go ahead and answer your comments:

> This way, it can be queried at any time throughout the workflow.

Sure. Let's create it in fMRIPrep then. I'd still rather it be a passed variable than a global, but if it needs to be a global there, then it can be.

> But in general, I think the BIDSLayout is a good singleton for most of the use-cases (counter-example: meta-analyses?).

In FitLins I use another one to build up the output derivatives directory, with additional entities that are not yet in the spec. Or you can imagine a situation where you want to separately index a BIDS-Raw dataset and a related BIDS-Derivatives dataset (to reduce the overhead in separate contexts).

Granted, I'm not using niworkflows over there, at this point.

But anyway, I think it's a reasonable decision for an app, a little more tenuous for a library, which is harder to predict the usage of.

> I've been considering whether making the bids_dir argument to collect_* optional. If not None, then create a new BIDSLayout, otherwise use the global. Let me know if you think that would address some of your concerns.

But what if we can pass in a layout *or* a directory that we can index in a new layout? Wouldn't that cover most of the cases?

oesteban:
Okay, I'll see whether we can do this in fMRIPrep proper.

PR #298, opened 2019-02-01
[ENH] Option to sanitize NIfTI headers when writing derivatives

oesteban:


oesteban:
New tests added and passing here: https://circleci.com/workflow-run/20fd74dd-6adf-4723-a831-e16144084758

@effigies can you see any big side effects that prevent from merging this?

niworkflows/interfaces/bids.py
```diff
@@ -255,6 +262,32 @@ def _run_interface(self, runtime):
...
+                units = (curr_units[0] or 'mm', 'sec' if dtype == '_bold' else None)
+                xcodes = (1, 1)  # Derivative in its original scanner space
+                if self.inputs.space:
+                    xcodes = (4, 4) if self.inputs.space in STANDARD_SPACES \
+                        else (2, 2)
+
+                if curr_codes != xcodes or curr_units != units:
+                    self._results['fixed_hdr'][i] = True
+                    hdr.set_qform(hdr.get_qform(), xcodes[0])
+                    hdr.set_sform(hdr.get_sform(), xcodes[1])
```
effigies: Can `qform` and `sform` be different? I think we should just set both to `nii.affine`.

Issue #301, opened 2019-02-06
Add multilabel masks to SignalExtraction 


wiheto:
At the moment SignalExtraction only take a a list of masks at the same size as the input image. The enhancement would be to allow the label_files to be multi-label (i.e different integers for each ROI/parcel) and also check if the label_files needs to be resized prior to signal extraction. 

Will submit PR soon.

oesteban:
Closed via 038d61fbc86ed7031a586175ebd32e48b5f7b8de

PR #303, opened 2019-02-06
[ENH] signal extraction of parcels/rois from single nifti file

wiheto:
Related to enhancement suggested in #301 

Previous behaviour is kept but adds features: (1) resampling of mask if it is not the size of the input image (2) the template can be a single nifti image with multiple non-zero labels.


wiheto:
@oesteban maybe I'm misunderstanding the circleci test error here, but it does not appear to be anything with this PR right? MRICoregRPT doesn't call SignalExtraction from what I can tell.

effigies:
Oh, this is that problem we ran into with #292, where every contributor has to set up Circle to use their own FreeSurfer license file.

I think maybe if we're going to continue down this road it will make more sense to skip tests if the license can't be found.

wiheto:
Thanks. Oscar helped me set up the FS license . Seems like the tests are running now.

oesteban:
Is this ready for us to review?

wiheto:
I believe so. It is only stage 1 of getting time series extraction into fmriprep.

effigies:
Can you undo the style changes not directly related to your changes? It's hard to review with a bunch of extraneous changes.

wiheto:
Oops. I autopep8:ed the document out of flake8 frustration. Fixed.

niworkflows/interfaces/images.py
```diff
@@ -670,25 +670,38 @@ class SignalExtraction(SimpleInterface):
...
         mask_imgs = [nb.load(fname) for fname in self.inputs.label_files]
-        if len(mask_imgs) == 1:
+        multi_index_mask = False
+        if len(mask_imgs) == 1 and len(mask_imgs[0].shape) == 4:
             mask_imgs = nb.four_to_three(mask_imgs[0])
 
-        masks = [mask_img.get_data().astype(np.bool) for mask_img in mask_imgs]
+        for i, mask in enumerate(mask_imgs):
+            if img.shape != mask.shape:
+                mask_imgs[i] = nli.resample_to_img(
```
oesteban: I would enforce that masks and image have the same shape and zooms, raise a RuntimeError otherwise. Let the responsibility of resampling be outside this interface.

niworkflows/interfaces/images.py
```diff
@@ -632,7 +632,7 @@ class SignalExtractionInputSpec(BaseInterfaceInputSpec):
         desc='a 3-D label image, with 0 denoting '
         'background, or a list of 3-D probability '
```
oesteban: These can't be probability maps as they are (and were before this PR) cast to bool. These are masks, let's fix it now that we are at it. Let's add one more input to at least threshold at that value (better than > 0.0 as it currently works).

In the future we can remove the bool casting and use a weighted mean in line 690 instead (or you can tackle that now, which would be much better)

niworkflows/interfaces/images.py
oesteban: That permits the image to be loaded after the masks, as it was before.

niworkflows/interfaces/images.py
```diff
@@ -670,25 +670,38 @@ class SignalExtraction(SimpleInterface):
...
+                mask_imgs[i] = nli.resample_to_img(
+                    mask, img, interpolation='nearest')
 
-        n_masks = len(masks)
+        if len(mask_imgs) > 1:
+            masks = [mask_img.get_data().astype(np.bool)
+                     for mask_img in mask_imgs]
+            n_masks = len(masks)
+        else:
+            masks = mask_imgs[0].get_data()
```
oesteban: why not preparing everything in this block? And what happens when one hemisphere has positive labels and the other negative?

```Python
            labelsmap = mask_imgs[0].get_data()
            labels = np.unique(labelsmap)
            labels = labels[labels != 0]  # drop background (zero)
            masks = [labelsmap == l for l in labels]
```

niworkflows/interfaces/images.py
```diff
@@ -670,25 +670,38 @@ class SignalExtraction(SimpleInterface):
...
+            uniquevals = np.unique(masks)
+            n_masks = len(uniquevals[uniquevals > 0])
+            multi_index_mask = True
 
         if n_masks != len(self.inputs.class_labels):
```
oesteban:
```suggestion
         if len(masks) != len(self.inputs.class_labels):
```

niworkflows/interfaces/images.py
```diff
@@ -670,25 +670,38 @@ class SignalExtraction(SimpleInterface):
...
             raise ValueError("Number of masks must match number of labels")
 
-        img = nb.load(self.inputs.in_file)
-
         series = np.zeros((img.shape[3], n_masks))
 
         data = img.get_data()
         for j in range(n_masks):
```
oesteban:
```suggestion
         for j, mask in enumerate(masks):
             series[:, j] = data[mask, :].mean(axis=0)
```

niworkflows/interfaces/images.py
```diff
@@ -670,25 +670,38 @@ class SignalExtraction(SimpleInterface):
...
-            series[:, j] = data[masks[j], :].mean(axis=0)
-
+            if multi_index_mask:
```
oesteban: this would go away, if you followed along

niworkflows/interfaces/images.py
```diff
@@ -670,25 +670,38 @@ class SignalExtraction(SimpleInterface):
...
 
-        masks = [mask_img.get_data().astype(np.bool) for mask_img in mask_imgs]
+        for i, mask in enumerate(mask_imgs):
+            if img.shape != mask.shape:
+                mask_imgs[i] = nli.resample_to_img(
+                    mask, img, interpolation='nearest')
 
-        n_masks = len(masks)
+        if len(mask_imgs) > 1:
+            masks = [mask_img.get_data().astype(np.bool)
```
oesteban: Let's add an input with 0.5 by default, wdyt about prob_thres?
```suggestion
            masks = [mask_img.get_data() > self.inputs.prob_thres
```

niworkflows/interfaces/images.py
```diff
@@ -632,7 +632,7 @@ class SignalExtractionInputSpec(BaseInterfaceInputSpec):
         desc='a 3-D label image, with 0 denoting '
         'background, or a list of 3-D probability '
         'maps (one per label) or the equivalent 4D '
-        'file.')
+        'file or a 3-D label image containing parcels for each non-zero value.')
```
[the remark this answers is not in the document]
wiheto: So I misunderstood "3-D label image" a tad confusing. But I'll just remove what I wrote.

niworkflows/interfaces/images.py
wiheto: NotImplementedError gets raised if mask and img are of different sizes.

niworkflows/interfaces/images.py
wiheto: At the moment I have added an input if probability threshold input which defaults at 0.5. But can change this to 0 if you want (which is what I think you are suggesting here).

niworkflows/interfaces/images.py
wiheto: So everything you suggested should be added now.

wiheto:
Finally got around to this. Really wasnt a lot to do. Sorry it took so long!!

effigies:
I suspect the test is failing because you need to merge master. Could you give that a try?

wiheto:
That did it. 

@oesteban I think this is done.

niworkflows/interfaces/images.py
```diff
@@ -629,10 +629,15 @@ class SignalExtractionInputSpec(BaseInterfaceInputSpec):
     label_files = InputMultiPath(
         File(exists=True),
         mandatory=True,
-        desc='a 3-D label image, with 0 denoting '
-        'background, or a list of 3-D probability '
+        desc='a 3D label image, with 0 denoting '
+        'background, or a list of 3D probability '
         'maps (one per label) or the equivalent 4D '
         'file.')
+    prob_thres = traits.Float(
```
effigies:
```suggestion
    prob_thres = traits.Range(
        low=0.0,
        high=1.0,
```

niworkflows/interfaces/images.py
```diff
@@ -629,10 +629,15 @@ class SignalExtractionInputSpec(BaseInterfaceInputSpec):
...
+        0.5,
```
effigies:
```suggestion
        value=0.5,
```

oesteban:
Hi @wiheto, could you commit @effigies' suggestions? I've tried myself and I don't seem to be allowed to push to your repo.

wiheto:
Done. Sorry I missed the suggestion!

oesteban:
Thanks!

effigies:
Just a note that the new behavior didn't get tested: https://codecov.io/gh/poldracklab/niworkflows/commit/038d61fbc86ed7031a586175ebd32e48b5f7b8de#D1-696...700

PR #304, opened 2019-02-07
[FIX] Improve bounding box computation from masks

oesteban:
Before (fieldmap of ds054 on fmriprep's reports at Circle):
before
After:
after

### DIFF
--- a/niworkflows/viz/utils.py
+++ b/niworkflows/viz/utils.py
@@ -201,12 +201,23 @@ def cuts_from_bbox(mask_nii, cuts=3):
     from nibabel.affines import apply_affine
 
     mask_data = mask_nii.get_data() > 0.0
+
+    # First, project the number of masked voxels on each axes
     ijk_counts = [
-        mask_data.sum(2).sum(0),  # project to i
-        mask_data.sum(2).sum(1),  # project to j
-        mask_data.sum(1).sum(0),  # project to k
+        mask_data.sum(2).sum(0),  # project to i (sagittal)
+        mask_data.sum(2).sum(1),  # project to j (coronal)
+        mask_data.sum(1).sum(0),  # project to k (axial)
     ]
 
+    # If all voxels are masked in a slice (say that happens at k=10),
+    # then the value for ijk_counts for the projection to k (ie. ijk_counts[2])
+    # at that element of the orthogonal axes (ijk_counts[2][10]) is
+    # the total number of voxels in that slice (ie. Ni x Nj).
+    # Here we define some thresholds to consider the plane as "masked"
+    # The thresholds vary because of the shape of the brain
+    # I have manually found that for the axial view requiring 30%
+    # of the slice elements to be masked drops almost empty boxes
+    # in the mosaic of axial planes (and also addresses #281)
     ijk_th = [
         int((mask_data.shape[1] * mask_data.shape[2]) * 0.2),   # sagittal
         int((mask_data.shape[0] * mask_data.shape[2]) * 0.0),   # coronal
--- a/niworkflows/viz/utils.py
+++ b/niworkflows/viz/utils.py
@@ -225,14 +225,19 @@ def cuts_from_bbox(mask_nii, cuts=3):
     ]
 
     vox_coords = []
-    for c, th in zip(ijk_counts, ijk_th):
+    for ax, (c, th) in enumerate(zip(ijk_counts, ijk_th)):
         B = np.argwhere(c > th)
-        # Avoid too narrow selections of cuts
-        if th > 0 and (B.min() + cuts + 1) >= B.max():
+        if B.size:
+            smin, smax = B.min(), B.max()
+
+        # Avoid too narrow selections of cuts (very small masks)
+        if not B.size or (th > 0 and (smin + cuts + 1) >= smax):
             B = np.argwhere(c > 0)
-        ijk = (B.min(), B.max())
-        inc = (ijk[1] - ijk[0]) / (cuts + 1)
-        vox_coords.append([ijk[0] + (i + 1) * inc for i in range(cuts)])
+
+        # Resort to full plane if mask is seemingly empty
+        smin, smax = B.min(), B.max() if B.size else (0, mask_data.shape[ax])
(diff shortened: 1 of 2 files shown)


niworkflows/viz/utils.py
```diff
@@ -199,20 +199,31 @@ def extract_svg(display_object, dpi=300, compress='auto'):
...
+        mask_data.sum(2).sum(0),  # project to i
+        mask_data.sum(2).sum(1),  # project to j
+        mask_data.sum(1).sum(0),  # project to k
+    ]
+
+    ijk_th = [
+        int((mask_data.shape[1] * mask_data.shape[2]) * 0.2),   # sagittal
+        int((mask_data.shape[0] * mask_data.shape[2]) * 0.0),   # coronal
+        int((mask_data.shape[0] * mask_data.shape[1]) * 0.3),   # axial
+    ]
```
effigies: This is opaque to me. `ijk_th` reads like threshold, but I don't see how this should produce a threshold.

oesteban: Added some comments :+1:

oesteban:
This patch still does not address the real underlying problem (it would involve resampling the mask into an affine that is aligned with the canonical axes) but it does the job.

oesteban:
@effigies, please let me know if I managed to clarify the idea with the comments. If so, feel free to merge.

niworkflows/viz/utils.py
```diff
@@ -199,20 +199,45 @@ def extract_svg(display_object, dpi=300, compress='auto'):
...
-    mask_data = mask_nii.get_data()
-    B = np.argwhere(mask_data > 0)
-    start_coords = B.min(0)
-    stop_coords = B.max(0) + 1
+
+    mask_data = mask_nii.get_data() > 0.0
+
+    # First, project the number of masked voxels on each axes
+    ijk_counts = [
+        mask_data.sum(2).sum(0),  # project to i (sagittal)
```
effigies: I see, this is sagittal in the sense that the longitudinal-vertical plane is sagittal and so you're collapsing each sagittal plane onto the transverse (i) axis. If that's the case, and we're working with transverse-longitudinal-vertical images, then, we should be collapsing the `j` and `k` axes:

```suggestion
        mask_data.sum(2).sum(1),  # project sagittal planes to transverse (i) axis
```

niworkflows/viz/utils.py
```diff
@@ -199,20 +199,45 @@ def extract_svg(display_object, dpi=300, compress='auto'):
...
+        mask_data.sum(2).sum(1),  # project to j (coronal)
```
effigies: Likewise

```suggestion
        mask_data.sum(2).sum(0),  # project coronal planes to to longitudinal (j) axis
```

niworkflows/viz/utils.py
```diff
@@ -199,20 +199,45 @@ def extract_svg(display_object, dpi=300, compress='auto'):
...
+        mask_data.sum(1).sum(0),  # project to k (axial)
```
effigies: And just to be consistent:

```suggestion
        mask_data.sum(1).sum(0),  # project axial planes to vertical (k) axis
```

niworkflows/viz/utils.py
```diff
@@ -199,20 +199,45 @@ def extract_svg(display_object, dpi=300, compress='auto'):
...
+        ijk = (B.min(), B.max())
+        inc = (ijk[1] - ijk[0]) / (cuts + 1)
+        vox_coords.append([ijk[0] + (i + 1) * inc for i in range(cuts)])
 
     ras_coords = []
     for cross in np.array(vox_coords).T:
-        ras_coords.append(apply_affine(mask_nii.affine, cross).tolist())
-
+        ras_coords.append(apply_affine(
+            mask_nii.affine, cross).tolist())
```
effigies: Is it possible that the problem is actually just that we don't make sure these are integers, being sure to round up at the bottom and down at the top?

niworkflows/viz/utils.py
```diff
@@ -199,20 +199,45 @@ def extract_svg(display_object, dpi=300, compress='auto'):
 def cuts_from_bbox(mask_nii, cuts=3):
     """Finds equi-spaced cuts for presenting images"""
     from nibabel.affines import apply_affine
-    mask_data = mask_nii.get_data()
-    B = np.argwhere(mask_data > 0)
-    start_coords = B.min(0)
-    stop_coords = B.max(0) + 1
```
effigies: Or maybe this should have just been `stop_coords = B.max(0)`?

oesteban: This indeed seemingly fixed the issue (that was the first thing I tried). However, being at it, I wanted to improve the visualization because such a liberal calculation usually gave almost empty views of the axial plane in many images.

niworkflows/viz/utils.py
oesteban: nilearn takes physical coordinates, so I doubt this is the problem.

effigies: Right, but if we round the physical coordinates to outside the mask, then we might get undefined behavior. We certainly get something.

oesteban: A warning saying that no contours were found was thrown. Nilearn used to not plot anything in that case, but after some release the weird contours would show up. The thresholds solution should take care of most of the cases where this rounding happened, as we don't compute the bounding box of the mask, but the bounding box of the mask with a decent amount of masked pixels instead. The tests were failing with very small masks, but 55aa451 should fix that falling back to the old calculation (with your suggestion in https://github.com/poldracklab/niworkflows/pull/304#discussion_r254707942).

That said, the issue I'm afraid, is that nilearn rotates the images (and the masks) to align with canonical axes in visualization. As the brain is not perfectly spherical, these rotations might force the coordinates outside the brain if you calculate them exactly at the brain surface (as we were doing) for very tilted brains.

I think that this implementation will basically satisfy 99.9% of the cases and the remaining 0.1% will have the weird contours - which are confusing but not harmful.

oesteban:
Reports are looking good - thumbs up?

PR #305, opened 2019-02-07
[MNT] Deprecate getters module

### LINKED ISSUE #250: Factor out OSF getter into small package
The OSF dataset getters are a pretty useful piece of code that would be worth factoring into a small package that we import here. This would make transitioning to niflows simpler, and may be useful for other scientific packages.

I would picture this package mostly consisting of the equivalent of `get_dataset()`:

https://github.com/poldracklab/niworkflows/blob/b57a21974a05fb7148ba310701c847fc83a63122/niworkflows/data/getters.py#L52-L76

And then we provide a standard way to register a batch of names with file IDs and hashes.

oesteban:
Closes #250


effigies:
This looks reasonable. Have you looked at `nibabel.deprecated.deprecate_with_version`?

[Example](https://github.com/nipy/nibabel/blob/30c0bc561e0f34763c415a3fdc2b39cf0789a4ea/nibabel/gifti/giftiio.py#L16-L18):
```Python
@deprecate_with_version('giftiio.read function deprecated. '
                        "Use nibabel.load() instead.",
                        '2.1', '4.0')
def read(filename):
```

oesteban:
That's pretty neat. I'll modify accordingly. Thanks

oesteban:
I can't use your suggestion at the moment (nipy/nibabel#723). Should we just merge and use `nibabel.deprecated.deprecate_with_version` when fixed?

effigies:
Sure. I'm not sure it'll get fixed... It would need to maintain current compatibility with nibabel version strings.

Issue #306, opened 2019-02-11
Regression of int-type run entity on ReadSidecarJSON


oesteban:
My latest re-implementation with pybids reintroduced this problem. A unit-test needs to be added.

```
Traceback (most recent call last):
  File "/usr/local/miniconda/lib/python3.7/site-packages/nipype/pipeline/plugins/multiproc.py", line 69, in run_node
    result['result'] = node.run(updatehash=updatehash)
  File "/usr/local/miniconda/lib/python3.7/site-packages/nipype/pipeline/engine/nodes.py", line 473, in run
    result = self._run_interface(execute=True)
  File "/usr/local/miniconda/lib/python3.7/site-packages/nipype/pipeline/engine/nodes.py", line 557, in _run_interface
    return self._run_command(execute)
  File "/usr/local/miniconda/lib/python3.7/site-packages/nipype/pipeline/engine/nodes.py", line 637, in _run_command
    result = self._interface.run(cwd=outdir)
  File "/usr/local/miniconda/lib/python3.7/site-packages/nipype/interfaces/base/core.py", line 371, in run
    outputs = self.aggregate_outputs(runtime)
  File "/usr/local/miniconda/lib/python3.7/site-packages/nipype/interfaces/base/core.py", line 472, in aggregate_outputs
    raise error
  File "/usr/local/miniconda/lib/python3.7/site-packages/nipype/interfaces/base/core.py", line 465, in aggregate_outputs
    setattr(outputs, key, val)
  File "/usr/local/miniconda/lib/python3.7/site-packages/traits/trait_handlers.py", line 172, in error
    value )
traits.trait_errors.TraitError: The 'run' trait of a ReadSidecarJSONOutputSpec instance must be a unicode string, but a value of 1  was specified.
```

oesteban:
In the meantime, if you get this error, you can easily workaround it by renaming your `run` entities to have zero padding, e.g.:
```
fmap/sub-01_run-1_phasediff.json -> fmap/sub-01_run-01_phasediff.json
fmap/sub-01_run-1_phasediff.nii.gz -> fmap/sub-01_run-01_phasediff.nii.gz
fmap/sub-01_run-2_phasediff.json -> fmap/sub-01_run-02_phasediff.json
fmap/sub-01_run-2_phasediff.nii.gz -> fmap/sub-01_run-02_phasediff.nii.gz
```

effigies:
[...]

PR #307, opened 2019-02-12
[FIX] Do not cast ``run`` BIDS-entity to string

oesteban:
Addresses #306


oesteban:
This patch is fully compatible with current fMRIPrep and sMRIPrep and tests are passing. Merging.

Issue #242, opened 2018-05-21
Suboptimal scaling of some confounds in reports


oesteban:
[from poldracklab/fmriprep#1143]

![image](https://user-images.githubusercontent.com/238759/40269233-e4cd8862-5b40-11e8-8ba4-9c5a79a2913b.png)

It's hard to see variation in WM and CSF with this scaling.

effigies:
Been thinking about this a little. Let's use -1 to 1 to describe the vertical axis, and for the time series `X`, let `a = min`, `b = max`, `m = mean`.

So `(X - m) / m` will be centered at the mean and scaled by the mean. In this scale, a change of 10% of the mean will appear as +/-0.1, or a 20th of the [-1, 1] range. If we want that to be a quarter of the plot range, then multiply by 5, so the effective time series is `5 (X - m) / m`.

Thus I think a viable scaling option takes the form `(X - m) / max(m / P, m - a, b - m)`, where `P` is an inflation factor from % change of mean to % distance to y-axis bounds. `m - a` and `b - m` ensure that the max and min do not go off the plot.

(I'll note that we definitely do not want to scale by variance, as that would make comparing variance between BOLD series impossible.)

wiheto:
So I think this can be solved by multiplying the range of values with a scalar instead of the min/max

Something like: 

```
nonnan_range = (nonnan.max() - nonnan.min())
def_ylims = [nonnan.min() - 0.1 * nonnan_range, 1.1 * nonnan_range]
```

Maybe the scalar still needs to be tuned, but this should stop the squashing phenomena, I think.

wiheto:
Did some testing, this scaling works quite well with varying y-axis ranges. 

Anyone objects? Can submit PR otherwise.

oesteban:
No objections at all, please send that PR :)

oesteban:
Fixed in https://github.com/poldracklab/niworkflows/commit/a71c8346da5e100cb596e4f91a54a1f769d78dfc

PR #310, opened 2019-02-13
[FIX] Scaling of confound fix

wiheto:
Realized I forgot to submit this. 

Fixing the scaling issue in #242. 

Scales the confounds based on the range of the values rather than scaling the min and max. 

if min value was 2 and max value was 5:
Before. ylim: 1.8 - 5.5 (plot displayed in 81% of ylim range)
Now, ylim: 1.7 - 5.3 (plot displayed in 83% of ylim range)

if min value was 900 and max value was 1000:
Before, ylim: 810 - 1100 (plot displayed in 34% of ylim range <- when squashing was occurring)
Now, ylim: 890 - 1010 (plot displayed in 83% of ylim range)

### DIFF
--- a/niworkflows/viz/plots.py
+++ b/niworkflows/viz/plots.py
@@ -479,7 +479,8 @@ def confoundplot(tseries, gs_ts, gs_dist=None, name=None,
     nonnan = tseries[~np.isnan(tseries)]
     if nonnan.size > 0:
         # Calculate Y limits
-        def_ylims = [nonnan.min() - 0.1 * abs(nonnan.min()), 1.1 * nonnan.max()]
+        valrange = (nonnan.max() - nonnan.min())
+        def_ylims = [nonnan.min() - 0.1 * nonnan_range, nonnan.max() + 0.1 * nonnan_range]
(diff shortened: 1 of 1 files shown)


effigies:
Thanks. Could you show a plot before and after?

wiheto:
Figures below based on: 

```
import numpy as np 

# mean and std taken from examples in issues
ts1 = np.random.normal(872, 5, 500)
ts2 = np.random.normal(1, 0.1, 500)
ts3 = np.random.normal(0.163, 1.29, 500)
```

### Previously

confoundplot_old

### Now 

confoundplot_new

wiheto:
I think percentiles could be problematic if there is a large spike in the data.

oesteban:
Because those would get clipped right?

wiheto:
Yepp. Using the full range assures the min/max are present in the plot. But then the obvious downside is that one large confound can squash the entire data. 

One solution to this however is to allow extreme values to overlap into the subplot above. Something like:  https://seaborn.pydata.org/examples/kde_ridgeplot.html

PR #314, opened 2019-02-23
patch for empty mask after erosion

rciric:
If erosion (as implemented in `TPM2ROI`) results in an empty mask, the original mask is instead returned, and a warning report printed. I'm not sure if the warning report is a good idea, and I can revert that aspect if not. Thus, I haven't yet updated my `fmriprep` PR to sink it.


effigies:
There should definitely be a warning in this case. My main question is whether it's better to have some mask or to handle empty masks gracefully. I'm leaning toward the latter, as time series derived from un-eroded masks are extremely suspect.

rciric:
I updated the nipype CompCor PR to handle empty masks instead, so this can probably be closed.

PR #324, opened 2019-03-27
ENH: Ensure brain mask's conformity

oesteban:
This PR adds a node to the `antsBrainExtraction` workflow to ensure that the mask produced matches the xform matrices and codes of the original reference image, and sets some other relevant metadata of the headers.


oesteban:
This should be pretty harmless, any chance you can have a look @effigies ?

PR #330, opened 2019-04-04
FIX: "NONE of the components..." banner was printed even when no AROMA file was present

oesteban:
Builds on #292 to improve the way these plots are written. Also adds a `[noise]` tag in the title of those components classified as such.

### DIFF
--- a/niworkflows/tests/test_viz.py
+++ b/niworkflows/tests/test_viz.py
@@ -27,12 +27,15 @@ def test_plot_melodic_components():
     import numpy as np
     # save the artifacts
     save_artifacts = os.getenv('SAVE_CIRCLE_ARTIFACTS', False)
+    all_noise = 'melodic_all_noise.svg'
+    no_noise = 'melodic_no_noise.svg'
+    no_classified = 'melodic_no_classified.svg'
+
     if save_artifacts:
         all_noise = os.path.join(save_artifacts, 'melodic_all_noise.svg')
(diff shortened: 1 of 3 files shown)


niworkflows/tests/test_viz.py
```diff
@@ -75,3 +78,9 @@ def test_plot_melodic_components():
     viz.utils.plot_melodic_components(melodic_dir, 'in_file.nii.gz', tr=2.0,
                                       report_mask='report_mask.nii.gz',
```
jdkent: post-hoc review, I see I wrote this test with incorrect assumptions, the `noise_components_file` [should exist](https://github.com/maartenmennes/ICA-AROMA/blob/b53bd93c6cad34655d5bc122168351eeb6652e1c/ICA_AROMA_functions.py#L501-L506), but be empty if there were no noise components identified.

I noticed this because both [no_noise](https://605-74174878-gh.circle-artifacts.com/0/tmp/tests/melodic_no_noise.svg) has no warning printed (but should), but [no_classified](https://605-74174878-gh.circle-artifacts.com/0/tmp/tests/melodic_no_classified.svg) is correctly printed without a warning.

I can open a quick pull request to rectify this mistake since you didn't touch this code.

jdkent: Oops, saw that you already caught and fixed that mistake with [this commit](https://github.com/poldracklab/niworkflows/commit/db69566419afb9b9152b5f15ae5ed0f9601b4b6d). sorry for the noise.

PR #332, opened 2019-04-05
ENH: ``DerivativesDataSink`` now accepts metadata dictionaries too

oesteban:
Add a new ``meta_dict`` input to ``DerivativesDataSink``s, so that this can be done:

```Python
    >>> dsink = DerivativesDataSink(base_directory='/path/to/BIDS/derivatives', check_hdr=False,
    ...                             SkullStripped=True)
    >>> dsink.inputs.in_file = 'output_file_to_store.nii.gz'
    >>> dsink.inputs.source_file = 'sub-10/func/sub-10_task-rest_run-1_bold.nii.gz'
    >>> dsink.inputs.keep_dtype = True
    >>> dsink.inputs.desc = 'preproc'
    >>> dsink.inputs.RepetitionTime = 0.75
    >>> dsink.inputs.meta_dict = {'RepetitionTime': 1.75, 'SkullStripped': False, 'Z': 'val'}
```

In the example, ``RepetitionTime`` and ``SkullStripped`` will be overriden by the values passed in dynamically, and ``Z`` will be added to the output metadata.

cc @rciric

### DIFF
--- a/niworkflows/interfaces/bids.py
+++ b/niworkflows/interfaces/bids.py
@@ -241,6 +241,7 @@ class DerivativesDataSinkInputSpec(DynamicTraitedSpec, BaseInterfaceInputSpec):
     in_file = InputMultiObject(File(exists=True), mandatory=True,
                                desc='the object to be saved')
     keep_dtype = traits.Bool(False, usedefault=True, desc='keep datatype suffix')
+    meta_dict = traits.DictStrAny(desc='an input dictionary containing metadata')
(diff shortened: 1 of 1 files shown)


rciric:
Looks good -- thanks!

I realised though that the confounds metadata will have to be merged into a single file before I pass it to the data sink (so that it can be associated with the confounds data).

oesteban:
You mean that in an ideal world you would need two inputs of the ``meta_dict`` kind?

PR #337, opened 2019-04-10
ENH: Create informative HTML reportlet on missing MELODIC mix

### LINKED ISSUE #336: Handle reports gracefully when MELODIC fails to converge
### Summary
MELODIC may not converge, and when that happens, the log is closed with a message:

```
No convergence -- giving up 
```

Since that is a possible (and valid) outcome, the interface should handle it correctly (instead of crashing).

### Actual behavior
The interface crashes with:

```
  File "/usr/local/miniconda/lib/python3.6/site-packages/nipype/pipeline/plugins/multiproc.py", line 69, in run_node
    result['result'] = node.run(updatehash=updatehash)
  File "/usr/local/miniconda/lib/python3.6/site-packages/nipype/pipeline/engine/nodes.py", line 471, in run
    result = self._run_interface(execute=True)
  File "/usr/local/miniconda/lib/python3.6/site-packages/nipype/pipeline/engine/nodes.py", line 555, in _run_interface
    return self._run_command(execute)
  File "/usr/local/miniconda/lib/python3.6/site-packages/nipype/pipeline/engine/nodes.py", line 635, in _run_command
    result = self._interface.run(cwd=outdir)
  File "/usr/local/miniconda/lib/python3.6/site-packages/nipype/interfaces/base/core.py", line 523, in run
    runtime = self._post_run_hook(runtime)
  File "/usr/local/miniconda/lib/python3.6/site-packages/niworkflows/interfaces/segmentation.py", line 131, in _post_run_hook
    return super(MELODICRPT, self)._post_run_hook(runtime)
  File "/usr/local/miniconda/lib/python3.6/site-packages/nipype/interfaces/mixins/reporting.py", line 47, in _post_run_hook
    self._generate_report()
  File "/usr/local/miniconda/lib/python3.6/site-packages/niworkflows/interfaces/segmentation.py", line 120, in _generate_report
    report_mask=self.inputs.report_mask)
  File "/usr/local/miniconda/lib/python3.6/site-packages/niworkflows/viz/utils.py", line 527, in plot_melodic_components
    timeseries = np.loadtxt(os.path.join(melodic_dir, "melodic_mix"))
  File "/usr/local/miniconda/lib/python3.6/site-packages/numpy/lib/npyio.py", line 860, in loadtxt
    fh = iter(open(fname))
FileNotFoundError: [Errno 2] No such file or directory: '/data/test_katenuss_mriqc/wdir/workflow_enumerator/funcMRIQC/_in_file_..data..hartleylab..memFreqMRI112..sub-112..func..sub-112_task-freq1_acq-normal_run-01_echo-3_bold.nii.gz/ICA/melodic_mix'
```

### Expected behavior
The interface finishes in a controlled way.

### How to replicate the behavior
To replicate this, it is necessary to run the interface on a dataset were convergence is not reached.

### Script/Workflow details
Comes from MRIQC (see poldracklab/mriqc#779 for a larger context)

----

Above text copied from nipy/nipype#2914.

effigies:
This is currently a proof-of-concept. We should decide what the appropriate output should be. This is an SVG file, so I don't think it's quite as easy as an HTML snippet... @oesteban What would you like the output to be?

Closes #336.

Addresses poldracklab/mriqc#779.

### DIFF
--- a/niworkflows/interfaces/segmentation.py
+++ b/niworkflows/interfaces/segmentation.py
@@ -114,7 +114,10 @@ class MELODICRPT(reporting.ReportCapableInterface, fsl.MELODIC):
         from niworkflows.viz.utils import plot_melodic_components
         mix = os.path.join(self._melodic_dir, "melodic_mix")
         if not os.path.exists(mix):
-            open(self.inputs.out_report, 'w').close()
+            self._out_report = self.inputs.out_report.replace('svg', 'html')
+            snippet = '<h4>MELODIC did not converge, no output</h4>'
+            with open(self._out_report, 'w') as fobj:
+                fobj.write(snippet)
             return
         plot_melodic_components(melodic_dir=self._melodic_dir,
                                 in_file=self.inputs.in_files[0],
--- a/niworkflows/interfaces/segmentation.py
+++ b/niworkflows/interfaces/segmentation.py
@@ -112,6 +112,10 @@ class MELODICRPT(reporting.ReportCapableInterface, fsl.MELODIC):
 
     def _generate_report(self):
         from niworkflows.viz.utils import plot_melodic_components
+        mix = os.path.join(self._melodic_dir, "melodic_mix")
+        if not os.path.exists(mix):
+            open(self.inputs.out_report, 'w').close()
+            return
         plot_melodic_components(melodic_dir=self._melodic_dir,
                                 in_file=self.inputs.in_files[0],
                                 tr=self.inputs.tr_sec,


oesteban:
I think just writing an HTML snippet should work out as well.

The problem I see to this solution is that we really want to grep the MELODIC log and find the "did not converge, giving up" message (I don't remember if it is exactly like this).

I've seen MELODIC being killed by the OOM and then of course the melodic_mix folder does not exist. The nipype interface breaks exactly the same way in both cases (did not converge and killed by kernel).

effigies:
Okay, commented over on nipy/nipype#2914. That seems to be a separate issue from the reporting, though, as it's the escalation of an error case to a raised exception.

So if an HTML snippet is okay, then we'll need to verify that MRIQC/fMRIPrep can accommodate receiving either HTML or SVG as the reportlet.

PR #253, opened 2018-09-14
[WIP] resample sbref

chrisgorgo:
Resample SBRef images if they don't match their BOLD files. Makes it more robust - SBRef with a non-matching grid is unlikely, but not inconceivable.

PR #338, opened 2019-04-11
FIX: Do not generate 4D references out of 4D single-band references (SBRefs)

oesteban:
``EstimateReferenceImage`` generates 4D references if 4D SBRefs are supplied as input. This patch checks the dimensionality of SBRefs and generates an average if the SBRef is 4D.

Fixes poldracklab/fmriprep#1579.

Intimately related to #253, although I'm not sure the proposed resampling there is absolutely necessary.
It could be detrimental for ``bbregister`` in co-registration.
It is very likely necessary for head-motion correction with either ``mcflirt`` or ``3dVolReg``.

### DIFF
--- a/niworkflows/interfaces/registration.py
+++ b/niworkflows/interfaces/registration.py
@@ -370,45 +370,55 @@ class EstimateReferenceImage(SimpleInterface):
     output_spec = EstimateReferenceImageOutputSpec
 
     def _run_interface(self, runtime):
-        in_nii = nb.load(self.inputs.in_file)
-        n_volumes_to_discard = _get_vols_to_discard(in_nii)
+        ref_name = self.inputs.in_file
+        ref_nii = nb.load(ref_name)
+        n_volumes_to_discard = _get_vols_to_discard(ref_nii)
 
         self._results["n_volumes_to_discard"] = n_volumes_to_discard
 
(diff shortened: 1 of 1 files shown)


oesteban:
Also fixes poldracklab/fmriprep#1463.

effigies:
Have we checked the spec to make sure sbrefs are legal to have as 4d?

niworkflows/interfaces/registration.py
```diff
@@ -370,45 +370,55 @@ class EstimateReferenceImage(SimpleInterface):
...
+        out_ref_fname = os.path.join(runtime.cwd, "ref_bold.nii.gz")
         if isdefined(self.inputs.sbref_file):
-            self._results['ref_image'] = self.inputs.sbref_file
-            return runtime
+            out_ref_fname = os.path.join(runtime.cwd, "ref_sbref.nii.gz")
+            ref_name = self.inputs.sbref_file
+            ref_nii = nb.squeeze_image(nb.load(ref_name))
+
+            # If reference is only 1 volume, return it directly
+            if ref_nii.get_data().ndim == 3:
```
effigies:
```suggestion
            if len(ref_nii.shape) == 3:
```

oesteban:
> Are you sure we don't just want the first volume, though, rather than taking the average?

The averaged should have much higher SNR without a cost in contrast. I think it makes more sense and computationally is not a big deal.

Issue #333, opened 2019-04-09
ReadSidecarJSON: "Some connections were not found"


oesteban:
A design flaw in ReadSidecarJSON causes this issue. The following example should work:

```Python
from nipype.pipeline import engine as pe
from nipype.interfaces import utility as niu
from niworkflows.interfaces.bids import ReadSidecarJSON

n = pe.Node(ReadSidecarJSON(fields=['RepetitionTime'], in_file='bids-file.nii.gz'), name='node')
o = pe.Node(niu.IdentityInterface(fields=['RepetitionTime']), name='o')
wf = pe.Workflow(name='json')
wf.connect([
    (n, o, [('RepetitionTime', 'RepetitionTime')]),
])
```

but fails with:

```
---------------------------------------------------------------------------
Exception                                 Traceback (most recent call last)
<ipython-input-18-bcc725938bc3> in <module>
      1 wf.connect([
----> 2     (n, o, [('RepetitionTime', 'RepetitionTime')]),
      3 ])

~/workspace/nipype/nipype/pipeline/engine/workflows.py in connect(self, *args, **kwargs)
    212         if not_found:
    213             raise Exception(
--> 214                 '\n'.join(['Some connections were not found'] + infostr))
    215 
    216         # turn functions into strings

Exception: Some connections were not found
Module node has no output called RepetitionTime
```

PR #340, opened 2019-04-18
FIX: allow pipelining dynamic outputs of ``ReadSidecarJSON``

oesteban:
Added two regression tests.

See #333 .

### DIFF
--- a/niworkflows/interfaces/bids.py
+++ b/niworkflows/interfaces/bids.py
@@ -21,6 +21,7 @@ from nipype.interfaces.base import (
     File, Directory, InputMultiObject, OutputMultiObject, Str,
     SimpleInterface,
 )
+from nipype.interfaces.io import add_traits
 from templateflow.api import templates as _get_template_list
 from ..utils.bids import BIDS_NAME, _init_layout
 from ..utils.misc import splitext as _splitext, _copy_any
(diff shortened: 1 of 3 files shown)


niworkflows/interfaces/tests/test_bids.py
```diff
@@ -112,3 +112,35 @@ def test_DerivativesDataSink_t1w(tmpdir, space, size, units, xcodes, fixed):
...
+                           'sub-100185_phasediff.nii.gz')
+    o = pe.Node(niu.IdentityInterface(fields=['out_port']), name='o')
+    wf = pe.Workflow(name='json')
+
+    if field in reg_fields:  # This should work
+        wf.connect([
+            (n, o, [(field, 'out_port')]),
+        ])
+    else:
+        with pytest.raises(Exception) as excinfo:
```
effigies: You can check the exception message in one shot:

```suggestion
        with pytest.raises(Exception, match=r'.*Some connections were not found.*'):
```

niworkflows/interfaces/bids.py
```diff
@@ -533,6 +530,22 @@ class ReadSidecarJSON(SimpleInterface):
     layout = None
     _always_run = True
 
+    def __init__(self, fields=None, undef_fields=False, **inputs):
+        super(ReadSidecarJSON, self).__init__(**inputs)
+        self._fields = fields or []
+        self._undef_fields = undef_fields
+
+        if fields:
+            add_traits(self.inputs, fields)
```
effigies: Why are you adding fields to the inputs? You don't seem to test or advertise this functionality.

oesteban: Ups, you are right - this doesn't make any sense.

PR #341, opened 2019-04-19
ENH: Dictionary manipulation / TSV to dict, merge multiple dicts

rciric:
This pull request implements the following:
* `TSV2JSON` can now return a dictionary instead of printing contents to a JSON file. This is used to convert the TSV generated by the nipype `CompCor` interfaces into a dictionary, which can be passed to metadata sinks to produce JSON-formatted metadata. Perhaps this entails renaming the function?
* A new interface, `MergeDict`, merges dictionaries. This is used to combine metadata produced by different confound modelling steps in `fmriprep`. It wouldn't surprise me to learn that there's already an interface somewhere that does this; if so, please let me know so that I can update the code to use that instead. (The node is configured to support ordered output, but I haven't added the code to do that yet because I wasn't sure if there was already a better existing solution.)

### DIFF
--- a/niworkflows/interfaces/utils.py
+++ b/niworkflows/interfaces/utils.py
@@ -710,6 +710,31 @@ class JoinTSVColumns(SimpleInterface):
         return runtime
 
 
+class DictMergeInputSpec(BaseInterfaceInputSpec):
+    in_dicts = traits.List(
+        traits.Either(traits.Dict, traits.Instance(OrderedDict)),
+        desc='Dictionaries to be merged. In the event of a collision, values '
+             'from dictionaries later in the list receive precedence.')
+
+
+class DictMergeOutputSpec(TraitedSpec):
+    out_dict = traits.Either(traits.Dict, traits.Instance(OrderedDict),
+                             desc='Merged dictionary')
+
+
+class DictMerge(SimpleInterface):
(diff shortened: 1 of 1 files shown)


niworkflows/interfaces/utils.py
```diff
@@ -710,15 +710,45 @@ def _run_interface(self, runtime):
...
 
+class DictMergeInputSpec(BaseInterfaceInputSpec):
+    in_dicts = traits.List(
+        traits.Either(traits.Dict, traits.Instance(OrderedDict)),
+        desc='Dictionaries to be merged. In the event of a collision, values '
+             'from dictionaries later in the list receive precedence.')
+
+
+class DictMergeOutputSpec(TraitedSpec):
+    out_dict = traits.Either(traits.Dict, traits.Instance(OrderedDict),
```
oesteban: As it is written right now, out_dict is always a dict.

```suggestion
    out_dict = traits.Dict(desc='Merged dictionary')
```

niworkflows/interfaces/utils.py
```diff
@@ -710,15 +710,45 @@ def _run_interface(self, runtime):
...
+                             desc='Merged dictionary')
```
oesteban: just so you make all changes in one commit:

```suggestion
```

PR #347, opened 2019-04-26
ENH: Add a ``KeySelect`` interface

oesteban:
This interface is akin to Python's ``OrderedDict``. It allows
to demultiplex list of dynamic lists (of same length) at the input
selecting the values corresponding to the requested key.

The implementation is proposed with several doctests ensuring the proper
functionality.

Pinging @satra and @effigies to know whether we should upstream this
interface directly to nipype.interfaces.utility

### DIFF
--- a/niworkflows/interfaces/utility.py
+++ b/niworkflows/interfaces/utility.py
@@ -8,7 +8,7 @@ from __future__ import absolute_import, division, print_function, unicode_litera
 
 from nipype.interfaces.io import add_traits
 from nipype.interfaces.base import (
-    traits, Str, DynamicTraitedSpec, BaseInterface
+    traits, Str, DynamicTraitedSpec, BaseInterface, isdefined
 )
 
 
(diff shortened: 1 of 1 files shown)


satra:
@oesteban -  in general this could be ok, but perhaps something simpler is possible:

so something like the following in a function:

```python
def select(in_dict, key):
    return tuple(in_dict[key])
```
and we could still create the dynamic fields. 

more generally, we should think about what "ops" we provide that replace standard python structures within the context of a workflow.

oesteban:
> @oesteban - in general this could be ok, but perhaps something simpler is possible:
> 
> so something like the following in a function:
> 
> ```python
> def select(in_dict, key):
>     return tuple(in_dict[key])
> ```
>
> and we could still create the dynamic fields.

Not sure this is correct. Maybe part of your reasoning is escaping my understanding, but this interface is like a `nipype.interfaces.utility.Select`, but more general because it accepts arbitrary inputs and they all get selected by the same index (i.e., as if they had `inlist1`, `inlist2`, `inlist3`...). If that feature existed, then I agree that setting an inline function in the connection to the `index` input like:
```Python
def _select(key):
    return SOME_LIST_IN_SCOPE.index(key)
```
 
> more generally, we should think about what "ops" we provide that replace standard python structures within the context of a workflow.
Are you thinking of nipype 2 here, right?

satra:
@oesteban - let's see how this plays out, and we can upstream. at this point, let's wait a bit to get pydra-related pieces stable before considering upstreaming.

PR #348, opened 2019-04-29
FIX: Allow ``native`` grids (i.e. pass-through) for ``GenerateSamplingReference``

oesteban:
This PR makes ``GenerateSamplingReference`` more flexible, accepting a
new input called ``resolution`` so that the decision about whether
generating a new resampling grid or not can be done in the context of
this interface.

### DIFF
--- a/niworkflows/interfaces/utils.py
+++ b/niworkflows/interfaces/utils.py
@@ -22,7 +22,7 @@ from nipype import logging
 from nipype.utils.filemanip import fname_presuffix
 from nipype.utils.misc import normalize_mc_params
 from nipype.interfaces.base import (
-    traits, Str, isdefined, File, InputMultiPath,
+    traits, isdefined, File, InputMultiPath,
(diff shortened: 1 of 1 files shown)


niworkflows/interfaces/utils.py
```diff
@@ -126,6 +126,8 @@ class GenerateSamplingReferenceInputSpec(BaseInterfaceInputSpec):
                              desc='force xform code')
     fov_mask = traits.Either(None, File(exists=True), usedefault=True,
                              desc='mask to clip field of view (in fixed_image space)')
+    resolution = traits.Either('native', Str, traits.Int, None, usedefault=True,
```
effigies: This isn't used except for `'native'`. What's the intended function with `Str` and `Int`?

oesteban: Probably a boolean would be better. I'll change that.

PR #360, opened 2019-05-15
FIX: Refine implementation of plotting confounds correlations

oesteban:
This PR cleans up the code a little bit, and fixes poldracklab/fmriprep#1620

result


oesteban:
This seems to be working out: https://3812-53608443-gh.circle-artifacts.com/0/tmp/ds210/derivatives/fmriprep/sub-02.html

PR #375, opened 2019-07-11
MAINT: Switch to a ``setup.cfg``-based setup

### LINKED ISSUE #373: MAINT: Installation using ``setup.cfg``

effigies:
Validating installation methods and version stamping with Travis. Going to end up with a pretty big matrix (3 versions by 7 environments, at least), but for a project that a lot of our stuff depends on, I think it's worth it.

Closes #373.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -206,7 +206,7 @@ jobs:
           command: |
             python3 setup.py check -r -s
             python3 setup.py sdist bdist_wheel
-            twine check dist/*
+            python3 -m twine check dist/*
       - run:
           name: Validate version
           command: |
@@ -232,7 +232,7 @@ jobs:
           command: |
             python3 setup.py check -r -s
             python3 setup.py sdist bdist_wheel
-            twine check dist/*
+            python3 -m twine check dist/*
       - run:
           name: Validate version
           command: |
@@ -245,7 +245,7 @@ jobs:
       - run:
           name: Upload to PyPi
           command: |
-            twine upload dist/*
+            python3 -m twine upload dist/*
 
 workflows:
   version: 2
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -191,6 +191,30 @@ jobs:
       - store_test_results:
           path: /tmp/tests
 
+  test_package:
+    machine:
+      image: circleci/classic:201711-01
+    working_directory: /tmp/src/niworkflows
+    steps:
+      - checkout
+      - run:
+          name: Install build depends
+          command: python3 -m pip install "setuptools>=30.3.0" twine docutils
+      - run:
+          name: Build and check
+          command: |
+            python3 setup.py check -r -s
+            python3 setup.py sdist bdist_wheel
+            twine check dist/*
+      - run:
+          name: Validate version
+          command: |
+            THISVERSION=$( python3 get_version.py )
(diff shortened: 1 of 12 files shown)


versioneer.py
```diff
@@ -1404,6 +1405,12 @@ def get_versions(verbose=False):
 
     Returns dict with two keys: 'version' and 'full'.
     """
+    ver_file = Path(__file__).parent / 'fmriprep' / 'VERSION'
+
+    if ver_file.is_file():
+        ver = ver_file.read_text().splitlines()[0].strip()
+        return {"version": ver, 'full': ver}
+
```
effigies: I think this part probably isn't necessary. Since we don't distribute containers, we don't really need the `VERSION` hack.

versioneer.py
```diff
@@ -1404,6 +1405,12 @@ def get_versions(verbose=False):
 
     Returns dict with two keys: 'version' and 'full'.
     """
+    ver_file = Path(__file__).parent / 'fmriprep' / 'VERSION'
```
oesteban:
```suggestion
    ver_file = Path(__file__).parent / 'niworkflows' / 'VERSION'
```

versioneer.py
oesteban: probably not, I agree

.circleci/config.yml
```diff
@@ -193,22 +191,61 @@ jobs:
...
+            python3 setup.py sdist
+            python3 -m twine check dist/*
+      - run:
+          name: Validate version
+          command: |
+            THISVERSION=$( python3 get_version.py )
+            python3 -m pip install dist/*.tar.gz
+            mkdir empty
+            cd empty
+            INSTALLED=$( python3 -c 'import niworkflows; print(niworkflows.__version__)' )
```
effigies: This fails if we build and install from a wheel... I'm not sure why. Perhaps we need to upgrade wheel?

effigies:
@oesteban This should be set.

setup.cfg
```diff
@@ -1,3 +1,84 @@
...
+    pytest
+    pytest-xdist
+    pytest-cov == 2.5.1
+    coverage
+packages = find:
+include_package_data = True
+
+[options.package_data]
+niworkflows =
+    VERSION
```
oesteban: Is this still necessary?

niworkflows/tests/conftest.py
```diff
@@ -27,11 +26,6 @@ def _run_interface_mock(objekt, runtime):
     return runtime
 
 
-def pytest_runtest_setup(item):
```
oesteban: Are we not chdir'ing before every test now?

.circleci/config.yml
oesteban: I honestly have no clue

niworkflows/tests/conftest.py
effigies: Do we need to? I can restore it, since the problem was actually pytest-cov, but I don't see the purpose.

oesteban: Sure, it was there just for prophylaxis of testing, but there's no particular reason to have it. If removing it makes coverage collecting easier there's no point in restoring it :+1:

effigies: It doesn't actually seem to have any effect. I'll restore it to declutter this PR and add a note to reconsider later.

PR #376, opened 2019-07-12
FIX: Require scikit-learn because nilearn does not

effigies:
It looks like nilearn doesn't correctly specify requirements, so scikit-learn is not installed when nilearn is. However, importing nilearn fails if scikit-learn is missing, so we'll do it ourselves.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -205,7 +205,6 @@ jobs:
       - run:
           name: Build and check
           command: |
-            python3 setup.py check -r -s
             python3 setup.py sdist
             python3 -m twine check dist/*
       - run:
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -205,7 +205,6 @@ jobs:
       - run:
           name: Build and check
           command: |
-            python3 setup.py check -r -s
```
effigies: This was in response to the CI output:

```
running check
warning: Check: This command has been deprecated. Use `twine check` instead: https://packaging.python.org/guides/making-a-pypi-friendly-readme#validating-restructuredtext-markup

The project's long description is valid RST.
```

PR #378, opened 2019-07-18
FIX: Load file with mmap=False when modifying on-disk dtype

effigies:
This causes a BusError reported in [NeuroStars #4641](https://neurostars.org/t/bus-error-fmriprep/4641).

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,5 +1,6 @@
 # data
 include niworkflows/data/*.json
+include niworkflows/data/*.txt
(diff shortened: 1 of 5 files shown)


effigies:
Looking into it more, this should resolve the issue with poldracklab/fmriprep#1555 after all. I had forgotten we had fixed the other one of these a bit ago.

effigies:
@oesteban This is all set and the case is [covered](https://codecov.io/gh/poldracklab/niworkflows/pull/378/diff).

PR #379, opened 2019-07-20
[FIX][ENH] Confounds infrastructure update

rciric:
This should have been two separate PRs. Sorry about that.

* Includes necessary changes for poldracklab/fmriprep#1617
* Includes necessary changes for poldracklab/fmriprep#1618
* Includes a proposed solution for poldracklab/fmriprep#1680


effigies:
Yeah, let's split the viz stuff from the metadata stuff. I'm uncomfortable with the metadata stuff and will need to look more into `_tsv2json` to figure out whether this is the best way or if there's a better.

niworkflows/viz/plots.py
```diff
@@ -644,9 +644,13 @@ def compcor_variance_plot(metadata_files, metadata_sources=None,
...
                        fontsize='x-small', bbox=bbox_txt)
-            ax[m].text(varexp[thr][0], 25,
+            try:
+                n_components = varexp[thr][0]
+            except IndexError:
+                n_components = varexp[thr]
+            ax[m].text(n_components, 25,
                        '{} components explain\n{:.0f}% of variance'.format(
-                            varexp[thr][0], 100 * thr),
+                            n_components, 100 * thr),
```
effigies: I think the cleaner fix would actually be to change L640-L641:

```Diff
             varexp[thr] = np.searchsorted(
-                components['cumulative_variance_explained'], thr) + 1
+                np.atleast1d(components['cumulative_variance_explained']), thr) + 1
```

Then `varexp[thr]` is always an array, for any of the uses later.

niworkflows/interfaces/utils.py
```diff
@@ -896,6 +894,9 @@ def less_breakable(a_string):
     json_data = tsv_data.to_json(orient='index')
     json_data = json.JSONDecoder(
         object_pairs_hook=OrderedDict).decode(json_data)
+    for i in json_data:
+        for k, v in additional_metadata.items():
```
effigies:
```suggestion
        json_data[i].update(additional_metadata)
```

niworkflows/interfaces/utils.py
```diff
@@ -896,6 +894,9 @@ def less_breakable(a_string):
...
+            json_data[i][k] = v
```
effigies:
```suggestion
```

rciric:
Superseded by #382 and #383

PR #381, opened 2019-07-26
FIX: Revision of previous PR #337 / MELODIC did not converge

oesteban:
This PR tries to resolve a bug introduced in #337. When MELODIC did not
converge, `self._out_report` was rewritten from the input although it
had been manipulated before.

For reliability, the MELODICRPT interface is not a mixin anymore.

Screen Shot 2019-07-25 at 11 12 27 PM

### DIFF
--- a/niworkflows/interfaces/segmentation.py
+++ b/niworkflows/interfaces/segmentation.py
@@ -9,7 +9,7 @@ ReportCapableInterfaces for segmentation tools
 from __future__ import absolute_import, division, print_function, unicode_literals
 import os
 
-from nipype.interfaces.base import File
+from nipype.interfaces.base import File, BaseInterface, isdefined
 from nipype.interfaces import fsl, freesurfer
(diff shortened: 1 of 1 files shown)


oesteban:
Tested on a separate dataset I knew was failing. This fix is fully operational.

niworkflows/interfaces/segmentation.py
```diff
@@ -106,37 +106,64 @@ class MELODICOutputSpecRPT(reporting.ReportCapableOutputSpec,
     pass
 
 
-class MELODICRPT(reporting.ReportCapableInterface, fsl.MELODIC):
+class MELODICRPT(fsl.MELODIC):
     input_spec = MELODICInputSpecRPT
     output_spec = MELODICOutputSpecRPT
+    _out_report = None
+    generate_report = False
```
effigies: Since this always set in `__init__`, no need to place it in the class attributes.

```suggestion
```

PR #383, opened 2019-08-02
Handle singleton decompositions: fix poldracklab/fmriprep#1680

rciric:
* Resolves poldracklab/fmriprep#1680 and adds a unit test to cover the singleton case


niworkflows/viz/plots.py
```diff
@@ -587,7 +587,7 @@ def compcor_variance_plot(metadata_files, metadata_sources=None,
             metadata_sources = ['Decomposition {:d}'.format(i)
                                 for i in range(len(metadata_files))]
     for file, source in zip(metadata_files, metadata_sources):
-        metadata[source] = pd.read_table(str(file))
+        metadata[source] = pd.read_csv(str(file), sep='\t')
```
oesteban: Perhaps more general?
```suggestion
        metadata[source] = pd.read_csv(str(file), sep=r'\s+')
```

PR #385, opened 2019-08-19
Update .zenodo.json 

utooley:
Wasn't sure if it was appropriate for me to `[skip ci] ` for a non-code edit.

### DIFF
--- a/.zenodo.json
+++ b/.zenodo.json
@@ -53,6 +53,12 @@
       "type": "Researcher"
     },
     {
+      "name": "Tooley, Ursula A.",
+      "affiliation": "Department of Neuroscience, University of Pennsylvania, PA, USA",
+      "orcid": "0000-0001-6377-3885",
+      "type": "Researcher"
+    },
+    {
       "affiliation": "McGovern Institute for Brain Research, MIT, MA, USA",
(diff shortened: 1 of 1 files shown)


effigies:
LGTM. Yeah, there's no good way to validate `.zenodo.json` files (I've looked into it... see https://github.com/bids-standard/pybids/pull/470#issuecomment-520179305), so these are good candidates for `[skip ci]`.

@oesteban I think you set up the Zenodo file for niworkflows. Do we have any preferred ordering? I don't see a `prep_zenodo.py` or anything that will reorder by some criterion.

utooley:
The fMRIPrep ordering says new contributors should go right before Russ (which ends up being before both Chris and then Russ), so I put myself there for fMRIPrep. Should the same apply here? I didn't initially realize that they just showed up in the order they were added in.

oesteban:
We don't have any particular order.

effigies:
With fMRIPrep, we don't really distinguish between authors and contributors, so we treat Russ and Chris as co-last-authors, and then do everybody else by chronological order, just because that's easy.

I don't remember why we split authors and contributors, here. I also suspect it matters even less, because nobody knows what niworkflows is...

Merging.

Issue #374, opened 2019-07-10
Less fragile tests


effigies:
Niworkflows tests are currently sensitive to whether you've set them up in the Docker container that is built on CircleCI, making local iteration very difficult and thus slow.

The following should work:

```Shell
virtualenv --python=python /tmp/venv
source /tmp/venv/bin/activate
pip install ".[test]"
python -m pytest niworkflows
```

If non-Python dependencies are missing, they should be marked with an expected fail dependent on that.

If a template does not exist in TemplateFlow, it should either be downloaded on-the-fly or produce an expected failure.

Some instructions on how to install templates to ensure maximal test coverage would be nice.

PR #387, opened 2019-08-23
TEST: Skip tests with non-Python dependencies when missing

effigies:
Starting on #374.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,6 +19,7 @@ env:
     - INSTALL_DEPENDS="pip setuptools"
   matrix:
     - CHECK_TYPE="style"
+    - CHECK_TYPE="test"
     - INSTALL_TYPE="install"
     - INSTALL_TYPE="develop"
     - INSTALL_TYPE="sdist"
@@ -40,7 +41,6 @@ matrix:
 
 before_install:
   - python -m pip install --upgrade $INSTALL_DEPENDS
-  - pip install "flake8<3.0" flake8-putty
 
 install:
   - |
@@ -60,6 +60,8 @@ install:
   - |
     if [ "$CHECK_TYPE" == "style" ]; then
         travis_retry pip install "flake8<3.0" flake8-putty
+    elif [ "$CHECK_TYPE" == "test" ]; then
+        travis_retry pip install "niworkflows[test]"
     fi
 
 script:
@@ -75,6 +77,8 @@ script:
       echo "Intended: $INTENDED_VERSION"
       echo "Installed: $INSTALLED_VERSION"
       test "$INTENDED_VERSION" == "$INSTALLED_VERSION"
+    elif [ "$CHECK_TYPE" == "test" ]; then
+      pytest -n 2 -v --cov niworkflows --cov-config .coveragerc --cov-report xml:cov.xml niworkflows
     else
       false
     fi
--- a/niworkflows/interfaces/tests/test_itk.py
+++ b/niworkflows/interfaces/tests/test_itk.py
@@ -1,7 +1,9 @@
 import pytest
 from ..itk import _applytfms
+from nipype.interfaces.ants.base import Info
 
 
+@pytest.mark.skipif(Info.version() is None, "Missing ANTs")
 @pytest.mark.parametrize('ext', ('.nii', '.nii.gz'))
 @pytest.mark.parametrize('copy_dtype', (True, False))
 @pytest.mark.parametrize('in_dtype', ('i2', 'f4'))


effigies:
Looks like we have some code that doesn't like to work with Python 3.5. @oesteban Would you rather make the code work with 3.5 or just drop 3.5?

I'm a little inclined to drop, but there are [some dependent packages](https://github.com/poldracklab/niworkflows/network/dependents).

.travis.yml
```diff
@@ -39,42 +40,60 @@ matrix:
     env: INSTALL_DEPENDS="pip==10.0.1 setuptools==30.3.0"
 
 before_install:
+  - python -m pip install --upgrade pip virtualenv
```
effigies: The preinstalled `pytest` clashed with `pytest-xdist` but didn't get upgraded by `pytest-xdist`'s dependencies for some reason. Virtualenv starts clean.

.travis.yml
```diff
@@ -39,42 +40,60 @@ matrix:
...
+  - |
+    if [ "$INSTALL_TYPE" == "pip" ]; then
+        PACKAGE="."
+    elif [ "$INSTALL_TYPE" == "sdist" ]; then
+        python setup.py sdist
+        PACKAGE="$( ls dist/*.tar.gz )"
+    elif [ "$INSTALL_TYPE" == "wheel" ]; then
+        python setup.py bdist_wheel
+        PACKAGE="$( ls dist/*.whl )"
+    fi
```
effigies: Split out packaging from installation just to make the `install:` section technically accurate. Feel like it's a little cleaner, conceptually.

.travis.yml
```diff
@@ -39,42 +40,60 @@ matrix:
...
-    elif [ "$INSTALL_TYPE" == "pip" ]; then
-        pip install .
-    elif [ "$INSTALL_TYPE" == "sdist" ]; then
-        python setup.py sdist
-        pip install dist/*.tar.gz
-    elif [ "$INSTALL_TYPE" == "wheel" ]; then
-        python setup.py bdist_wheel
-        pip install dist/*.whl
+    else
+        pip install $PACKAGE
```
effigies: See? Cleaner.

.travis.yml
```diff
@@ -39,42 +40,60 @@ matrix:
...
-        travis_retry pip install "flake8<3.0" flake8-putty
+    INTENDED_VERSION="$(python -c 'import versioneer; print(versioneer.get_version())')"
+    pushd /tmp
+    INSTALLED_VERSION="$(python -c 'import niworkflows; print(niworkflows.__version__)')"
+    python -c 'import niworkflows; print(niworkflows.__file__)'
+    echo "Intended: $INTENDED_VERSION"
+    echo "Installed: $INSTALLED_VERSION"
+    test "$INTENDED_VERSION" == "$INSTALLED_VERSION"
+    popd
+  - if [ "$CHECK_TYPE" = "install" ]; then exit 0; fi
```
effigies: Run the version check on all cases, and then short circuit if we're just checking the installation.

.travis.yml
```diff
@@ -39,42 +40,60 @@ matrix:
...
+
+before_script:
+  - travis_retry pip install "niworkflows[$CHECK_TYPE]"
```
effigies: All the above makes this nice and clean, too.

niworkflows/interfaces/utils.py
```diff
@@ -589,10 +589,8 @@ class AddTSVHeader(SimpleInterface):
 
     .. testsetup::
 
-    >>> import pandas as pd
-    >>> from tempfile import TemporaryDirectory
-    >>> tmpdir = TemporaryDirectory()
-    >>> os.chdir(tmpdir.name)
+    >>> cwd = os.getcwd()
+    >>> os.chdir(tmpdir)
```
effigies: If you don't mind `Path(tmpdir / 'data.tsv')` below, we could even drop the `chdir` altogether.

niworkflows/interfaces/utils.py
```diff
@@ -616,7 +614,7 @@ class AddTSVHeader(SimpleInterface):
 
     .. testcleanup::
 
-    >>> tmpdir.cleanup()
+    >>> os.chdir(cwd)
```
effigies: We clean up in the fixture, so just getting out. The `chdir`s into directories that get deleted are ultimately what's causing the `pytest-cov` weirdness.

niworkflows/reports/core.py
```diff
@@ -16,6 +16,12 @@
 import jinja2
 from nipype.utils.filemanip import copyfile
 
+# Add a new figures spec
+try:
+    add_config_paths(figures=pkgrf('niworkflows', 'reports/figures.json'))
+except ValueError as e:
+    if "Configuration 'figures' already exists" != str(e):
+        raise
```
effigies: This is pretty cheap to add in general, and putting it inside an `__init__` made this file a bit brittle with `xdist`.

niworkflows/tests/conftest.py
```diff
@@ -27,12 +31,6 @@ def _run_interface_mock(objekt, runtime):
     return runtime
 
 
-# XXX This may not be necessary. Look into more when overhauling tests.
-def pytest_runtest_setup(item):
-    """Change to temporary directory"""
-    os.chdir(mkdtemp())
```
effigies: Removed this as a potential cause of the `pytest-cov` issue. Individually `chdir`ing tests that need it feels cleaner.

This has led to my repo getting polluted, so I'll track down what's causing that and `chdir` or adjust paths as needed.

Issue #388, opened 2019-08-28
Improvements when using N4BiasFieldCorrection


oesteban:
As a result of a thorough bug report from @Shotgunosine, in ANTsX/ANTs#822 several recommendations we should adopt for fMRIPrep are given:

  - Rescale the input if negative values are found (https://github.com/ANTsX/ANTs/issues/822#issuecomment-525351111)
  - Sample more voxels (i.e., shrink factor of 2 or 1) - https://github.com/ANTsX/ANTs/issues/822#issuecomment-525364416

effigies:
From my understanding, we don't need to rescale, we just need to threshold the outputs of registration to >= 0.

Shotgunosine:
Thresholding actually didn't fully address the issues. Rescaling did work
on that one test case though. I'm debugging another issue that may be
related to this one and will try to get more information shortly.

oesteban:
Thresholding changes the distribution and the step it creates is exaggerated by the initial log transform of N4. It makes sense that rescaling works better.

effigies:
If the issue is interpolation-induced negative values, then it's not clear that preserving that particular aspect of the distribution is worthwhile, as it's already an artifact. Rescaling while preserving these artifacts could shift the background noise up significantly.

oesteban:
It is undesired and it doesn't make sense as MRI signal, but not an artifact. BSpline interpolation can do this to ensure continuity of the distribution, which is necessary for N4 to find the bias field, which will become an additive offset after log transform.

Finding truncated values breaks the continuity constraints of the BSplines modeling the bias field.

Shotgunosine:
As I'm trying to track my other masking bug down with this data, I've found that the output of split_opt_comb appears to already be partially brain masked. Is that correct behavior? If it is supposed to be masked, is that discontinuity going to disrupt the N4 normalization that happens when masking the bold after transformation to standard space?

effigies:
Okay. So what if we transform the uniformized image, rather than uniformize the transformed image?

effigies:
But @Shotgunosine, I think the issue here was specifically with blip-up/blip-down, which are registered to each other and then N4-corrected? Or was it the EPI masking in template space?

Shotgunosine:
[...]

PR #389, opened 2019-08-30
FIX: Nonpositive values entered to N4 when calculating BOLDrefs

oesteban:
Addresses #388.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,7 +24,7 @@ env:
     - INSTALL_TYPE="develop"
     - INSTALL_TYPE="sdist"
     - INSTALL_TYPE="wheel"
-    - INSTALL_DEPENDS="pip==18.1 setuptools==30.2.1"
+    - INSTALL_DEPENDS="pip==18.1 setuptools==30.4.0"
     - INSTALL_DEPENDS="pip==10.0.1 setuptools==30.4.0"
 
 matrix:
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,8 +41,10 @@ matrix:
 
 before_install:
   - python --version
-  - echo $PYTHONPATH
-  - python -c "import sys; print('\n'.join(sys.path))"
+  - python -m venv /tmp/venv
+  - deactivate
+  - source /tmp/venv/bin/activate
+  - echo $PATH
   - python -m pip install --upgrade $INSTALL_DEPENDS
   - python -m pip --version
   - |
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,7 +24,7 @@ env:
     - INSTALL_TYPE="develop"
     - INSTALL_TYPE="sdist"
     - INSTALL_TYPE="wheel"
-    - INSTALL_DEPENDS="pip==18.1 setuptools==30.4.0"
+    - INSTALL_DEPENDS="pip==19.2.3 setuptools==41.2.0"
     - INSTALL_DEPENDS="pip==10.0.1 setuptools==30.4.0"
 
 matrix:
@@ -40,12 +40,11 @@ matrix:
     env: INSTALL_DEPENDS="pip==10.0.1 setuptools==30.4.0"
 
 before_install:
-  - python -m pip install --upgrade pip virtualenv
-  - virtualenv --python=python /tmp/venv
-  - source /tmp/venv/bin/activate
   - python --version
-  - pip --version
+  - echo $PYTHONPATH
+  - python -c "import sys; print('\n'.join(sys.path))"
   - python -m pip install --upgrade $INSTALL_DEPENDS
+  - python -m pip --version
   - |
     if [ "$INSTALL_TYPE" == "pip" ]; then
         PACKAGE="."
@@ -64,7 +63,7 @@ install:
     elif [ "$INSTALL_TYPE" == "develop" ]; then
         python setup.py develop
     else
-        pip install $PACKAGE
+        python -m pip install $PACKAGE
     fi
   - |
     INTENDED_VERSION="$(python -c 'import versioneer; print(versioneer.get_version())')"
@@ -78,7 +77,7 @@ install:
   - if [ "$CHECK_TYPE" = "install" ]; then exit 0; fi
 
 before_script:
-  - travis_retry pip install "niworkflows[$CHECK_TYPE]"
+  - travis_retry python -m pip install "niworkflows[$CHECK_TYPE]"
   - |
     if [ "$CHECK_TYPE" == "test" ]; then
         mkdir -p $HOME/.cache/stanford-crn
@@ -99,5 +98,5 @@ script:
     fi
 
 after_script:
-  - pip install codecov
+  - python -m pip install codecov
   - codecov
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,7 +24,7 @@ env:
     - INSTALL_TYPE="develop"
     - INSTALL_TYPE="sdist"
     - INSTALL_TYPE="wheel"
-    - INSTALL_DEPENDS="pip==19.2.3 setuptools==41.2.0"
+    - INSTALL_DEPENDS="pip==18.1 setuptools==30.2.1"
     - INSTALL_DEPENDS="pip==10.0.1 setuptools==30.4.0"
 
 matrix:
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,11 +40,11 @@ matrix:
     env: INSTALL_DEPENDS="pip==10.0.1 setuptools==30.4.0"
 
 before_install:
-  - python --version
-  - python -m venv /tmp/venv
-  - deactivate
+  - python -m pip install --upgrade pip virtualenv
+  - virtualenv --python=python /tmp/venv
   - source /tmp/venv/bin/activate
-  - echo $PATH
+  - python --version
+  - python -m pip --version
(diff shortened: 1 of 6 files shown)


oesteban:
Before - https://1479-63891310-gh.circle-artifacts.com/0/tmp/masks/reports/ds001362/sub-01_task-taskname_run-01_bold_masks.svg
After - https://799-74174878-gh.circle-artifacts.com/0/tmp/masks/reports/ds001362/sub-01_task-taskname_run-01_bold_masks.svg

[the other tests remained the same, I had to update the original file for this one because it was truncated very badly]

Evolution of mask tests are posted here - https://circleci.com/gh/oesteban/workflows/niworkflows/tree/masks%2F388

oesteban:
The installation of nipype from a repo is failing with Python 3.7. It seems like the `safe_zip` option is not being read and setuptools fails while checking the package. It looks like a Python 3.7 bug, or that nipype's installation infrastructure has been finally abandoned.

My impression is that we would have the same problem with other dependencies, if we start using the bleeding edge version of them, instead of wheel packages (i.e., this is not nipype's fault).

cc/ @effigies for a better diagnosis and @satra for opinions about what can be updated in nipype -in particular, could we migrate to a `setup.cfg` type of installation?

effigies:
There might be a slight issue with nipype configuration, but if you notice the basic `pip` install works with the latest versions of pip and setuptools, so it's almost certainly one, the other or an interaction of those on 3.7.

I can try a batch of test environments to try to establish the cause and set a new minimum setuptools.

oesteban:
> so it's almost certainly one, the other or an interaction of those on 3.7

Would you say it is worth holding this up until we find out or we can temporarily change these tests for the latest pip and setuptools?

oesteban:
Latest pip and setuptools [seem to work](https://travis-ci.org/poldracklab/niworkflows/jobs/580361902). Opened #392 to track the installation issues.

setup.cfg
```diff
@@ -72,8 +73,8 @@ style =
     flake8 < 3.0
     flake8-putty
 test =
-    pytest
-    pytest-xdist
+    pytest >= 4.4
```
oesteban: I had to set these two pinnings for travis-ci to work too.

effigies:
The easy thing right now is just to make the failing Travis entry an allowed failure, and move resolution to #392.

Note, though, that we can't send a package to PyPI with this kind of dependency. So we can send out fixed Docker images, but releases will be messy.

oesteban:
Can we issue a pre-release at least?

.travis.yml
```diff
@@ -24,7 +24,7 @@ env:
     - INSTALL_TYPE="develop"
     - INSTALL_TYPE="sdist"
     - INSTALL_TYPE="wheel"
-    - INSTALL_DEPENDS="pip==18.1 setuptools==30.2.1"
+    - INSTALL_DEPENDS="pip==19.2.3 setuptools==41.2.0"
```
effigies: The latest will be installed and tested in `CHECK_TYPE="test"`. This entry is here specifically to test `pip`'s ability to upgrade `setuptools` using `pyproject.toml` when a too-old version is present. So it's almost certainly a `pip` problem, not a `setuptools` problem.

.travis.yml
```diff
@@ -40,12 +40,11 @@ matrix:
     env: INSTALL_DEPENDS="pip==10.0.1 setuptools==30.4.0"
 
 before_install:
-  - python -m pip install --upgrade pip virtualenv
-  - virtualenv --python=python /tmp/venv
-  - source /tmp/venv/bin/activate
```
effigies: I would prefer `virtualenv`s in general, because they are clean and implicit dependencies are harder to creep in. If we want to have a test entry that doesn't use them, to make sure that we upgrade existing test setups, that's fine, but I would make it conditional.

oesteban: Please note that you are already within a virtualenv in travis. We could use venv though.

.travis.yml
oesteban: I believe you are right. I think this is a pipenv problem

.travis.yml
effigies: The problem is that it's a virtualenv that's precreated with various packages, such as `pytest`.

I've tried `venv`, but didn't have the patience to figure out how to use it. But I don't object to you figuring it out and letting me steal it for other projects. :-)

effigies:
> Can we issue a pre-release at least?

I don't know. I don't remember if PyPI rejects the package, or pip rejects that metadata from PyPI packages.

It might also be wheels that fail, while sdists work fine...

.travis.yml
oesteban: > but didn't have the patience to figure out how to use it

Exactly the same as virtualenv :D

oesteban:
> From both transparency and complexity perspectives, I think an interface to perform rescaling would be preferable to this approach.

Complexity concern: code would be simpler, but the graph executed by nipype would be one node ( per workflow (we compute these at least once per run, but usually two or three times). Conceptually is also more complex to detach the inputs massaging from the interface - after all this operation only makes sense to do before N4. When writing a new workflow, it would be easy to forget to write the rescale node before.

Transparency concern: I agree this is less transparent. For forensics, I'll add a boolean output to tell whether the input was rescaled. That should be enough to know the full provenance. When a user doesn't want this behavior, they can always use the original Nipype's interface to be sure the rescaling didn't happen.

Finally, just for the record, this is an offset (i.e., intercept) - there is no scaling (i.e., slope).

setup.cfg
```diff
@@ -18,14 +18,15 @@ classifiers =
     Programming Language :: Python :: 3.7
 
 [options]
+zip_safe = True
```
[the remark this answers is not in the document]
oesteban: This would make it easier for dependants the installation via wheels (https://setuptools.readthedocs.io/en/latest/setuptools.html#setting-the-zip-safe-flag).

.travis.yml
oesteban: Okay, I take it back - having a virtualenv imposed by Travis makes it updating virtualenv the only solution.

effigies: Ah, so that's why it wasn't a drop-in replacement when I tried it. I guess you could try `deactivate`, but you may also find it not worth the time to fiddle with.

oesteban:
@effigies if I successfully replied to your concerns about complexity and transparency this should be ready to merge :+1:

.travis.yml
effigies: This is still pointed to the latest.

niworkflows/anat/ants.py
```diff
@@ -287,6 +287,8 @@ def init_brain_extraction_wf(name='brain_extraction_wf',
             n_iterations=[50] * 5, convergence_threshold=1e-7, shrink_factor=4,
             bspline_fitting_distance=bspline_fitting_distance),
         n_procs=omp_nthreads, name='inu_n4_final', iterfield=['input_image'])
+    if _ants_version and parseversion(_ants_version) >= Version('2.1.0'):
+        inu_n4_final.inputs.rescale_intensities = True
```
effigies: Does this indicate that we're diverging from `antsBrainExtraction.sh`, or is this a change they also make in 2.1.0? Would we consider the result just as valid with old ANTs, and if not, should we just set this and let the `min_ver` check fail?

niworkflows/func/util.py
```diff
@@ -251,9 +252,12 @@ def init_enhance_and_skullstrip_bold_wf(
...
 
     # Run N4 normally, force num_threads=1 for stability (images are small, no need for >1)
-    n4_correct = pe.Node(ants.N4BiasFieldCorrection(
-        dimension=3, copy_header=True, bspline_fitting_distance=200),
+    n4_correct = pe.Node(N4BiasFieldCorrection(
+        dimension=3, copy_header=True, bspline_fitting_distance=200), shrink_factor=2,
         name='n4_correct', n_procs=1)
+    _n4_version = N4BiasFieldCorrection().version
+    if _n4_version and parseversion(_n4_version) >= Version('2.1.0'):
+        n4_correct.inputs.rescale_intensities = True
```
effigies: Again. Is there an advantage to accommodating ANTs versions that don't do what we need?

niworkflows/anat/ants.py
oesteban: This is applied to a refinement step not found in the original script. Looking at it with perspective, we should've taken out this final bias correction away from this workflow.

However, to keep minimal changes in fMRIPrep, we stuffed this in the workflow because it seemed better not to modify the API.

In this context, adding the rescale flag makes sense for this final INU step (please note I didn't include it in the initial INU-correction which comes from the original ANTs' script).

niworkflows/func/util.py
oesteban: I see it a bit differently here, I totally agree with you.

niworkflows/anat/ants.py
effigies: Okay, but shouldn't we set it to`True` always, and error if bad version?

oesteban: Not having the flag set would not change the output dramatically. I would recommend using it, but would not prevent people from running the workflow with older versions of ANTs. Since this is a "general purpose" workflow, I'd keep the option for the low-risk it entails.

oesteban: I'm adding a warning - would that address your concern?

effigies: Sure.

PR #391, opened 2019-09-03
FIX: Retain newlines in corrected LTA files

davhunt:


effigies:
To sketch out a test, you could probably just create a temporary directory with >=255 characters in the path, create a trivial NIfTI image, and use `lta_convert` to create an identity LTA with too-long filenames. Then you should be able to reproduce the segfault with `mri_vol2vol`, run a stub interface that uses this mixin to simply copy (and implicitly fix) the bad LTA file, and then successfully run `mri_vol2vol` with the result.

davhunt:
Um sure I can try, what code exactly am I trying to replace? Do you mean try to catch the exception here https://github.com/poldracklab/niworkflows/blob/e77e1ab0b01fc6f563d970314e1924ca83c8d81f/niworkflows/interfaces/registration.py#L235 maybe? Not sure exactly what you're saying.

oesteban:
So the (my) idea is:

1. Strip this block off of `_post_run_hook()` to a new function:

https://github.com/poldracklab/niworkflows/blob/65f70c5dfbcce6f6f5b8624895917e58f341ea66/niworkflows/interfaces/freesurfer.py#L203-L220

2. Add two LTA files to the project (one with a short filename and one with a long filename)

3. Write doctests for this new functions checking the two LTA files. The one with a short filename should remain untouched.

PR #396, opened 2019-09-13
Create separate "fix_lta_length" function to shorten filename headers > 255 characters in lta file

davhunt:


davhunt:
Hi @oestaban I'm trying to create the sample .lta's in the doctest with freesurfer's mri_coreg but I don't think $FREESURFER_HOME is set in the environment when the doctest runs. Any advice?

PR #401, opened 2019-09-24
ENH: Add ``copy_header`` inputs to some ANTs interfaces

oesteban:
This PR adds an input ``copy_header`` to the ``ImageMath`` and ``ThresholdImage`` interfaces because ANTs fails to produce images with appropriate x-form headers.

In particular, these changes are necessary for poldracklab/mriqc#803 to go forward.

In addition to the ``copy_header`` change, some PEP stylistic issues have been addressed.

### DIFF
--- a/niworkflows/interfaces/ants.py
+++ b/niworkflows/interfaces/ants.py
@@ -21,7 +21,7 @@ class _ImageMathInputSpec(ANTSCommandInputSpec):
     op2 = traits.Either(base.File(exists=True), base.Str, position=-1,
                         argstr='%s', desc='second operator')
     copy_header = traits.Bool(
-        True, mandatory=True, usedefault=True,
+        True, usedefault=True,
         desc='copy headers of the original image into the output (corrected) file')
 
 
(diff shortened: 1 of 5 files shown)


effigies:
Are we testing this anywhere downstream? `super()._list_outputs()` is not guaranteed to work for all interfaces, so that we're not getting coverage on these lines worries me. A cheap test that exercises this functionality for each interface would be nice.

effigies:
Otherwise, this looks good.

niworkflows/interfaces/ants.py
```diff
@@ -516,3 +534,16 @@ def _list_outputs(self):
...
             )
         return outputs
+
+
+def _copy_header(header_file, in_file, set_dtype=True):
+    """Copy header from input image to an output image."""
+    import nibabel as nb
+    in_img = nb.load(header_file)
+    out_img = nb.load(in_file, mmap=False)
+    new_img = out_img.__class__(out_img.get_fdata(), in_img.affine,
```
oesteban: @effigies - shouldn't this be using the `out_img.dataobj` instead?

effigies: I would. It shouldn't change the values, just skip a cast or two.

PR #403, opened 2019-09-24
TEST/RF: Separate LTA length fixing and add doctest

davhunt:
Follow-up to #391.


effigies:
Hi @davhunt, sorry about the very slow turnaround... I'm going to review this now,

niworkflows/interfaces/freesurfer.py
```diff
@@ -326,6 +314,54 @@ def _run_interface(self, runtime):
...
+    0
+    >>> subprocess.call(["lta_convert", "--inlta", "identity.nofile", "--src", short_name,
+    ... "--trg", short_name, "--outlta", "lta_file_2.lta"], stdout=devnull)
+    0
+    >>> fix_lta_length('lta_file_1.lta')
+    True
+    >>> fix_lta_length('lta_file_2.lta')
+    False
+    >>> shutil.rmtree('/Q')
+    >>> os.remove(short_path); os.remove('lta_file_1.lta'); os.remove('lta_file_2.lta')
```
effigies: The code looks fine, but all of this setup should not go in a doctest.

Just add the two LTA files in https://github.com/poldracklab/niworkflows/tree/master/niworkflows/tests/data, perhaps with names like `valid_transform.lta` and `long_path_transform.lta`.

Then you'll need to make the files available from a doctest here:

https://github.com/poldracklab/niworkflows/blob/c4ceb2ea570a30a406ea50c0d97a7374e596f6b8/niworkflows/conftest.py#L17-L33

I would use:

```Python
doctest_namespace['test_data'] = pkg_resources.resource_filename('niworkflows', 'tests/data')
```

Then your doctest can become something like:

```
""" Fix the length of the filename field in an LTA file if too long.

Updates file in place.

Examples
--------

No changes are made to a valid transform file:

>>> valid_transform = Path(test_data) / 'valid_transform.lta'
>>> orig_contents = valid_transform.read_text()
>>> fix_lta_length(valid_transform)
False
>>> valid_transform.read_text() == orig_contents
True

Invalid transform files are files with > 255 characters in any line:

>>> invalid_transform = Path(test_data) / 'long_path_transform.lta'
>>> orig_contents = invalid_transform.readtext()
>>> any(len(line) > 255 for line in orig_contents.splitlines(keepends=True))
True
>>> local_copy = Path('invalid.lta')
>>> local_copy.write_text(orig_contents)
>>> fix_lta_length(local_copy)
True
>>> local_copy.read_text() == orig_contents
False
"""
```

niworkflows/interfaces/freesurfer.py
```diff
@@ -205,19 +205,7 @@ def _post_run_hook(self, runtime):
...
-            for line in lines:
-                if line.startswith('filename = ') and len(line.strip("\n")) >= 255:
-                    fixed = True
-                    newfile.append('filename = path_too_long\n')
-                else:
-                    newfile.append(line)
-
-            if fixed:
-                Path(lta_file).write_text(''.join(newfile))
+        fix_lta_length(lta_file)
```
effigies: This should be inside the for loop.

```suggestion
            fix_lta_length(lta_file)
```

davhunt:
@effigies Should be good now!

effigies:
Yup! LGTM. Thanks for all your effort!

oesteban:
@davhunt could not keep up with the progress - thanks very much!

PR #407, opened 2019-09-30
ENH: Enable optional BIDS entity filtering to data collection utility

bpinsard:
related to https://github.com/poldracklab/fmriprep/pull/1770


niworkflows/utils/bids.py
```diff
@@ -161,6 +161,9 @@ def collect_data(bids_dir, participant_label, task=None, echo=None,
         't1w': {'datatype': 'anat', 'suffix': 'T1w'},
         'roi': {'datatype': 'anat', 'suffix': 'roi'},
     }
+    if bids_filters:
```
oesteban:
```suggestion
    bids_filters = bids_filters or {}
```

niworkflows/utils/bids.py
```diff
@@ -161,6 +161,9 @@ def collect_data(bids_dir, participant_label, task=None, echo=None,
...
+        for acq, entities in bids_filters.items():
```
oesteban:
```suggestion
    for acq, entities in bids_filters.items():
```

niworkflows/utils/bids.py
```diff
@@ -161,6 +161,9 @@ def collect_data(bids_dir, participant_label, task=None, echo=None,
...
+            queries[acq].update(entities)
```
oesteban:
```suggestion
        queries[acq].update(entities)
```

bpinsard:
It is already a dict of dict. It copies the hard-coded structure within collect data. The filter as to be per suffix (_bold, _T1w...) it is not applied to all.

bpinsard:
The following doctest should work.
`t1w` in collect_data is lowercase contrary to BIDS suffix.

```
>>> bids_root, _ = collect_data(str(datadir / 'ds051'), '01',
    ...                             bids_validate=False, bids_filters={'t1w':{'run': 1}})
    >>> bids_root['t1w']  # doctest: +ELLIPSIS
```

effigies:
Let's just do this and see if we need to update it.

PR #408, opened 2019-10-01
ENH: Allow bold_reference_wf to accept multiple EPIs/SBRefs

### LINKED ISSUE #399: Accept multiple EPIs/SBRefs for reference image calculation
`init_bold_reference_wf` currently accepts a single SBRef image and a single EPI run. With multi-echo EPI, however, there will generally be one SBRef and one EPI run for each echo. The current idea for fMRIPrep is to use just the first echo's SBRef/EPI as the reference image, but I think that it's likely that at some point some method will be developed that leverages the information across echoes. Therefore, it might be good to make sure that `init_bold_reference_wf` can accept a list of EPI/SBRef images, even if it just selects the first image internally for now.

This references https://github.com/poldracklab/fmriprep/issues/1745#issuecomment-533629602.

tsalo:
Closes #399. 

This PR allows `bold_reference_wf` to accept multiple BOLD and/or SBRef files, under the assumption that multiple files reflect multiple echoes from the same run. It currently just selects the first image from each for use, but could be extended to use information from multiple echoes for reference image generation.

The workflow still returns just one BOLD file, since a number of workflows take the BOLD file stored in the output node as their inputs, and they are not currently set up to work with multi-echo data. To that end I also added a function to `utils.misc` to select the first item in a list.

These changes shouldn't impact behavior at all, but there will need to be a corresponding PR in fMRIPrep in order to feed multi-echo data and SBRefs into the reference image workflow, when available.

### DIFF
--- a/niworkflows/func/util.py
+++ b/niworkflows/func/util.py
@@ -6,7 +6,6 @@ from pkg_resources import resource_filename as pkgr_fn
 
 from nipype.pipeline import engine as pe
 from nipype.interfaces import utility as niu, fsl, afni
-from nipype.utils.filemanip import ensure_list
 
 from templateflow.api import get as get_template
 
@@ -21,6 +20,7 @@ from ..interfaces.images import ValidateImage, MatchHeader
 from ..interfaces.masks import SimpleShowMaskRPT
 from ..interfaces.registration import EstimateReferenceImage
 from ..interfaces.utils import CopyXForm
+from ..utils.connections import listify
 from ..utils.misc import pass_dummy_scans as _pass_dummy_scans
 
 
@@ -176,7 +176,7 @@ methodology of *fMRIPrep*.
 
     # fmt: off
     workflow.connect([
-        (inputnode, val_bold, [(("bold_file", ensure_list), "in_file")]),
+        (inputnode, val_bold, [(("bold_file", listify), "in_file")]),
         (inputnode, enhance_and_skullstrip_bold_wf, [
             ("bold_mask", "inputnode.pre_mask"),
         ]),
@@ -185,7 +185,7 @@ methodology of *fMRIPrep*.
         (gen_ref, enhance_and_skullstrip_bold_wf, [
             ("ref_image", "inputnode.in_file"),
         ]),
-        (val_bold, bold_1st, [(("out_file", ensure_list), "inlist")]),
+        (val_bold, bold_1st, [(("out_file", listify), "inlist")]),
         (gen_ref, calc_dummy_scans, [("n_volumes_to_discard", "algo_dummy_scans")]),
         (calc_dummy_scans, outputnode, [("skip_vols_num", "skip_vols")]),
         (gen_ref, outputnode, [
@@ -197,7 +197,7 @@ methodology of *fMRIPrep*.
             ("outputnode.mask_file", "bold_mask"),
             ("outputnode.skull_stripped_file", "ref_image_brain"),
         ]),
-        (val_bold, validate_1st, [(("out_report", ensure_list), "inlist")]),
+        (val_bold, validate_1st, [(("out_report", listify), "inlist")]),
         (bold_1st, outputnode, [("out", "bold_file")]),
         (validate_1st, outputnode, [("out", "validation_report")]),
     ])
@@ -218,7 +218,7 @@ methodology of *fMRIPrep*.
         )
         # fmt: off
         workflow.connect([
-            (inputnode, val_sbref, [(("sbref_file", ensure_list), "in_file")]),
+            (inputnode, val_sbref, [(("sbref_file", listify), "in_file")]),
             (val_sbref, gen_ref, [("out_file", "sbref_file")]),
         ])
         # fmt: on
--- a/niworkflows/func/util.py
+++ b/niworkflows/func/util.py
@@ -6,6 +6,7 @@ from pkg_resources import resource_filename as pkgr_fn
 
 from nipype.pipeline import engine as pe
 from nipype.interfaces import utility as niu, fsl, afni
+from nipype.utils.filemanip import ensure_list
 
 from templateflow.api import get as get_template
 
@@ -163,8 +164,8 @@ using a custom methodology of *fMRIPrep*.
         (inputnode, enhance_and_skullstrip_bold_wf, [
             ("bold_mask", "inputnode.pre_mask"),
         ]),
-        (inputnode, validate, [("bold_file", "in_file")]),
-        (inputnode, gen_ref, [("sbref_file", "sbref_file")]),
+        (inputnode, validate, [(("bold_file", ensure_list), "in_file")]),
+        (inputnode, gen_ref, [(("sbref_file", ensure_list), "sbref_file")]),
         (inputnode, calc_dummy_scans, [("dummy_scans", "dummy_scans")]),
         (validate, gen_ref, [("out_file", "in_file")]),
         (gen_ref, enhance_and_skullstrip_bold_wf, [
--- a/niworkflows/func/util.py
+++ b/niworkflows/func/util.py
@@ -32,6 +32,7 @@ def init_bold_reference_wf(
     bold_file=None,
     brainmask_thresh=0.85,
     pre_mask=False,
+    multiecho=False,
     name="bold_reference_wf",
     gen_report=False,
 ):
@@ -52,19 +53,22 @@ def init_bold_reference_wf(
 
     Parameters
     ----------
-    omp_nthreads : int
+    omp_nthreads : :obj:`int`
         Maximum number of threads an individual process may use
-    bold_file : str
+    bold_file : :obj:`str`
         BOLD series NIfTI file
     brainmask_thresh: :obj:`float`
         Lower threshold for the probabilistic brainmask to obtain
         the final binary mask (default: 0.85).
-    pre_mask : bool
+    pre_mask : :obj:`bool`
         Indicates whether the ``pre_mask`` input will be set (and thus, step 1
         should be skipped).
-    name : str
+    multiecho : :obj:`bool`
+        If multiecho data was supplied, data from the first echo
+        will be selected
+    name : :obj:`str`
         Name of workflow (default: ``bold_reference_wf``)
-    gen_report : bool
+    gen_report : :obj:`bool`
         Whether a mask report node should be appended in the end
 
     Inputs
@@ -105,7 +109,13 @@ def init_bold_reference_wf(
 
     """
     workflow = Workflow(name=name)
-    workflow.__desc__ = """\
+    if multiecho:
+        workflow.__desc__ = """\
+First, a reference volume and its skull-stripped version were generated
+from the first echo using a custom methodology of *fMRIPrep*.
+"""
+    else:
+        workflow.__desc__ = """\
 First, a reference volume and its skull-stripped version were generated
 using a custom methodology of *fMRIPrep*.
 """
@@ -136,12 +146,17 @@ using a custom methodology of *fMRIPrep*.
     if bold_file is not None:
         inputnode.inputs.bold_file = bold_file
 
-    validate = pe.MapNode(
-        ValidateImage(),
-        name="validate",
-        mem_gb=DEFAULT_MEMORY_MIN_GB,
-        iterfield=["in_file"],
-    )
+    if multiecho:
+        validate = pe.MapNode(
+            ValidateImage(),
+            name="validate",
+            mem_gb=DEFAULT_MEMORY_MIN_GB,
+            iterfield=["in_file"],
+        )
+    else:
+        validate = pe.Node(
+            ValidateImage(), name="validate", mem_gb=DEFAULT_MEMORY_MIN_GB
+        )
 
     gen_ref = pe.Node(
         EstimateReferenceImage(), name="gen_ref", mem_gb=1
@@ -164,16 +179,11 @@ using a custom methodology of *fMRIPrep*.
         (inputnode, enhance_and_skullstrip_bold_wf, [
             ("bold_mask", "inputnode.pre_mask"),
         ]),
-        (inputnode, validate, [(("bold_file", ensure_list), "in_file")]),
         (inputnode, gen_ref, [("sbref_file", "sbref_file")]),
         (inputnode, calc_dummy_scans, [("dummy_scans", "dummy_scans")]),
         (validate, gen_ref, [("out_file", "in_file")]),
         (gen_ref, enhance_and_skullstrip_bold_wf, [
             ("ref_image", "inputnode.in_file"),
-        ]),
-        (validate, outputnode, [
-            (("out_file", select_first), "bold_file"),
-            ("out_report", "validation_report"),
         ]),
         (gen_ref, calc_dummy_scans, [("n_volumes_to_discard", "algo_dummy_scans")]),
         (calc_dummy_scans, outputnode, [("skip_vols_num", "skip_vols")]),
@@ -187,6 +197,19 @@ using a custom methodology of *fMRIPrep*.
             ("outputnode.skull_stripped_file", "ref_image_brain"),
         ]),
     ])
+
+    if multiecho:
+        workflow.connect([
+            (inputnode, validate, [(("bold_file", ensure_list), "in_file")]),
+            (validate, outputnode, [(("out_file", select_first), "bold_file"),
+                                    ("out_report", "validation_report")]),
+        ])
+    else:
+        workflow.connect([
+            (inputnode, validate, [("bold_file", "in_file")]),
+            (validate, outputnode, [("out_file", "bold_file"),
+                                    ("out_report", "validation_report")]),
+        ])
     # fmt: on
 
     if gen_report:
--- a/niworkflows/func/util.py
+++ b/niworkflows/func/util.py
@@ -165,7 +165,7 @@ using a custom methodology of *fMRIPrep*.
             ("bold_mask", "inputnode.pre_mask"),
         ]),
         (inputnode, validate, [(("bold_file", ensure_list), "in_file")]),
-        (inputnode, gen_ref, [(("sbref_file", ensure_list), "sbref_file")]),
+        (inputnode, gen_ref, [("sbref_file", "sbref_file")]),
         (inputnode, calc_dummy_scans, [("dummy_scans", "dummy_scans")]),
         (validate, gen_ref, [("out_file", "in_file")]),
         (gen_ref, enhance_and_skullstrip_bold_wf, [
--- a/niworkflows/func/util.py
+++ b/niworkflows/func/util.py
@@ -21,7 +21,7 @@ from ..interfaces.images import ValidateImage, MatchHeader
 from ..interfaces.masks import SimpleShowMaskRPT
 from ..interfaces.registration import EstimateReferenceImage
 from ..interfaces.utils import CopyXForm
-from ..utils.misc import select_first, pass_dummy_scans as _pass_dummy_scans
+from ..utils.misc import pass_dummy_scans as _pass_dummy_scans
 
 
 DEFAULT_MEMORY_MIN_GB = 0.01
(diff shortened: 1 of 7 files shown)


niworkflows/interfaces/registration.py
```diff
@@ -347,8 +347,19 @@ def _post_run_hook(self, runtime):
 
 
 class EstimateReferenceImageInputSpec(BaseInterfaceInputSpec):
-    in_file = File(exists=True, mandatory=True, desc="4D EPI file")
-    sbref_file = File(exists=True, desc="Single band reference image")
+    in_file = traits.Either(File(exists=True),
+                            InputMultiPath(File(exists=True)),
```
effigies: `InputMultiPath` already incorporates the possibility we may get one or more. Let's drop the `Either`.

niworkflows/interfaces/registration.py
```diff
@@ -347,8 +347,19 @@ def _post_run_hook(self, runtime):
...
+                            mandatory=True,
+                            desc=('4D EPI file. If multiple files '
+                                  'are provided, they are assumed '
+                                  'to represent multiple echoes '
+                                  'from the same run.'))
+    sbref_file = traits.Either(File(exists=True),
+                               InputMultiPath(File(exists=True)),
```
effigies: Again.

niworkflows/interfaces/registration.py
```diff
@@ -372,17 +383,25 @@ class EstimateReferenceImage(SimpleInterface):
     output_spec = EstimateReferenceImageOutputSpec
 
     def _run_interface(self, runtime):
-        ref_name = self.inputs.in_file
+        # Select first EPI file
+        if not isinstance(self.inputs.in_file, File):
+            ref_name = self.inputs.in_file[0]
+        else:
+            ref_name = self.inputs.in_file
```
effigies: Using `InputMultiPath`, you can always treat this as a list.

```Python
ref_name = self.inputs.in_file[0]
```

niworkflows/interfaces/registration.py
```diff
@@ -372,17 +383,25 @@ class EstimateReferenceImage(SimpleInterface):
...
 
         out_ref_fname = os.path.join(runtime.cwd, "ref_bold.nii.gz")
         if isdefined(self.inputs.sbref_file):
-            out_ref_fname = os.path.join(runtime.cwd, "ref_sbref.nii.gz")
-            ref_name = self.inputs.sbref_file
+            # Select first SBRef file
+            if not isinstance(self.inputs.sbref_file, File):
+                ref_name = self.inputs.sbref_file[0]
+            else:
+                ref_name = self.inputs.sbref_file
```
effigies: Again:

```Python
ref_name = self.inputs.sbref_file[0]
```

niworkflows/utils/misc.py
```diff
@@ -184,6 +184,19 @@ def splitext(fname):
...
+    """
+    Select the first file from a list of filenames.
+    Used to grab the first echo's file when processing
+    multi-echo data through workflows that only accept
+    a single file.
+    """
+    if isinstance(in_files, list):
+        return in_files[0]
+    else:
+        return in_files
```
effigies: I'm a bit surprised we don't already have this somewhere. While we're here, could you just add a doctest showing the results of both conditions?

tsalo:
@effigies thanks! I've made the requested changes. I do plan to open a patch PR to fMRIPrep to match this one.

The tests seem to be failing due to a random connection error.

effigies:
Before you change any code in fMRIPrep, can you just update the niworkflows requirement, to see if that breaks tests? If we can maintain backwards compatibility, I think that would be ideal. If we don't currently have it, I think we can fix it up with:

```diff
     workflow.connect([
         (inputnode, enhance_and_skullstrip_bold_wf, [('bold_mask', 'inputnode.pre_mask')]),
-        (inputnode, validate, [('bold_file', 'in_file')]),
+        (inputnode, validate, [(('bold_file', ensure_list), 'in_file')]),
         (inputnode, gen_ref, [('sbref_file', 'sbref_file')]),
         (inputnode, calc_dummy_scans, [('dummy_scans', 'dummy_scans')]),
```

Using [`nipype.utils.filemanip.ensure_list`](https://github.com/nipy/nipype/blob/0fa28d5b6000ddaa513301d98322d48e6fe3c642/nipype/utils/filemanip.py#L614-L624).

niworkflows/func/util.py
```diff
@@ -131,12 +134,12 @@ def init_bold_reference_wf(omp_nthreads, bold_file=None, pre_mask=False,
 
     workflow.connect([
         (inputnode, enhance_and_skullstrip_bold_wf, [('bold_mask', 'inputnode.pre_mask')]),
-        (inputnode, validate, [('bold_file', 'in_file')]),
-        (inputnode, gen_ref, [('sbref_file', 'sbref_file')]),
+        (inputnode, validate, [(('bold_file', ensure_list), 'in_file')]),
+        (inputnode, gen_ref, [(('sbref_file', ensure_list), 'sbref_file')]),
```
effigies: Definitely don't need this one, since `sbref_file` is now an `InputMultiPath`.

```suggestion
        (inputnode, gen_ref, [('sbref_file', 'sbref_file')]),
```

Speaking of which...

niworkflows/interfaces/registration.py
```diff
@@ -347,8 +347,17 @@ def _post_run_hook(self, runtime):
 
 
 class EstimateReferenceImageInputSpec(BaseInterfaceInputSpec):
-    in_file = File(exists=True, mandatory=True, desc="4D EPI file")
-    sbref_file = File(exists=True, desc="Single band reference image")
+    in_file = InputMultiPath(File(exists=True),
```
effigies: `InputMultiPath` is an old name that is being deprecated eventually. We should move to `InputMultiObject`.

```suggestion
    in_file = InputMultiObject(File(exists=True),
```

niworkflows/interfaces/registration.py
```diff
@@ -347,8 +347,17 @@ def _post_run_hook(self, runtime):
...
+                             mandatory=True,
+                             desc=('4D EPI file. If multiple files '
+                                   'are provided, they are assumed '
+                                   'to represent multiple echoes '
+                                   'from the same run.'))
+    sbref_file = InputMultiPath(File(exists=True),
```
effigies:
```suggestion
    sbref_file = InputMultiObject(File(exists=True),
```

niworkflows/interfaces/registration.py
```diff
@@ -15,7 +15,7 @@
 from nilearn.image import index_img
 from nipype.utils.filemanip import fname_presuffix
 from nipype.interfaces.base import (
-    traits, isdefined, TraitedSpec, BaseInterfaceInputSpec, File, SimpleInterface)
+    traits, isdefined, TraitedSpec, BaseInterfaceInputSpec, File, SimpleInterface, InputMultiPath)
```
effigies:
```suggestion
    traits, isdefined, TraitedSpec, BaseInterfaceInputSpec, File, SimpleInterface, InputMultiObject)
```

niworkflows/func/util.py
tsalo: 🤦‍♂ Ohhh.... I get it now. Thanks!

effigies:
Don't worry about the pep8 speaks just yet... I need to configure it.

effigies:
Okay, should be configured. Can you merge `master` in?

effigies:
Hmm. I don't understand why it's not getting the 99-character limit correct. I'll look into it.

effigies:
Got it. It was looking for config in `setup.cfg`, not `tox.ini`. Updated. Would you mind rebasing this time, to avoid a bunch of merge commits?

tsalo:
Will do. Regarding fMRIPrep compatibility, what is the best way for me to check? I can run fMRIPrep (plus my niworkflows fork) locally in a Docker image, but is it enough to run with `--boilerplate` or should I process a dataset fully with `--sloppy`?

tsalo:
Oh sorry, I made a merge commit instead. Sorry about that. I assume that this can be squashed and merged though? To remove those commits?

effigies:
Eh, either way. But I don't think you're likely to run into any conflicts with:

```Shell
git fetch poldracklab
git rebase poldracklab/master
```

If it doesn't rebase cleanly, just `git rebase --abort` and don't worry about it.

effigies:
Cool, thanks.

For fMRIPrep, I just mean have a commit where you update the niworkflows dependency to `niworkflows @ git+https://github.com/tsalo/niworkflows.git@<commit>`, and let the tests run, before making any code changes.

effigies:
Okay. Hopefully last rebase needed...

niworkflows/func/util.py
```diff
@@ -109,14 +111,15 @@ def init_bold_reference_wf(omp_nthreads, bold_file=None, pre_mask=False,
...
-                                      'bold_mask', 'validation_report', 'mask _report']),
+                                      'bold_mask', 'validation_report', 'mask_report']),
         name='outputnode')
 
     # Simplify manually setting input image
     if bold_file is not None:
         inputnode.inputs.bold_file = bold_file
 
-    validate = pe.Node(ValidateImage(), name='validate', mem_gb=DEFAULT_MEMORY_MIN_GB)
+    validate = pe.MapNode(ValidateImage(), name='validate', mem_gb=DEFAULT_MEMORY_MIN_GB,
```
oesteban: Could we force the developer to explicitly set a new ``multiecho=True`` argument to the pipeline to wrap the interface into a `MapNode` instead of a `Node`?

niworkflows/interfaces/registration.py
```diff
@@ -347,8 +348,17 @@ def _post_run_hook(self, runtime):
...
+                               mandatory=True,
+                               desc=('4D EPI file. If multiple files '
+                                     'are provided, they are assumed '
+                                     'to represent multiple echoes '
+                                     'from the same run.'))
+    sbref_file = InputMultiObject(File(exists=True),
+                                  desc=("Single band reference image. "
+                                        "If multiple files are provided, "
+                                        "they are assumed to represent "
+                                        "multiple echoes."))
```
oesteban: Along the lines of the previous comment - perhaps add a new `Bool` input to this interface (something like `multiecho`) that allows `self.inputs.in_file` to be a list and then selects the first echo? Otherwise, identify the weird conditions (e.g., multiecho=False but list at input or multiecho=True and string at input).

niworkflows/interfaces/registration.py
```diff
@@ -372,17 +382,19 @@ class EstimateReferenceImage(SimpleInterface):
     output_spec = EstimateReferenceImageOutputSpec
 
     def _run_interface(self, runtime):
-        ref_name = self.inputs.in_file
+        # Select first EPI file
+        ref_name = self.inputs.in_file[0]
```
oesteban: This will select the first character of the input name if the `in_file` is only one path.
```suggestion
        ref_name = self.inputs.in_file
        if isinstance(self.inputs.in_file, list):
            ref_name = self.inputs.in_file[0]
```

niworkflows/interfaces/registration.py
```diff
@@ -372,17 +382,19 @@ class EstimateReferenceImage(SimpleInterface):
...
         n_volumes_to_discard = _get_vols_to_discard(ref_nii)
 
         self._results["n_volumes_to_discard"] = n_volumes_to_discard
 
         out_ref_fname = os.path.join(runtime.cwd, "ref_bold.nii.gz")
         if isdefined(self.inputs.sbref_file):
-            out_ref_fname = os.path.join(runtime.cwd, "ref_sbref.nii.gz")
-            ref_name = self.inputs.sbref_file
+            # Select first SBRef file
+            ref_name = self.inputs.sbref_file[0]
```
oesteban: Same here. InputMultiPath will flatten lists of len=1

niworkflows/interfaces/registration.py
effigies: Oscar, I tested this one to be sure. You've got input and output behavior reversed.

oesteban: Gotcha. Then this inline `ensure_list` is unnecessary right? https://github.com/poldracklab/niworkflows/pull/408/files#diff-7a5a9f735d33c7c55dc21e359cdac718R137

Also, there's no `ensure_list` for the sbref input.

effigies: I thought you needed to pass a list to a mapnode? And you don't need it for the sbref input because of the `InputMultiObject`.

niworkflows/interfaces/registration.py
tsalo: Per [this comment](https://github.com/poldracklab/niworkflows/pull/408#discussion_r331304129), should `self.inputs.in_file` and `self.inputs.sbref_file` be lists no matter what? Should I check the length of the list against the new `multiecho` parameter?

niworkflows/func/util.py
tsalo: Sorry, this was added in a04814f7c0c318194e17d651091cd436c2cca558 (and was fixed a couple of commits later).

tsalo:
Just wanted to ping this PR. The last time I checked, there was still some disagreement about the behavior of `InputMultiObject`. Should I treat single-item `InputMultiObject`s as strings or lists?

oesteban:
Hi, this is waiting for me to test it locally. Thank for your patience.

tsalo:
No pressure! I just wanted to make sure it wasn't waiting on me for anything.

oesteban:
Codacy Here is an overview of what got changed by this pull request:


```diff

Complexity increasing per file
==============================
- niworkflows/interfaces/registration.py  4
- niworkflows/func/util.py  2
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/nipreps/niworkflows/pullRequest?prid=5096912&bid=16766270)

oesteban:
Okay, I just rebased to upstream. @tsalo do you have some bandwidth to get this one to the finish line anytime soon?

niworkflows/utils/misc.py
```diff
@@ -212,6 +212,25 @@ def splitext(fname):
...
+    multi-echo data through workflows that only accept
+    a single file.
+
+    >>> select_first('some/file.nii.gz')
+    'some/file.nii.gz'
+    >>> select_first(['some/file1.nii.gz', 'some/file2.nii.gz'])
+    'some/file1.nii.gz'
+
+    """
+    if isinstance(in_files, list):
```
oesteban, line 228: just in case...
```suggestion
    if isinstance(in_files, (list, tuple)):
```

oesteban:
We have two regression tests on the masks with multi-echo data. Currently, we are just running the full masking workflow on each echo separately. I think it would make the most sense to test this PR with `multiecho=True`:

- ds210/sub-06 (3 echoes)
- ds216/sub-03 (4 echoes)

That will 1) exercise this PR making sure it does not introduce any regression; and 2) cut down test time as instead of 7 tests, we'll be running just 2.

WDYT?

tsalo:
> @tsalo do you have some bandwidth to get this one to the finish line anytime soon?

I think so.

> I think it would make the most sense to test this PR with `multiecho=True`

That makes sense to me, although I'm confused. Are you referring to a `skip ci`-like key phrase in my commit message?

oesteban:
> That makes sense to me, although I'm confused. Are you referring to a `skip ci`-like key phrase in my commit message?

I implemented the testing and sent you a PR (tsalo/niworkflows#1) to your branch. What I meant is that, when you accept that PR we are going to break this branch and I'll need to update the test data for it to pass again, if you are fine with that.

oesteban:
@effigies and @mgxd please have a last look before I rewrite the regression masks and break master with it, until this PR is merged. Before merging, I'll need to bump the identifiers for Circle's cache.

niworkflows/interfaces/registration.py
```diff
@@ -429,17 +455,27 @@ class EstimateReferenceImage(SimpleInterface):
...
+        # Select first EPI file
+        ref_name = self.inputs.in_file[0]
         ref_nii = nb.load(ref_name)
         n_volumes_to_discard = _get_vols_to_discard(ref_nii)
 
         self._results["n_volumes_to_discard"] = n_volumes_to_discard
 
+        if self.inputs.multiecho and (len(self.inputs.in_file) < 2):
+            raise ValueError("Argument 'multiecho' is True but "
+                             "'in_file' has only one element")
```
mgxd, line 467: we should keep all input checks either in `_format_arg()`, or at least at the top of the `_run_interface()` to avoid performing unnecessary operations

niworkflows/interfaces/registration.py
```diff
@@ -429,17 +455,27 @@ class EstimateReferenceImage(SimpleInterface):
...
+
         out_ref_fname = os.path.join(runtime.cwd, "ref_bold.nii.gz")
         if isdefined(self.inputs.sbref_file):
-            out_ref_fname = os.path.join(runtime.cwd, "ref_sbref.nii.gz")
-            ref_name = self.inputs.sbref_file
+            if self.inputs.multiecho and (len(self.inputs.sbref_file) < 2):
+                raise ValueError("Argument 'multiecho' is True but "
+                                 "'sbref_file' has only one element")
```
mgxd, line 473: same as above

niworkflows/func/util.py
```diff
@@ -148,15 +170,14 @@ def init_bold_reference_wf(
...
-        (inputnode, validate, [("bold_file", "in_file")]),
         (inputnode, gen_ref, [("sbref_file", "sbref_file")]),
         (inputnode, calc_dummy_scans, [("dummy_scans", "dummy_scans")]),
         (validate, gen_ref, [("out_file", "in_file")]),
         (gen_ref, enhance_and_skullstrip_bold_wf, [
             ("ref_image", "inputnode.in_file"),
         ]),
         (validate, outputnode, [
-            ("out_file", "bold_file"),
+            (("out_file", select_first), "bold_file"),
```
mgxd, line 180: I think we voted against using in-line functions within outputnode connections

oesteban: good catch - my bad!

effigies:
Not going to get to this tonight. I can look at it tomorrow afternoon, if this can wait.

niworkflows/interfaces/registration.py
oesteban: BaseInterfaces do not run `_format_arg()` I think, that's only for command lines I believe

oesteban:
I'm going to make this a draft so that it is isn't accidentally merged without updating the regression data (which I'll do right before merging).

niworkflows/func/util.py
```diff
@@ -104,10 +114,12 @@ def init_bold_reference_wf(
 
     """
     workflow = Workflow(name=name)
-    workflow.__desc__ = """\
+    workflow.__desc__ = f"""\
 First, a reference volume and its skull-stripped version were generated
-using a custom methodology of *fMRIPrep*.
+{'from the shortest echo of the BOLD run' * multiecho} using a custom
```
oesteban, line 119: It seems poldracklab/fmriprep#1803 is not picking this multi-echo note. Will follow-up when I catch the issue.

niworkflows/func/util.py
```diff
@@ -179,9 +197,39 @@ def init_bold_reference_wf(
...
+            (val_sbref, gen_ref, [("out_file", "sbref_file")]),
+        ])
+        # fmt: on
+
+        # Edit the boilerplate as the SBRef will be the reference
+        workflow.__desc__ = f"""\
+First, a reference volume and its skull-stripped version were generated
+by aligning and averaging{' the first echo of' * multiecho}
+{nsbrefs or ''} single-band references (SBRefs).
+"""
```
oesteban, line 231: Closes #413

niworkflows/func/util.py
oesteban: Fixed.

oesteban:
Thumbs up?

oesteban:
Okay, I just uploaded the new masks and started https://app.circleci.com/pipelines/github/oesteban/niworkflows/115/workflows/3236dc2a-0115-4f3e-88d5-ab31cdd332f4

Once that build is passing, I'll merge this PR.

PR #417, opened 2019-10-30
ENH: expand GenerateCifti & MedialNaNs interfaces to support HCP grayordinates / fsLR surfaces

mgxd:
Enhance `GenerateCifti`, `MedialNaNs`, and `CiftiNameSource` interfaces to support HCP grayordinates.

This change depends on templateflow/python-client#31

### DIFF
--- a/niworkflows/interfaces/cifti.py
+++ b/niworkflows/interfaces/cifti.py
@@ -135,7 +135,7 @@ class CiftiNameSource(SimpleInterface):
     def _run_interface(self, runtime):
         suffix = 'bold.dtseries'
         if 'hcp' in self.inputs.variant.lower():
-             suffix = 'space-fsLR_bold.dtseries'
+            suffix = 'space-fsLR_bold.dtseries'
         self._results['out_name'] = suffix
         return runtime
 
--- a/niworkflows/interfaces/cifti.py
+++ b/niworkflows/interfaces/cifti.py
@@ -58,9 +58,8 @@ class _GenerateCiftiInputSpec(BaseInterfaceInputSpec):
                                 desc="CIFTI volumetric output space")
     surface_target = traits.Enum("fsLR", "fsaverage5", "fsaverage6", usedefault=True,
                                  desc="CIFTI surface target space")
-    density = traits.Enum(32, 59, 164, usedefault=True,
+    density = traits.Enum('32k', '59k', usedefault=True,
                           help='surface hemisphere vertices')
-    resolution = traits.Enum(2, 1.6, usedefault=True, help='volume resolution (mm)')
     TR = traits.Float(mandatory=True, desc="Repetition time")
     surface_bolds = traits.List(File(exists=True), mandatory=True,
                                 desc="list of surface BOLD GIFTI files"
@@ -138,6 +137,7 @@ class GenerateCifti(SimpleInterface):
         annotation_files = None
         if self.inputs.volume_target == "MNI152NLin2009cAsym":
             tpl_kwargs.update({
+                'resolution': '2',
                 'desc': 'DKT31',
             })
             annotation_files = sorted(
@@ -148,9 +148,11 @@ class GenerateCifti(SimpleInterface):
             )
 
         elif self.inputs.volume_target == 'MNI152NLin6Asym':
+            res = '2' if self.inputs.density == '32k' else '5'
+
             tpl_kwargs.update({
                 'atlas': 'fsLR',
-                'resolution': self.inputs.resolution,
+                'resolution': res,
             })
             annotation_files = [
                 str(f) for f in get_template(
--- a/niworkflows/interfaces/cifti.py
+++ b/niworkflows/interfaces/cifti.py
@@ -13,7 +13,7 @@ from nilearn.image import resample_to_img
 from nipype.utils.filemanip import split_filename
 from nipype.interfaces.base import (
     BaseInterfaceInputSpec, TraitedSpec, File, traits,
-    SimpleInterface, Directory
+    SimpleInterface
 )
 from templateflow.api import get as get_template
 
@@ -58,13 +58,16 @@ class _GenerateCiftiInputSpec(BaseInterfaceInputSpec):
                                 desc="CIFTI volumetric output space")
     surface_target = traits.Enum("fsLR", "fsaverage5", "fsaverage6", usedefault=True,
                                  desc="CIFTI surface target space")
-    density = traits.Enum('32k', '59k', '164k', usedefault=True, help='surface hemisphere vertices')
+    density = traits.Enum('32k', '59k', '164k', usedefault=True,
+                          help='surface hemisphere vertices')
     resolution = traits.Enum(2, 1.6, usedefault=True, help='volume resolution (mm)')
     TR = traits.Float(mandatory=True, desc="Repetition time")
     surface_bolds = traits.List(File(exists=True), mandatory=True,
-                                desc="list of surface BOLD GIFTI files (length 2 with order [L,R])")
+                                desc="list of surface BOLD GIFTI files"
+                                     " (length 2 with order [L,R])")
     surface_annots = traits.List(File(exists=True),
-                                 help="Surface annotation files. Used to exclude medial wall vertices.")
+                                 help="Surface annotation files."
+                                      " Used to exclude medial wall vertices.")
 
 
 class _GenerateCiftiOutputSpec(TraitedSpec):
@@ -124,7 +127,7 @@ class GenerateCifti(SimpleInterface):
     def _fetch_data(self):
         """Converts inputspec to files"""
         if (
-            self.inputs.surface_target not in  ("fsaverage5", "fsaverage6", "fsLR") or
+            self.inputs.surface_target not in ("fsaverage5", "fsaverage6", "fsLR") or
             self.inputs.volume_target not in ("MNI152NLin2009cAsym", "MNI152NLin6Asym")
         ):
             raise NotImplementedError(
@@ -161,13 +164,13 @@ class GenerateCifti(SimpleInterface):
         return annotation_files, label_file
 
     @staticmethod
-    def _create_cifti_image(bold_img, label_file, bold_surfs,
+    def _create_cifti_image(bold_file, label_file, bold_surfs,
                             targets, tr, annotation_files=None):
         """
         Generate CIFTI image in target space.
 
         Parameters
-            bold_img : `SpatialImage`
+            bold_file : str
                 BOLD volumetric timeseries
             label_file : str
                 Subcortical label file
@@ -184,6 +187,7 @@ class GenerateCifti(SimpleInterface):
             out :
                 BOLD data saved as CIFTI dtseries
         """
+        bold_img = nb.load(bold_file)
         label_img = nb.load(label_file)
         if label_img.shape != bold_img.shape:
             bold_img = resample_to_img(bold_img, label_img)
@@ -282,7 +286,7 @@ class GenerateCifti(SimpleInterface):
         img = ci.Cifti2Image(bm_ts, hdr)
         img.nifti_header.set_intent('NIFTI_INTENT_CONNECTIVITY_DENSE_SERIES')
 
-        _, out_base, _ = split_filename(bold_img)
+        _, out_base, _ = split_filename(bold_file)
         out_file = "{}.dtseries.nii".format(out_base)
         ci.save(img, out_file)
         return os.path.join(os.getcwd(), out_file)
--- a/niworkflows/interfaces/cifti.py
+++ b/niworkflows/interfaces/cifti.py
@@ -58,7 +58,7 @@ class _GenerateCiftiInputSpec(BaseInterfaceInputSpec):
                                 desc="CIFTI volumetric output space")
     surface_target = traits.Enum("fsLR", "fsaverage5", "fsaverage6", usedefault=True,
                                  desc="CIFTI surface target space")
-    density = traits.Enum('32k', '59k', '164k', usedefault=True,
+    density = traits.Enum(32, 59, 164, usedefault=True,
                           help='surface hemisphere vertices')
     resolution = traits.Enum(2, 1.6, usedefault=True, help='volume resolution (mm)')
     TR = traits.Float(mandatory=True, desc="Repetition time")
@@ -92,10 +92,10 @@ class GenerateCifti(SimpleInterface):
         self._results["out_file"] = self._create_cifti_image(
             self.inputs.bold_file,
             label_file,
-            annotation_files,
-            self.inputs.surface_files,
+            self.inputs.surface_bolds,
             targets,
-            self.inputs.TR)
+            self.inputs.TR,
+            annotation_files=annotation_files)
         return runtime
 
     def _define_variant(self):
@@ -134,12 +134,11 @@ class GenerateCifti(SimpleInterface):
                 )
             )
 
-        tpl_kwargs = {'resolution': self.inputs.resolution}
+        tpl_kwargs = {'suffix': 'dseg'}
         annotation_files = None
         if self.inputs.volume_target == "MNI152NLin2009cAsym":
             tpl_kwargs.update({
                 'desc': 'DKT31',
-                'suffix': 'dseg',
             })
             annotation_files = sorted(
                 glob(os.path.join(self.inputs.subjects_dir,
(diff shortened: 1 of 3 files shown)


niworkflows/interfaces/cifti.py
```diff
@@ -55,14 +61,15 @@
 
 class GenerateCiftiInputSpec(BaseInterfaceInputSpec):
     bold_file = File(mandatory=True, exists=True, desc="input BOLD file")
-    volume_target = traits.Enum("MNI152NLin2009cAsym", mandatory=True, usedefault=True,
+    volume_target = traits.Enum("MNI152NLin2009cAsym", "MNI152NLin6Asym", mandatory=True,
```
oesteban:
```suggestion
    volume_target = traits.Enum("MNI152NLin6Asym", "MNI152NLin2009cAsym", usedefault=True,
```

niworkflows/interfaces/cifti.py
```diff
@@ -55,14 +61,15 @@
...
                                 desc="CIFTI volumetric output space")
-    surface_target = traits.Enum("fsaverage5", "fsaverage6", mandatory=True,
-                                 usedefault=True, desc="CIFTI surface target space")
-    subjects_dir = Directory(mandatory=True, desc="FreeSurfer SUBJECTS_DIR")
-    TR = traits.Float(mandatory=True, desc="repetition time")
-    gifti_files = traits.List(File(exists=True), mandatory=True,
-                              desc="list of surface geometry files (length 2 with order [L,R])")
+    surface_target = traits.Enum("fsaverage5", "fsaverage6", "fsLR", mandatory=True,
```
oesteban:
```suggestion
    surface_target = traits.Enum("fsLR", "fsaverage5", "fsaverage6", usedefault=True,
```

niworkflows/interfaces/cifti.py
```diff
@@ -75,98 +82,135 @@ class GenerateCifti(SimpleInterface):
...
-    * target surfaces: fsaverage5, fsaverage6
-    * target volumes: OASIS-TRT-20_DKT31 labels in MNI152NLin2009cAsym
+    target surfaces:
+        - fsaverage5
+        - fsaverage6
+        - fsLR
+    target volumes:
+        - OASIS-TRT-20_DKT31 labels in MNI152NLin2009cAsym (2mm)
+        - MNI152NLin6Asym (1.6)
     """
```
oesteban:
```suggestion
    Generate CIFTI image from BOLD file in target spaces.
    
    Currently supports ``fsLR`` or ``fsaverage`` for template surfaces and
    ``MNI152NLin6Asym`` or ``MNI152NLin2009cAsym`` as template volumes.
    
    """
```

niworkflows/interfaces/cifti.py
```diff
@@ -55,14 +61,15 @@
...
                                 desc="CIFTI volumetric output space")
-    surface_target = traits.Enum("fsaverage5", "fsaverage6", mandatory=True,
-                                 usedefault=True, desc="CIFTI surface target space")
-    subjects_dir = Directory(mandatory=True, desc="FreeSurfer SUBJECTS_DIR")
-    TR = traits.Float(mandatory=True, desc="repetition time")
-    gifti_files = traits.List(File(exists=True), mandatory=True,
-                              desc="list of surface geometry files (length 2 with order [L,R])")
+    surface_target = traits.Enum("fsaverage5", "fsaverage6", "fsLR", mandatory=True,
+                                 desc="CIFTI surface target space")
+    TR = traits.Float(mandatory=True, desc="Repetition time")
```
oesteban: Please, add inputs for density and resolution. Defaults should be ``fsLR``, ``32k``, and 1.6mm.

niworkflows/interfaces/cifti.py
```diff
@@ -75,98 +82,135 @@ class GenerateCifti(SimpleInterface):
...
+        ):
+            raise NotImplementedError(
+                "Target space (surface: {0}, volume: {1}) is not supported".format(
+                    self.inputs.surface_target, self.inputs.volume_target
+                )
+            )
+
+        tpl_kwargs = {}
+        annotation_files = None
+        if self.inputs.surface_target.startswith("fsav"):
```
oesteban: I guess this decision should be made based on whether MNI6 OR MNI2009c was used?

niworkflows/interfaces/cifti.py
```diff
@@ -75,98 +82,135 @@ class GenerateCifti(SimpleInterface):
...
+                            self.inputs.surface_target, self.inputs.subjects_dir))
+        elif self.inputs.surface_target == 'fsLR':
+            tpl_kwargs.update({
+                'resolution': 5,
+                'atlas': 'fsLR',
+                'desc': 'aseg',
+                'suffix': 'dseg'
+            })
+
+        label_file = str(get_template("tpl-{}".format(self.inputs.volume_target), **tpl_kwargs))
```
oesteban:
```suggestion
        label_file = str(get_template(self.inputs.volume_target, **tpl_kwargs))
```

niworkflows/interfaces/cifti.py
mgxd: if we want to be in line with 32k, it looks like the 2mm template is used
https://github.com/Washington-University/HCPpipelines/tree/master/global/templates/91282_Greyordinates

oesteban: Updated my comment

mgxd:
@oesteban @effigies i'm not sure why this is failing, any ideas?

oesteban:
That is a very outdated tests. I'd like to get rid of the data getters for niworkflows 1.0 (and then those tests will not be necessary as no environment variables or cache dirs will be necessary)

niworkflows/interfaces/freesurfer.py
```diff
@@ -294,7 +294,7 @@ def _run_interface(self, runtime):
 class _MedialNaNsInputSpec(BaseInterfaceInputSpec):
     in_file = File(exists=True, mandatory=True, desc='input surface file')
     subjects_dir = Directory(mandatory=True, desc='FreeSurfer SUBJECTS_DIR')
-    density = traits.Enum('32', '59', '164', desc="Input file density (fsLR only)")
+    density = traits.Enum(32, 59, 164, desc="Input file density (fsLR only)")
```
oesteban:
```suggestion
    density = traits.Enum('32k', '59k', '164k', desc="Input file density (fsLR only)")
```
density is string (actually, the resolution should be string too).

mgxd: Yup, I think as string would be best - I would love to just pass in the `den-` key but
```
In [1]: import templateflow.api as tf                                                                                                                         

In [2]: tf.get('fsLR', density='32k')                                                                                                                         
Out[2]: []

In [3]: tf.get('fsLR', density=32)  # or '32'
Downloading https://templateflow.s3.amazonaws.com/tpl-fsLR/tpl-fsLR_hemi-L_den-32k_nomedialwall.surf.gii
...
```

oesteban: ugh, can you file an issue in templateflow/python-client?

niworkflows/interfaces/cifti.py
```diff
@@ -68,112 +78,141 @@ class GenerateCifti(SimpleInterface):
...
+            })
+            annotation_files = sorted(
+                glob(os.path.join(self.inputs.subjects_dir,
+                                  self.inputs.surface_target,
+                                  'label',
+                                  '*h.aparc.annot'))
+            )
+
+        elif self.inputs.volume_target == 'MNI152NLin6Asym':
+            res = '2' if self.inputs.density == '32k' else '6'
```
oesteban: Perhaps a stricter value checking?
```suggestion
            res = {'59k': '6', '32k': '2}[self.inputs.density]
```

niworkflows/interfaces/cifti.py
```diff
@@ -68,112 +78,141 @@ class GenerateCifti(SimpleInterface):
...
     @staticmethod
-    def _create_cifti_image(bold_file, label_file, annotation_files, gii_files,
-                            volume_target, surface_target, tr):
+    def _create_cifti_image(bold_file, label_file, bold_surfs,
+                            targets, tr, annotation_files):
         """
         Generate CIFTI image in target space.
 
         Parameters
-        ----------
```
oesteban:
```suggestion
        ----------
```

The section markers are needed for sphinxcontrib-napoleon to correctly format these docstrings

niworkflows/interfaces/cifti.py
```diff
@@ -15,47 +16,62 @@
...
     bold_file = File(mandatory=True, exists=True, desc="input BOLD file")
-    volume_target = traits.Enum("MNI152NLin2009cAsym", mandatory=True, usedefault=True,
+    volume_target = traits.Enum("MNI152NLin6Asym", "MNI152NLin2009cAsym", usedefault=True,
                                 desc="CIFTI volumetric output space")
-    surface_target = traits.Enum("fsaverage5", "fsaverage6", mandatory=True,
-                                 usedefault=True, desc="CIFTI surface target space")
+    surface_target = traits.Enum("fsLR", "fsaverage5", "fsaverage6", usedefault=True,
+                                 desc="CIFTI surface target space")
+    density = traits.Enum(None, '32k', '59k', usedefault=True,
+                          help='surface hemisphere vertices')
```
oesteban: Please define what `None` means in the desc (BTW, the appropriate argument name is `desc`).

```suggestion
                          desc='surface hemisphere vertices')
```

niworkflows/interfaces/cifti.py
```diff
@@ -272,7 +134,262 @@ class CiftiNameSource(SimpleInterface):
 
     def _run_interface(self, runtime):
         suffix = 'bold.dtseries'
-        if 'hcp' in self.inputs.variant:
+        if 'hcp' in self.inputs.variant.lower():
             suffix = 'space-hcp_bold.dtseries'
```
oesteban: shouldn't this be?
```suggestion
             suffix = 'space-fsLR_bold.dtseries'
```

niworkflows/interfaces/freesurfer.py
```diff
@@ -480,20 +480,33 @@ def grow_mask(anat, aseg, ants_segs=None, ww=7, zval=2.0, bw=4):
...
-    cortex = nb.freesurfer.read_label(os.path.join(
-        subjects_dir, target_subject, 'label', '{}.cortex.label'.format(fn[:2])))
     func = nb.load(in_file)
-    medial = np.delete(np.arange(len(func.darrays[0].data)), cortex)
+    if target_subject.startswith('fsaverage'):
+        cortex = nb.freesurfer.read_label(os.path.join(
+            subjects_dir, target_subject, 'label', '{}.cortex.label'.format(fn[:2])))
+        medial = np.delete(np.arange(len(func.darrays[0].data)), cortex)
+    elif target_subject == 'fslr' and den is not None:
+        hemi = 'L' if fn[:2] == 'lh' else 'R'
```
oesteban:
```suggestion
        hemi = fn[0].upper()
```

niworkflows/interfaces/cifti.py
```diff
@@ -272,7 +134,262 @@ class CiftiNameSource(SimpleInterface):
...
+    'HCP grayordinates'
+
+    """
+    variant = None
+    for space, targets in CIFTI_VARIANTS.items():
+        if all(target in targets for target in (surface_target, volume_target)):
+            variant = space
+            break
+    if variant is None:
+        raise NotImplementedError
```
oesteban: Please add a message

niworkflows/interfaces/cifti.py
mgxd: the CIFTI has `fsLR` surfaces, but `MNI152NLin6Asym` volumes - HCP outputs are in the same space, though not sure what the best label for this would be

oesteban: The definition of `fsLR` implicitly defines `MNI152NLin6Asym` as the volumetric template reference. Hence, saying `space-fsLR` univocally prescribes that surfaces are from the HCP atlas and the volumetric reference is FSL's MNI.

In other words, I can confirm this should be `space-fsLR`.

oesteban:
@mgxd can you address the overindentation issue?

niworkflows/interfaces/cifti.py
```diff
@@ -15,47 +16,62 @@
...
+    'CIFTI_STRUCTURE_PALLIDUM_LEFT': (13,),
+    'CIFTI_STRUCTURE_PALLIDUM_RIGHT': (52,),
+    'CIFTI_STRUCTURE_PUTAMEN_LEFT': (12,),
+    'CIFTI_STRUCTURE_PUTAMEN_RIGHT': (51,),
+    'CIFTI_STRUCTURE_THALAMUS_LEFT': (10,),
+    'CIFTI_STRUCTURE_THALAMUS_RIGHT': (49,),
+}
+CIFTI_VARIANTS = {
+    'HCP grayordinates': ('fsLR', 'MNI152NLin6Asym'),
+    'space1': ('fsaverage5', 'MNI152NLin2009cAsym'),
```
oesteban: Can we maybe merge space1 and space2 with:
```suggestion
    'fMRIPrep grayordinates': ('fsaverage', 'MNI152NLin2009cAsym'),
```

The idea would be that density could be encoded in the filenames, and this would be more consistent.

niworkflows/interfaces/cifti.py
```diff
@@ -15,47 +16,62 @@
...
 class _GenerateCiftiInputSpec(BaseInterfaceInputSpec):
     bold_file = File(mandatory=True, exists=True, desc="input BOLD file")
-    volume_target = traits.Enum("MNI152NLin2009cAsym", mandatory=True, usedefault=True,
+    volume_target = traits.Enum("MNI152NLin6Asym", "MNI152NLin2009cAsym", usedefault=True,
                                 desc="CIFTI volumetric output space")
-    surface_target = traits.Enum("fsaverage5", "fsaverage6", mandatory=True,
-                                 usedefault=True, desc="CIFTI surface target space")
+    surface_target = traits.Enum("fsLR", "fsaverage5", "fsaverage6", usedefault=True,
+                                 desc="CIFTI surface target space")
+    density = traits.Enum(None, '32k', '59k', usedefault=True,
```
oesteban: What does `None` mean here? - please add to the `desc` field.

niworkflows/interfaces/cifti.py
```diff
@@ -68,196 +84,42 @@ class GenerateCifti(SimpleInterface):
...
-        self._results["variant_key"], self._results["variant"] = self._define_variant()
-        annotation_files, label_file = self._fetch_data()
-        self._results["out_file"] = self._create_cifti_image(
+        annotation_files, label_file = _get_cifti_data(
+            self.inputs.surface_target,
+            self.inputs.volume_target,
+            self.inputs.subjects_dir,
+            self.inputs.density
+        )
+        self._results['variant_key'], self._results['variant'] = _get_cifti_variant(
```
oesteban: feels like `variant_key` and `variant` meanings are swapped. The `_key` suffix makes me think of a string not a file.

niworkflows/interfaces/cifti.py
```diff
@@ -272,7 +134,266 @@ class CiftiNameSource(SimpleInterface):
...
+    Returns
+    -------
+    out :
+        BOLD data saved as CIFTI dtseries
+    """
+    bold_img = nb.load(bold_file)
+    label_img = nb.load(label_file)
+    if label_img.shape != bold_img.shape[:3]:
+        warnings.warn("Resampling bold volume to match label dimensions")
+        bold_img = resample_to_img(bold_img, label_img)
```
oesteban: This should not be necessary (see comment below).

niworkflows/interfaces/cifti.py
```diff
@@ -272,7 +134,266 @@ class CiftiNameSource(SimpleInterface):
...
+        # add each brain structure to list
+        brainmodels.append(bm)
+
+    # add volume information
+    brainmodels.append(
+        ci.Cifti2Volume(
+            bold_img.shape[:3],
+            ci.Cifti2TransformationMatrixVoxelIndicesIJKtoXYZ(-3, bold_img.affine)
+        )
+    )
```
oesteban: I believe here's where you sample the subcortical structures from the BOLD data. Now that we have learned so much from nitransforms, you can do this step without the resampling above.

The pseudocode would be:

```
ijk_labels = np.where(label_data > 0)  # I don't remember if you need to set an upper bound threshold
ijk_bold = np.linalg.inv(bold_img.affine).dot(label_img.affine.dot(ijk_labels))  # You'll need to take care of ijk_labels, add a ones column and probably transpose here.
bold_values = scipy.ndimage.map_coordinates(bold_data, ijk_bold, order=3)
```

mgxd: this looks really cool, though i'm inclined to leave this as is for now - we will likely address this and other external resamplings once nitransforms is incorporated into the code base

niworkflows/interfaces/cifti.py
mgxd: i've removed `None` - this doesn't have to be set

PR #418, opened 2019-10-31
install: use scikit-image LTS for earlier python versions

mgxd:


setup.cfg
```diff
@@ -19,6 +19,8 @@ classifiers =
 
 [options]
 python_requires = >= 3.5
+setup_requires =
+    numpy
```
effigies: Why?

mgxd: ah! I was wondering why it wasn't working. when downgrading scikit-image:

```
Best match: scikit-image 0.14.5
Processing scikit-image-0.14.5.tar.gz
Writing /tmp/easy_install-yr5hpogx/scikit-image-0.14.5/setup.cfg
Running scikit-image-0.14.5/setup.py -q bdist_egg --dist-dir /tmp/easy_install-yr5hpogx/scikit-image-0.14.5/egg-dist-tmp-cozbgkai
To install scikit-image from source, you will need numpy.
Install numpy with pip:
pip install numpy
```

effigies: :-/ I don't think we should change our setup requires for scikit-image. They have Py3.5 wheels for [0.14.4](https://pypi.org/project/scikit-image/0.14.4/#files).

We can also drop Py3.5 support, if that makes life easier.

effigies: Ugh. Looks like they had to [remove `pyproject.toml`](https://github.com/scikit-image/scikit-image/pull/3688), so `pip install scikit-image` won't know to grab `numpy` anymore.

mgxd: we should probably just remove 3.5 support soon, but since this is so closely tied with fmriprep there should be some coordination. i just wanted to get a quick fix in since tests were broken

effigies:
I'm okay merging with just Travis passing.

PR #432, opened 2019-11-27
MNT: Specify junit_family to suppress pytest DeprecationWarning

effigies:
https://docs.pytest.org/en/latest/deprecations.html#junit-family-default-value-change-to-xunit2


effigies:
Test summaries still generated, `DeprecationWarning` resolved. Moved to the new system, so hopefully more future-stable.

PR #435, opened 2019-12-04
ENH: Add ``n4_only`` workflow -- to skip brain extraction

adelavega:
Allows skipping brain extraction with n4_only workflow.

See: https://github.com/poldracklab/smriprep/pull/116

I think the main remaining todo is this comment from @oesteban:

> I've been looking at some masks after binarization. I think they would benefit a lot from a binary opening followed by closing to remove the irregular boundary. Also, it would also make sense to fill holes. We might want to implement that as part of the Binarize interface. I'll have a look into this tomorrow.

### DIFF
--- a/niworkflows/anat/ants.py
+++ b/niworkflows/anat/ants.py
@@ -659,7 +659,7 @@ def init_n4_only_wf(atropos_model=None,
         Allows to specify a particular segmentation model, overwriting
         the defaults based on ``bids_suffix``
     name : str, optional
-        Workflow name (default: antsBrainExtraction)
+        Workflow name (default: ``'n4_only_wf'``).
 
     Inputs
     ------
@@ -728,6 +728,12 @@ def init_n4_only_wf(atropos_model=None,
 
     # If atropos refine, do in4 twice
     if atropos_refine:
+        # Morphological dilation, radius=2
+        dil_brainmask = pe.Node(ImageMath(operation='MD', op2='2'),
+                                name='dil_brainmask')
+        # Get largest connected component
+        get_brainmask = pe.Node(ImageMath(operation='GetLargestComponent'),
+                                name='get_brainmask')
         atropos_model = atropos_model or list(
             ATROPOS_MODELS[bids_suffix].values())
         atropos_wf = init_atropos_wf(
@@ -752,8 +758,10 @@ def init_n4_only_wf(atropos_model=None,
                 ('output_image', 'inputnode.in_files')]),
             (thr_brainmask, atropos_wf, [
                 ('out_mask', 'inputnode.in_mask')]),
-            (thr_brainmask, atropos_wf, [
-                ('out_mask', 'inputnode.in_mask_dilated')]),  # Dilate?
+            (thr_brainmask, dil_brainmask, [('output_image', 'op1')]),
+            (dil_brainmask, get_brainmask, [('output_image', 'op1')]),
+            (get_brainmask, atropos_wf, [
+                ('output_image', 'inputnode.in_mask_dilated')]),
             (atropos_wf, sel_wm, [('outputnode.out_tpms', 'inlist')]),
             (sel_wm, inu_n4_final, [('out', 'weight_image')]),
             (atropos_wf, outputnode, [
--- a/niworkflows/anat/ants.py
+++ b/niworkflows/anat/ants.py
@@ -311,7 +311,7 @@ N4BiasFieldCorrection.""" % _ants_version, DeprecationWarning)
             ('reverse_transforms', 'transforms'),
             ('reverse_invert_flags', 'invert_transform_flags')]),
         (map_brainmask, thr_brainmask, [('output_image', 'input_image')]),
-        (thr_brainmask, dil_brainmask, [('out_mask', 'op1')]),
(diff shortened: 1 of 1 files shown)


niworkflows/anat/ants.py
```diff
@@ -613,6 +614,156 @@ def init_atropos_wf(name='atropos_wf',
...
+    atropos_refine : bool
+        Enables or disables the whole ATROPOS sub-workflow
+    atropos_use_random_seed : bool
+        Whether ATROPOS should generate a random seed based on the
+        system's clock
+    atropos_model : tuple or None
+        Allows to specify a particular segmentation model, overwriting
+        the defaults based on ``bids_suffix``
+    name : str, optional
+        Workflow name (default: antsBrainExtraction)
```
oesteban:
```suggestion
        Workflow name (default: ``'n4_only_wf'``).
```

niworkflows/anat/ants.py
```diff
@@ -613,6 +614,156 @@ def init_atropos_wf(name='atropos_wf',
...
+        (inputnode, inu_n4_final, [('in_files', 'input_image')]),
+        (inputnode, thr_brainmask, [(('in_files', _pop), 'in_file')]),
+        (thr_brainmask, outputnode, [('out_mask', 'out_mask')]),
+        (inu_n4_final, outputnode, [('output_image', 'out_file')]),
+        (inu_n4_final, outputnode, [('output_image', 'bias_corrected')]),
+        (inu_n4_final, outputnode, [('bias_image', 'bias_image')])
+    ])
+
+    # If atropos refine, do in4 twice
+    if atropos_refine:
```
oesteban: Let's add the binary dilation
```suggestion
    if atropos_refine:
        # Morphological dilation, radius=2
        dil_brainmask = pe.Node(ImageMath(operation='MD', op2='2'),
                                name='dil_brainmask')
        # Get largest connected component
        get_brainmask = pe.Node(ImageMath(operation='GetLargestComponent'),
                                name='get_brainmask')
```

niworkflows/anat/ants.py
```diff
@@ -613,6 +614,156 @@ def init_atropos_wf(name='atropos_wf',
...
+                shrink_factor=4, bspline_fitting_distance=200),
+            n_procs=omp_nthreads, name='inu_n4', iterfield=['input_image'])
+
+        wf.connect([
+            (inputnode, inu_n4, [('in_files', 'input_image')]),
+            (inu_n4, atropos_wf, [
+                ('output_image', 'inputnode.in_files')]),
+            (thr_brainmask, atropos_wf, [
+                ('out_mask', 'inputnode.in_mask')]),
+            (thr_brainmask, atropos_wf, [
```
oesteban:
```suggestion
            (thr_brainmask, dil_brainmask, [('output_image', 'op1')]),
            (dil_brainmask, get_brainmask, [('output_image', 'op1')]),
            (get_brainmask, atropos_wf, [
```

niworkflows/anat/ants.py
```diff
@@ -613,6 +614,156 @@ def init_atropos_wf(name='atropos_wf',
...
+                ('out_mask', 'inputnode.in_mask_dilated')]),  # Dilate?
```
oesteban:
```suggestion
                ('output_image', 'inputnode.in_mask_dilated')]),
```

niworkflows/anat/ants.py
```diff
@@ -310,7 +311,7 @@ def init_brain_extraction_wf(name='brain_extraction_wf',
             ('reverse_transforms', 'transforms'),
             ('reverse_invert_flags', 'invert_transform_flags')]),
         (map_brainmask, thr_brainmask, [('output_image', 'input_image')]),
-        (thr_brainmask, dil_brainmask, [('output_image', 'op1')]),
+        (thr_brainmask, dil_brainmask, [('out_mask', 'op1')]),
```
oesteban:
```suggestion
        (thr_brainmask, dil_brainmask, [('output_image', 'op1')]),
```

adelavega: Are you sure? That connection wasn't working, which is why I changed it.

adelavega: Ah, I see the confusion. We are using `Binarize`, which has an output `out_mask` not `output_image` (that's if we were using the ants function, but we're not).

niworkflows/anat/ants.py
```diff
@@ -613,6 +614,164 @@ def init_atropos_wf(name='atropos_wf',
...
+    try:
+        inu_n4_final.inputs.rescale_intensities = True
+    except ValueError:
+        warn("The installed ANTs version too old. Please consider upgrading to "
+             "2.1.0 or greater.", DeprecationWarning)
+
+    wf.connect([
+        (inputnode, inu_n4_final, [('in_files', 'input_image')]),
+        (inputnode, thr_brainmask, [(('in_files', _pop), 'in_file')]),
+        (thr_brainmask, outputnode, [('out_mask', 'out_mask')]),
```
adelavega: Should the dilated mask be output, or the undilated one (which is what we're doing here)

oesteban: I'd say the undilated. I believe we should refactor the Binarize interface so that it can do the binary closing and opening, which will give a smoother mask (and that will be good for the dilated mask too).

PR #438, opened 2019-12-06
[FIX] handle case when dummy_scans is 0

### LINKED ISSUE #437: _pass_dummy_scans returns wrong value when dummy_scans set to 0
ref poldracklab/fmriprep#1904

jdkent:
fixes #437 

adds tests to confirm functionality.

### DIFF
--- a/niworkflows/func/util.py
+++ b/niworkflows/func/util.py
@@ -446,5 +446,4 @@ def _pass_dummy_scans(algo_dummy_scans, dummy_scans=None):
     """
     if dummy_scans is None:
         return algo_dummy_scans
-    else:
-        return dummy_scans
+    return dummy_scans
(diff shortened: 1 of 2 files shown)


niworkflows/func/util.py
```diff
@@ -444,4 +444,7 @@ def _pass_dummy_scans(algo_dummy_scans, dummy_scans=None):
         number of volumes to skip
 
     """
-    return dummy_scans or algo_dummy_scans
+    if dummy_scans is None:
+        return algo_dummy_scans
+    else:
```
oesteban:
```suggestion
    return dummy_scans
```

niworkflows/func/util.py
```diff
@@ -444,4 +444,7 @@ def _pass_dummy_scans(algo_dummy_scans, dummy_scans=None):
...
+        return dummy_scans
```
oesteban:
```suggestion
```

effigies:
LGTM. Obviously correct, so I'm +1 for merging without waiting on tests.

PR #441, opened 2019-12-17
FIX: ``IntraModalMerge`` failed for dims (x, y, z, 1)

oesteban:
See poldracklab/sdcflows#77.

I'll try to add a regression test for this ASAP. For now, it'd be good to have your opinion on how I'm using nibabel, Chris.

### DIFF
--- a/niworkflows/interfaces/images.py
+++ b/niworkflows/interfaces/images.py
@@ -12,7 +12,7 @@ from nipype import logging
 from nipype.utils.filemanip import fname_presuffix
 from nipype.interfaces.base import (
     traits, TraitedSpec, BaseInterfaceInputSpec, SimpleInterface,
-    File, InputMultiPath, OutputMultiPath)
+    File, InputMultiPath, OutputMultiPath, isdefined)
 from nipype.interfaces import fsl
 
 LOGGER = logging.getLogger('nipype.interface')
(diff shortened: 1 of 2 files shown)


niworkflows/interfaces/images.py
```diff
@@ -42,50 +50,48 @@ def _run_interface(self, runtime):
...
-                              ref_vol=0, in_file=in_files)
-        mcres = mcflirt.run()
-        self._results['out_mats'] = mcres.outputs.mat_file
-        self._results['out_movpar'] = mcres.outputs.par_file
-        self._results['out_file'] = mcres.outputs.out_file
-
-        hmcnii = nb.load(mcres.outputs.out_file)
-        hmcdat = hmcnii.get_fdata().mean(axis=3)
+                raise RuntimeError('Input image (%s) is 5D.' % f)
+            nii_list += nb.four_to_three(filenii)
```
effigies: `four_to_three` assumes 4D images, so 3D images and 4D images with a single volume (thus squeezed down to 3D) will fail.

niworkflows/interfaces/images.py
```diff
@@ -42,50 +50,48 @@ def _run_interface(self, runtime):
...
+
+        merged_fname = fname_presuffix(self.inputs.in_files[0],
+                                       suffix='_merged', newpath=runtime.cwd)
+        filenii = nb.concat_images(nii_list)
+        filenii.to_filename(merged_fname)
+
+        self._results['out_file'] = merged_fname
+        self._results['out_avg'] = merged_fname
+        if filenii.dataobj.ndim < 4:
```
effigies: I don't think this will ever happen. I'd need to check, but I think concatenating a single image will still throw on an extra dimension at the end.

oesteban: oookay. thanks again!

Issue #440, opened 2019-12-11
More generalizable summary interfaces


oesteban:
It is becoming more apparent that a more generalizable code for summary interfaces is necessary.

Additionally, in the context of multi-site studies, we should have more flexibility to add more metadata details such as the scanner vendor, model and software version.

This issue, therefore, implies:

- [ ] Centralizing summary interfaces in niworkflows (i.e., remove the code from fMRIPrep and dMRIPrep, so that just setting a template would suffice).
- [ ] Flexibilizing the inputs of the interface - i.e. parse the template to make out the inputs of a DynamicInputTrait specification.
- [ ] Implement a language/protocol to tell these interfaces when to pull information out of the BIDS metadata.

PR #443, opened 2020-01-03
ENH/FIX: parse orderings to account for missing entities

jdkent:
ref poldracklab/fmriprep#1923

I decided to add optional syntax in the yaml file for users to include "None" for their ordering entities.

Alternatives could be:
1. add `None` to all values.
    - pro: do not need to change existing yaml syntax
    - con: an example combination `[('session', None), ('task', None), ('run', None)]` may return unexpected files, and may produce unwanted behavior for other people's use cases.
2. Find all existing combinations of entities somehow for the relevant files.
    - pro: do not need to change existing yaml syntax
    - con: it appears difficult to implement. (my current thought process would be to grab all the possible files as BidsLayoutFiles and extract their entities. Then I would keep/filter the relevant entities and sort them/change them into tuples)

Thoughts?


jdkent:
The second commit is for alternative 2, the code is almost unreadable, but it works for the test cases.

jdkent:
I suppose I do not know if this will be superseded by #440

niworkflows/reports/core.py
```diff
@@ -386,6 +382,43 @@ def generate_report(self):
...
+        if none_member in all_value_combos:
+            all_value_combos.remove(tuple([None for k in orderings]))
+        # see what values exist for each entity
+        unique_values = [{value[idx] for value in all_value_combos}
+                         for idx in range(len(orderings))]
+        # if all values are None for an entity, we do not want to keep that entity
+        keep_idx = [False if (len(val_set) == 1 and None in val_set) or not val_set
+                    else True for val_set in unique_values]
+        # the "kept" entities
+        entities = list(compress(orderings, keep_idx))
```
jdkent: entities should be the same size as value_combos, right now it is just a single list.

niworkflows/reports/core.py
```diff
@@ -386,6 +382,43 @@ def generate_report(self):
         (self.out_dir / self.out_filename).write_text(report_render, encoding='UTF-8')
         return len(self.errors)
 
+    @staticmethod
+    def _process_orderings(orderings, layout):
+        """
+        generate relevant combinations of orderings with observed values
+
```
oesteban:
```suggestion

Arguments
---------
orderings : :obj:`list` of :obj:`list` of :obj:`str`
    Sections prescribing an ordering to select across sessions, acquisitions, runs, etc.
layout : :obj:`bids.layout.BIDSLayout`
    The BIDS layout

```

niworkflows/reports/tests/test_core.py
```diff
@@ -0,0 +1,55 @@
...
+                  subject_id='01', packagename='fmriprep')
+
+
+@pytest.mark.parametrize(
+    "orderings,expected_entities,expected_value_combos",
+    [
+        (['session', 'task', 'run'],
+         ['task', 'run'],
+         [
+            ('faketask', None),
```
oesteban: Now that you are at it, could you add svgs (I'd be fine with a placeholder) so that the layout finds:

1. A `faketask2` with `run` = `None`
2. A `faketaskwithruns` with `run` = {1, 2}

?

Along the same lines, within this PR or in a new one, we should add tests for more possible layouts (e.g., one with sessions comes to mind).

oesteban:
BTW, I'd love to hear the opinion of @josephmje on this one.

niworkflows/reports/core.py
jdkent: nvm, this is correct.

niworkflows/reports/tests/test_core.py
jdkent: good idea, I've added more combinations, and another test that is easier to dynamically create different BIDS datasets.
When you say placeholder you mean like a minimal `svg` file that does not really look like a report that would go into fmriprep?

jdkent:
[here is an example report with the all the functionals](https://www.dropbox.com/s/yo8t8o4p7er95g6/sub-01.html?dl=0) (anats are ignored because they are within the sessions?)

oesteban:
@jdkent is this the right syntax? https://github.com/nipreps/dmriprep/pull/50/files#diff-071866c6808620f9441951ae140c7de4R30

oesteban:
If that is the correct syntax, it doesn't seem to work for dmriprep - https://463-174237413-gh.circle-artifacts.com/0/tmp/THP002/derivatives/dmriprep/sub-THP0005.html (there should be 2 runs but only one is shown).

jdkent:
@oesteban, you actually shouldn't have to change anything about the syntax with this pull request, but regardless, it should have worked, is there a way I can access the artifacts of the run?

jdkent:
got it working for dmriprep - https://16-234180549-gh.circle-artifacts.com/0/tmp/THP002/derivatives/dmriprep/sub-THP0005.html

PR #447, opened 2020-01-19
ENH/TEST: Add all valid entities to fmriprep report specification

### LINKED ISSUE #446: include the other possible entities for ordering the reportlets
[fmriprep.yml](https://github.com/poldracklab/niworkflows/blob/4e9181fd31f3a1b216b5edcf040bd5533c400e9d/niworkflows/reports/fmriprep.yml#L36) should include [all the valid entities](https://bids-specification.readthedocs.io/en/stable/99-appendices/04-entity-table.html) for ordering the reportlets. With #443 merged, including the other entities should not impact any files that do not include those entities.

I believe the ordering should be:
```
session,task,acquisition,ceagent,reconstruction,direction,run,echo(?)
```

for example, the current yml file will not generate reports for both:
```
sub-01_task-rest_acq-mb_bold.nii.gz
sub-01_task-rest_acq-sb_bold.nii.gz
```

jdkent:
fixes #446

### DIFF
--- a/niworkflows/reports/fmriprep.yml
+++ b/niworkflows/reports/fmriprep.yml
@@ -33,7 +33,7 @@ sections:
     caption: Brain extraction of the magnitude image from the fieldmap.
     subtitle: Skull stripped magnitude image
 - name: Functional
-  ordering: session,task,run
+  ordering: session,task,acquisition,ceagent,reconstruction,direction,run,echo
   reportlets:
   - bids: {datatype: func, desc: summary, suffix: bold}
   - bids: {datatype: func, desc: validation, suffix: bold}
--- a/niworkflows/reports/tests/test_core.py
+++ b/niworkflows/reports/tests/test_core.py
@@ -202,7 +202,7 @@ def test_generated_reportlets(bids_sessions, ordering):
     # the last section is the most recently run
     reportlets_num = len(report.sections[-1].reportlets)
     # if ordering does not contain all the relevent entities
-    # then their should be fewer reportlets than expected
+    # then there should be fewer reportlets than expected
     if all(ent in ordering for ent in needed_entities):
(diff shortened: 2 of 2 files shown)


effigies:
This looks reasonable to me. Guess I should have held off on 1.0.4. Do we want this in a bug-fix release of fMRIPrep? If so, can you target `maint/1.0.x`?

jdkent:
I should have realized this is also a change I wanted to make to get all the reports out as expected, but does fmriprep use [reports-spec.yml](https://github.com/poldracklab/fmriprep/blob/63a661b20ec68e1aa08949d5dd6fd17381ea1d52/fmriprep/data/reports-spec.yml) as its reports specification?

effigies:
Ah, no. It looks like fMRIPrep has its own: https://github.com/poldracklab/fmriprep/blob/master/fmriprep/data/reports-spec.yml#L30

Should we just apply this change over there?

jdkent:
I believe that would be the solution to get the behavior I wanted. Is `fmriprep.yml` in niworkflows just for testing purposes?

effigies:
IDK. You'll have to ask @oesteban.

effigies:
Anyway, would you like to open that PR, or should I? You can do it off the `maint/1.5.x` branch, and it will get merged into `master` afterwards.

jdkent:
once poldracklab/fmriprep#1937 is merged, I can open the followup pull request to change reports-spec against the `maint/1.5.x` branch

effigies:
Oh. Want to just add that change to poldracklab/fmriprep#1937? Or keep it separate? I'm good to merge that one (I kind of thought I already had...), so either way.

jdkent:
I'll add it to that pull request (for simplicity)

PR #448, opened 2020-01-20
[FIX] Backport gh-447 to 1.0.x series

jdkent:
ref #447 
original issue #446


effigies:
Closing in favor of https://github.com/poldracklab/fmriprep/pull/1937. Please reopen if we think we need to release a new niworkflows 1.0.x with this change, but my impression is that it will only serve to improve the tests, and that can wait for 1.1.x and #447.

PR #449, opened 2020-01-21
ENH+RF: Spaces object, refine interfaces to handle tuple representation of spaces

mgxd:
upstream PR - poldracklab/fmriprep#1915

### Additions
- `Spaces` class for managing output/internal/standard/volumetric/surface/etc. spaces.

### Changes
- Expand support for `KeySelect` interface beyond just strings.
   - New option `no_hash` was added to avoid hashing `keys` (avoids checking for duplicates, i.e., multiple copies of a `key` can exist)
- Expand `DerativesDataSink.inputs.space` to support tuple representation of spaces `(space, {space_specs})`
  - Outputs now include space specs in final filenames (i.e., a space of `('MNI152', {'res': '2'})` will produce a filename matching `*_space-MNI152_res-2_*`

TODO:
 - [x] update doctests with new behaviors

Technically backwards compatible but we'll probably need to release a new minor version (wdyt @effigies)


effigies:
Yeah, I would go with a new minor version.

niworkflows/interfaces/bids.py
```diff
@@ -240,7 +240,10 @@ class _DerivativesDataSinkInputSpec(DynamicTraitedSpec, BaseInterfaceInputSpec):
     keep_dtype = traits.Bool(False, usedefault=True, desc='keep datatype suffix')
     meta_dict = traits.DictStrAny(desc='an input dictionary containing metadata')
     source_file = File(exists=False, mandatory=True, desc='the input func file')
-    space = Str('', usedefault=True, desc='Label for space field')
+    space = traits.Either(Str, traits.Tuple(Str, traits.Dict),
+                          default='', usedefault=True,
+                          desc='Label for space field. Accepts a template string or a '
+                               'tuple in the form of (template, {template specifications})')
```
effigies: Does it make sense to make `res` and `den` subordinate to `space`? Conceptually, that makes sense, but in terms of generic BIDS entities, maybe it would make more sense to abstract into `entities`, which would allow any additional entities to be specified.

But possibly that's too much of a pain. IDK. We might just need to revisit this whole interface sooner or later.

niworkflows/utils/spaces.py
```diff
@@ -0,0 +1,114 @@
+"""Utilities for tracking and filtering spaces"""
+
+FSAVERAGE_DENSITY = {
+    '642': 'fsaverage3',
+    '2562': 'fsaverage4',
+    '10k': 'fsaverage5',
+    '41k': 'fsaverage6',
+    '164k': 'fsaverage',
+}
+NONSTANDARD_REFERENCES = ['anat', 'T1w', 'run', 'func', 'sbref', 'session', 'fsnative']
```
oesteban: Do we want these two to be under the `Spaces` namespace as class attributes?

niworkflows/utils/spaces.py
```diff
@@ -0,0 +1,114 @@
...
+
+
+class Spaces:
+    """Helper class for managing fMRIPrep spaces."""
```
oesteban: Can we add doctest examples?

niworkflows/utils/spaces.py
```diff
@@ -0,0 +1,114 @@
...
+
+    def __init__(self, output=None):
+        """
+        Tracks and distinguishes internal and output spaces.
+
+        Internal spaces are normalizations required for pipeline execution which
+        can vary based on user arguments.
+        Output spaces are desired user outputs.
+        """
+        self._output = []
```
oesteban: In reality this represents template specs, as opposed to the "internal" - which are actually spaces in BIDS terms. Please note the thin line between template and space.

I guess I don't love the nomenclature of internal/outputs for this new data structure.

niworkflows/interfaces/bids.py
mgxd: Yeah, I guess in a BIDS context `space`, `res`, and `den` are all separate entities, so this can be a little confusing. Though they are closely related in an internal fmriprep aspect (the upstream PR adds support for multiple configurations of the same space).

I'll try playing around with the interface's `allowed_entities` functionality to see if I can get the desired behavior without altering this.

niworkflows/utils/spaces.py
mgxd: I wanted to make the object as lightweight as possible - do you think we should include these as part of the class?

niworkflows/utils/spaces.py
mgxd: definitely open to suggestions here

niworkflows/utils/spaces.py
oesteban: I guess, because these names are not standard (there's only a concept of individual space in BIDS), a class attribute is cleaner to update by applications

niworkflows/interfaces/bids.py
mgxd: actually, i don't think this is possible using `allowed_entities`, since the dictionary containing the space specifications is unable to be passed through workflow connections.

or were you suggesting adding a new input `entities` that would handle extra fields?

oesteban:
Can we close this one in favor of #451?

mgxd:
sure - if i need to rework `KeySelect` + `DerivativesDataSink` i'll build off that one.

PR #451, opened 2020-01-24
ENH: Separate ``Space`` and ``SpatialReferences``

oesteban:
(brought in from mgxd/niworkflows#1)

Okay, after much thinking, I believe this captures the whole set of requirements.

Now there is an ``Space`` object that makes all the validations and sanity checks. There is still room for enhancements:
- ~~We might want to implement a classmethod (something like ``Space.from_string``) that implements the parsing of ``--output-spaces`` arguments (see https://www.attrs.org/en/stable/init.html).~~ DONE
- We might want to check that the specs contain valid keywords against templateflow (at the moment, this is only done for cohort).

Other than that, I think the doctest examples show pretty much all the functionality.

Then, the ``SpatialReferences`` class is the one in charge of sorting out the available spaces. It has three convenience methods to filter out references:
- ``get_std_spaces()`` returns template names (including cohort if necessary) that is of interest to the spatial normalizations that sMRIPrep will run.
- ``get_templates()`` returns full ``Space`` objects with the desired outputs, as set by the user.
- ``get_nonstd_spaces()`` returns nonstandard ``Space``s or their names (depending on the ``only_names`` argument)

Added tests on examples like those in https://github.com/poldracklab/fmriprep/issues/1604#issuecomment-562243214

### DIFF
--- a/niworkflows/utils/spaces.py
+++ b/niworkflows/utils/spaces.py
@@ -490,28 +490,24 @@ def _expand_entities(entities):
 
     Ported from PyBIDS
 
+
+    .. testsetup::
+
+        >>> if PY_VERSION < (3, 6):
+        ...     pytest.skip("This doctest does not work on python <3.6")
+
     Examples
     --------
     >>> entities = {'subject': ['01', '02'], 'session': ['1', '2'], 'task': ['rest', 'finger']}
-    >>> out = _expand_entities(entities)
-    >>> len(out)
-    8
-    >>> {'subject': '01', 'session': '1', 'task': 'rest'} in out
-    True
-    >>> {'subject': '02', 'session': '1', 'task': 'rest'} in out
-    True
-    >>> {'subject': '01', 'session': '2', 'task': 'rest'} in out
-    True
-    >>> {'subject': '02', 'session': '2', 'task': 'rest'} in out
-    True
-    >>> {'subject': '01', 'session': '1', 'task': 'finger'} in out
-    True
-    >>> {'subject': '02', 'session': '1', 'task': 'finger'} in out
-    True
-    >>> {'subject': '01', 'session': '2', 'task': 'finger'} in out
-    True
-    >>> {'subject': '02', 'session': '2', 'task': 'finger'} in out
-    True
+    >>> _expand_entities(entities)
+    [{'subject': '01', 'session': '1', 'task': 'rest'},
+     {'subject': '02', 'session': '1', 'task': 'rest'},
+     {'subject': '01', 'session': '2', 'task': 'rest'},
+     {'subject': '02', 'session': '2', 'task': 'rest'},
+     {'subject': '01', 'session': '1', 'task': 'finger'},
+     {'subject': '02', 'session': '1', 'task': 'finger'},
+     {'subject': '01', 'session': '2', 'task': 'finger'},
+     {'subject': '02', 'session': '2', 'task': 'finger'}]
 
     """
     keys = list(entities.keys())
--- a/niworkflows/utils/spaces.py
+++ b/niworkflows/utils/spaces.py
@@ -264,6 +264,9 @@ class Space:
         [Space(name='MNIPediatricAsym', spec={'cohort': '1'}),
          Space(name='MNIPediatricAsym', spec={'cohort': '2'})]
 
+        >>> Space.from_string("fsaverage:den-10k:den-164k")
+        [Space(name='fsaverage', spec={'den': '10k'}),
+         Space(name='fsaverage', spec={'den': '164k'})]
 
         >>> Space.from_string("MNIPediatricAsym:cohort-5:cohort-6:res-2")
         [Space(name='MNIPediatricAsym', spec={'cohort': '5', 'res': '2'}),
@@ -414,9 +417,11 @@ class SpatialReferences:
                 return True
         return False
 
-    def __repr__(self):
+    def __str__(self):
         """Representation of this object."""
-        return 'Imaging spaces:%s' % '\n  - '.join([] + [str(s) for s in self.spaces])
+        if not self._spaces:
+            return 'Spatial References: <none>.'
+        return 'Spatial References:%s' % '\n  - '.join([] + [str(s) for s in self.spaces])
 
     @property
     def spaces(self):
--- a/niworkflows/utils/spaces.py
+++ b/niworkflows/utils/spaces.py
@@ -136,40 +136,6 @@ class Space:
     >>> sp1 == sp2
     True
 
-    >>> Space.from_string("MNI152NLin2009cAsym")
-    [Space(name='MNI152NLin2009cAsym', spec={})]
-
-    >>> # Bad name
-    >>> Space.from_string("shouldraise")
-    Traceback (most recent call last):
-      ...
-    ValueError: space identifier "shouldraise" is invalid.
-    ...
-
-    >>> # Missing cohort
-    >>> Space.from_string("MNIPediatricAsym")
-    Traceback (most recent call last):
-      ...
-    ValueError: standard space "MNIPediatricAsym" is not fully defined.
-    ...
-
-    >>> Space.from_string("MNIPediatricAsym:cohort-1")
-    [Space(name='MNIPediatricAsym', spec={'cohort': '1'})]
-
-    >>> Space.from_string("MNIPediatricAsym:cohort-1:cohort-2")
-    [Space(name='MNIPediatricAsym', spec={'cohort': '1'}),
-     Space(name='MNIPediatricAsym', spec={'cohort': '2'})]
-
-    >>> Space.from_string("MNIPediatricAsym:cohort-5:cohort-6:res-2")
-    [Space(name='MNIPediatricAsym', spec={'cohort': '5', 'res': '2'}),
-     Space(name='MNIPediatricAsym', spec={'cohort': '6', 'res': '2'})]
-
-    >>> Space.from_string("MNIPediatricAsym:cohort-5:cohort-6:res-2:res-iso1.6mm")
-    [Space(name='MNIPediatricAsym', spec={'cohort': '5', 'res': '2'}),
-     Space(name='MNIPediatricAsym', spec={'cohort': '5', 'res': 'iso1.6mm'}),
-     Space(name='MNIPediatricAsym', spec={'cohort': '6', 'res': '2'}),
-     Space(name='MNIPediatricAsym', spec={'cohort': '6', 'res': 'iso1.6mm'})]
-
     """
 
     _standard_spaces = tuple(_tfapi.templates())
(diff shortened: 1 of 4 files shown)


niworkflows/utils/spaces.py
```diff
@@ -0,0 +1,418 @@
...
+
+    name: str = attr.ib(default=None)
+    """Unique name designating this space."""
+    cohort: str = attr.ib(init=False, default=None)
+    """An attribute to accomodate cohorts from TemplateFlow."""
+    spec: typing.Dict = attr.ib(factory=dict)
+    """The dictionary of specs."""
+    standard: bool = attr.ib(default=False, repr=False)
+    """Whether this space is standard or not."""
+    dim: bool = attr.ib(default=3, repr=False)
```
oesteban:
```suggestion
    dim: int = attr.ib(default=3, repr=False)
```

niworkflows/utils/spaces.py
```diff
@@ -0,0 +1,418 @@
...
+    """The dictionary of specs."""
+    standard: bool = attr.ib(default=False, repr=False)
+    """Whether this space is standard or not."""
+    dim: int = attr.ib(default=3, repr=False)
+    """Dimensionality of the sampling manifold."""
+
+    def __attrs_post_init__(self):
+        """Extract cohort out of spec."""
+        if self.spec is None:
+            object.__setattr__(self, "spec", {})
```
oesteban: This trick is necessary when the class is frozen.

niworkflows/utils/spaces.py
```diff
@@ -0,0 +1,418 @@
...
+        raise ValueError('space "%s" already in spaces.' % str(value))
+
+    def insert(self, index, value, error=True):
+        """Concatenate one more space."""
+        if value not in self:
+            self.spaces.insert(index, self.check_space(value))
+        elif error is True:
+            raise ValueError('space "%s" already in spaces.' % str(value))
+
+    def get_std_spaces(self, dim=(2, 3)):
```
mgxd: having the `only_names` keyword like `get_nonstd_spaces` would be useful

niworkflows/utils/spaces.py
```diff
@@ -0,0 +1,418 @@
...
+        for modifier in _args[1:]:
+            mitems = modifier.split('-', 1)
+            spec[mitems[0]].append(len(mitems) == 1 or mitems[1])
+
+        allspecs = _expand_entities(spec)
+
+        return [cls(_args[0], s) for s in allspecs]
+
+
+class SpatialReferences:
```
mgxd: playing off `Space.from_string()`, a class method could be implemented to process the `opt.output_spaces` input

niworkflows/utils/spaces.py
```diff
@@ -0,0 +1,418 @@
+"""Utilities for tracking and filtering spaces."""
+import attr
+import typing
+from bids.layout.writing import _expand_entities
```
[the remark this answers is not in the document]
oesteban: That makes sense. Will port the code

niworkflows/utils/spaces.py
```diff
@@ -0,0 +1,418 @@
...
+            _cohorts = ', '.join(['"cohort-%s"' % c for c in _cohorts])
+            raise ValueError(
+                'standard space "%s" is not fully defined.\n'
+                'Set a valid cohort selector from: %s.' % (self.name, _cohorts))
+
+    @name.validator
+    def _check_name(self, attribute, value):
+        if value.startswith('fsaverage'):
+            return
+        valid = list(self._standard_spaces) + NONSTANDARD_REFERENCES
```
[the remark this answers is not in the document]
oesteban: They just need to comply with TemplateFlow's organization and place their template in the templateflow home folder. I think that is the easiest way.

niworkflows/utils/spaces.py
oesteban: Working on this on the sMRIPrep side.

niworkflows/utils/spaces.py
oesteban: this method intendedly provides only names. It could have some use below, in ``get_templates`` (although missing out the spec kind of defeats the purpose). Actually, I would think more of removing the ``only_names`` from ``get_nonstd_spaces``.

mgxd: I'm a still a bit confused how this plans on separating user requested spaces from internally added ones.

the way most outputnodes are currently set up, a list of all spaces are returned. how can we then filter out only the requested spaces with this class?

oesteban: Let's break it down into smaller pieces.

- Volumes (spatial normalization):
  In this case, we just need to feed this iterable with the right information https://github.com/poldracklab/smriprep/blob/master/smriprep/workflows/norm.py#L141.
  I might be blindsided at this moment, but I believe having template ID + cohort should suffice. Since it is an iterable, Nipype will take care of all the synchronization (and that is how it actually works currently).

- Surfaces:
  I will need to revise things, but my impression is that sMRIPrep does not need anything about surfaces, other than whether recon-all is run or not.
  In fMRIPrep proper, we need surface output spaces to decide whether we generate derivatives or not (therefore, no concept of "internal" space is necessary). With this way of thinking, we will be able to run recon-all but abstain from generating derivatives if the user does not explicitly request them.

As I said, I'll need to go through https://github.com/poldracklab/smriprep/pull/159 to make sure this all makes sense and works out.

mgxd: > I might be blindsided at this moment, but I believe having template ID + cohort should suffice. Since it is an iterable, Nipype will take care of all the synchronization (and that is how it actually works currently).

We'll have to pass in specs as well, to fully differentiate the spaces. IIRC, duplicate iterables will overwrite working directories.

niworkflows/utils/spaces.py
```diff
@@ -0,0 +1,497 @@
...
+    >>> sp = SpatialReferences([
+    ...     'func',
+    ...     'fsnative',
+    ...     'MNI152NLin2009cAsym',
+    ...     'anat',
+    ...     'fsaverage5',
+    ...     'fsaverage6',
+    ...     ('MNIPediatricAsym', {'cohort': '2'}),
+    ...     ('MNI152NLin2009cAsym', {'res': 2}),
+    ...     ('MNI152NLin2009cAsym', {'res': 1}),
```
effigies: Are res `int`s? Can they be `str`?

niworkflows/utils/spaces.py
```diff
@@ -0,0 +1,497 @@
...
+
+    @spec.validator
+    def _check_spec(self, attribute, value):
+        if value is not None and not isinstance(value, dict):
+            raise TypeError(
+                "invalid space specification: %s." % str(value))
+
+    @classmethod
+    def from_string(cls, value):
+        """Create a new Space from string."""
```
effigies: This seems to return a list of `Space`s.

niworkflows/utils/spaces.py
effigies: Why are we freezing it? It seems if we need to hack around, then we're probably not using the API as intended.

niworkflows/utils/spaces.py
mgxd: i believe it can be both, similar to density

oesteban: Since the BIDS PR is not merge, these are still `int`s. We will make them `str` when that is official in BIDS-Derivatives.

niworkflows/utils/spaces.py
oesteban: Once you create a Space object, it should remain untouched. The trick is suggested within the attrs documentation for this particular situation.

niworkflows/utils/spaces.py
oesteban: That is because it needs to generate lists of spaces to generate the 4 spaces of ``MNIPediatricAsym:cohort-1:cohort-2:res-1:res-2``.

Your comment is more about that it should be mentioned in the docstring? - I should be able to fix that :)

oesteban: Does this https://1088-74174878-gh.circle-artifacts.com/0/root/project/docs/_build/no_version_html/html/api/niworkflows.utils.spaces.html#niworkflows.utils.spaces.Space.from_string address your point?

effigies: Yup, looks good.

niworkflows/utils/spaces.py
oesteban: https://github.com/mgxd/smriprep/pull/1/files#diff-816eb5ff17b071ecf371cf48577d5a3cR246

``Space.fullname`` is doing the magic - unique template ID (+cohort) are passed.

mgxd:
looks good to me - i'll merge later today if no objections

niworkflows/utils/spaces.py
```diff
@@ -0,0 +1,519 @@
...
+     'MNIPediatricAsym:cohort-1']
+
+    >>> sp.insert(0, ('MNIPediatricAsym', {'cohort': '3'}))
+    Traceback (most recent call last):
+      ...
+    ValueError: space ...
+
+    """
+
+    __slots__ = ('_spaces')
```
effigies: This is currently interpreted as `{'_', 's', 'p', 'a', 'c', 'e'}`. See https://1104-74174878-gh.circle-artifacts.com/0/root/project/docs/_build/no_version_html/html/api/niworkflows.utils.spaces.html#niworkflows.utils.spaces.SpatialReferences.a

```suggestion
    __slots__ = ('_spaces',)
```

mgxd: good catch

oesteban:
Good catch!

niworkflows/utils/spaces.py
```diff
@@ -0,0 +1,519 @@
...
+    >>> {'subject': '02', 'session': '2', 'task': 'rest'} in out
+    True
+    >>> {'subject': '01', 'session': '1', 'task': 'finger'} in out
+    True
+    >>> {'subject': '02', 'session': '1', 'task': 'finger'} in out
+    True
+    >>> {'subject': '01', 'session': '2', 'task': 'finger'} in out
+    True
+    >>> {'subject': '02', 'session': '2', 'task': 'finger'} in out
+    True
```
effigies: This is deterministic in Python 3.6+, too, right? Can we do the `PY_VERSION` trick again?

oesteban: done!

oesteban:
Really happy with how this is looking - thank you both very much for your time :)

mgxd:
working on a few fixes here.

oesteban:
Should I hold off fixing that doctest?

mgxd:
opened oesteban/niworkflows#1 with some fixes.

mainly to allow proper parsing of spaces passed through the commandline

niworkflows/utils/spaces.py
```diff
@@ -414,9 +417,11 @@ def __contains__(self, item):
                 return True
         return False
 
-    def __repr__(self):
+    def __str__(self):
         """Representation of this object."""
-        return 'Imaging spaces:%s' % '\n  - '.join([] + [str(s) for s in self.spaces])
+        if not self._spaces:
```
mgxd:
```suggestion
        if not self.spaces:
```

PR #452, opened 2020-01-29
FIX: Parsing of floats in ``ResampleImageBySpacing``

oesteban:


effigies:
What's the context for this fix? Do we want this in 1.0.x, or is 1.1.x soon enough?

effigies:
Either way, I'm :+1: to merge.

oesteban:
This is a bug @eilidhmacnicol found while working with rats - hence, sub-mm resolution.

I don't think we need to backport it, we will use the latest release available.

PR #454, opened 2020-01-29
ENH: Add helper function for cleaning a directory

mgxd:
instead of deleting and remaking (https://github.com/nipy/nipype/blob/792988f19c427c6f631df4d38a5942063f4b8d46/nipype/utils/filemanip.py#L768) (lol docstring), keep and clean directory

### DIFF
--- a/niworkflows/utils/misc.py
+++ b/niworkflows/utils/misc.py
@@ -3,10 +3,12 @@
 """
 Miscellaneous utilities
 """
+import os
+import shutil
 
 
 __all__ = ['get_template_specs', 'fix_multi_T1w_source_name', 'add_suffix', 'read_crashfile',
-           'splitext', '_copy_any']
+           'splitext', '_copy_any', 'clean_directory']
 
 
 def get_template_specs(in_template, template_spec=None, default_resolution=1):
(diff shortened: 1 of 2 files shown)


effigies:
This is going to be subject to race conditions. Instead of returning `False`, you may want to `continue`, and at the end `return not list(os.scandir(path)).`

mgxd:
sorry - i'm not seeing the race condition. can you explain?

effigies:
Two threads or processes calling this function could race. `rmtree` and (I think) `unlink` will raise an error if the thing they're trying to remove doesn't exist.

effigies:
Should we just ratchet to 3.6+ and stop worrying about `PathLike`?

mgxd:
i think that finally did the trick (i can't wait til we drop 3.5)

re: the race condition - I think that may be beyond the intended use of this (planned to port over to fmriprep to make reruns easier). what about adding a line in the docstring explicitly warning this is not thread/multiprocess safe?

effigies:
> what about adding a line in the docstring explicitly warning this is not thread/multiprocess safe?

Sure.

PR #455, opened 2020-02-03
enh: standard space querying

mgxd:
while progressing in https://github.com/poldracklab/fmriprep/pull/1955, we find ourselves extending the queries that i believe the `get_` helper methods were trying to solve. This PR improves the capabilities of `Spaces` querying for standard spaces.

I've only updated one get method for feedback, and if this looks alright, we can extend this to the others.

### DIFF
--- a/niworkflows/utils/spaces.py
+++ b/niworkflows/utils/spaces.py
@@ -512,15 +512,16 @@ class SpatialReferences:
             raise ValueError('space "%s" already in spaces.' % str(value))
 
     def get_std_spaces(self, only_names=True, dim=(2, 3)):
-        """Return standard spaces.
+        """
+        Return standard spaces.
 
         Parameters
         ----------
-        only_names : :obj:bool, optional
-            If `True` (default), return all unique :obj:Spaces names. If `False`,
-            return all `Space` entities.
-        dim : :obj:tuple, optional
-            Desired dimensions of the standard spaces (default is (2, 3))
+        only_names : :obj:`bool`, optional
+            If ``True`` (default), return all unique :class:`Space` names. If ``False``,
+            return all :class:`Space` entities.
+        dim : :obj:`tuple`, optional
+            Desired dimensions of the standard spaces (default is ``(2, 3)``)
 
         Examples
         --------
--- a/niworkflows/utils/spaces.py
+++ b/niworkflows/utils/spaces.py
@@ -341,8 +341,8 @@ class SpatialReferences:
     ValueError: Spaces have not ...
 
     >>> sp.checkpoint()
-    >>> sp.cached
-    (Space(name='func', spec={}),
+    >>> sp.cached.spaces
+    [Space(name='func', spec={}),
      Space(name='fsnative', spec={}),
      Space(name='MNI152NLin2009cAsym', spec={}),
      Space(name='anat', spec={}),
@@ -350,7 +350,10 @@ class SpatialReferences:
      Space(name='fsaverage', spec={'den': '41k'}),
      Space(name='MNIPediatricAsym', spec={'cohort': '2'}),
      Space(name='MNI152NLin2009cAsym', spec={'res': 2}),
-     Space(name='MNI152NLin2009cAsym', spec={'res': 1}))
+     Space(name='MNI152NLin2009cAsym', spec={'res': 1})]
+
+    >>> sp.cached.get_fs_spaces()
+    ['fsnative', 'fsaverage5', 'fsaverage6']
 
     >>> sp.add(('MNIPediatricAsym', {'cohort': '2'}))
     >>> sp.get_std_spaces(dim=(3,))
@@ -482,7 +485,7 @@ class SpatialReferences:
         """Cache and freeze current spaces to separate attribute"""
         if self._cached is not None:
             raise ValueError("Spaces have already been cached")
-        self._cached = tuple(self.spaces)
+        self._cached = self.__class__(self.spaces)
 
     @spaces.setter
     def spaces(self, value):
@@ -508,7 +511,7 @@ class SpatialReferences:
         elif error is True:
             raise ValueError('space "%s" already in spaces.' % str(value))
 
-    def get_std_spaces(self, only_names=True, dim=(2, 3), cached=False):
+    def get_std_spaces(self, only_names=True, dim=(2, 3)):
         """Return standard spaces.
 
         Parameters
@@ -518,8 +521,6 @@ class SpatialReferences:
             return all `Space` entities.
         dim : :obj:tuple, optional
             Desired dimensions of the standard spaces (default is (2, 3))
-        cached : :obj:bool, optional
-            If `True`, query cached :obj:Spaces. If `False` (default), query all :obj:Spaces.
 
         Examples
         --------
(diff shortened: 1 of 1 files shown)


effigies:
Is this necessary before an fMRIPrep release?

mgxd:
this aims to simplify filtering of spaces in the above PR, so ideally before we release 20.0.0

niworkflows/utils/spaces.py
```diff
@@ -508,17 +508,55 @@ def insert(self, index, value, error=True):
...
-        std_spaces = []
-        for s in self._spaces:
-            if s.standard and s.dim in dim and s.fullname not in names:
-                names.append(s.fullname)
-                std_spaces.append(s)
-        if only_names:
-            return names
-        return std_spaces
+    def get_std_spaces(self, only_names=True, dim=(2, 3), cached=False):
+        """Return standard spaces.
```
oesteban:
```suggestion
        """
        Return standard spaces.
```

niworkflows/utils/spaces.py
```diff
@@ -508,17 +508,55 @@ def insert(self, index, value, error=True):
...
+
+        Parameters
+        ----------
+        only_names : :obj:bool, optional
```
oesteban:
```suggestion
        only_names : :obj:`bool`, optional
```

niworkflows/utils/spaces.py
```diff
@@ -508,17 +508,55 @@ def insert(self, index, value, error=True):
...
+            If `True` (default), return all unique :obj:Spaces names. If `False`,
```
oesteban:
```suggestion
            If ``True`` (default), return all unique :class:`Space` names. If ``False``,
```

niworkflows/utils/spaces.py
```diff
@@ -508,17 +508,55 @@ def insert(self, index, value, error=True):
...
+            return all `Space` entities.
```
oesteban:
```suggestion
            return all :class:`Space` entities.
```

niworkflows/utils/spaces.py
```diff
@@ -508,17 +508,55 @@ def insert(self, index, value, error=True):
...
+        dim : :obj:tuple, optional
```
oesteban:
```suggestion
        dim : :obj:`tuple`, optional
```

niworkflows/utils/spaces.py
```diff
@@ -508,17 +508,55 @@ def insert(self, index, value, error=True):
...
+            Desired dimensions of the standard spaces (default is (2, 3))
```
oesteban:
```suggestion
            Desired dimensions of the standard spaces (default is ``(2, 3)``)
```

niworkflows/utils/spaces.py
```diff
@@ -508,17 +508,55 @@ def insert(self, index, value, error=True):
...
+        cached : :obj:bool, optional
```
oesteban:
```suggestion
       cached : :obj:`bool`, optional
```

niworkflows/utils/spaces.py
```diff
@@ -508,17 +508,55 @@ def insert(self, index, value, error=True):
...
+            If `True`, query cached :obj:Spaces. If `False` (default), query all :obj:Spaces.
```
oesteban:
```suggestion
            If ``True``, query only cached :obj:`Space` items. If ``False`` (default), query all available.
```

niworkflows/utils/spaces.py
```diff
@@ -508,17 +508,55 @@ def insert(self, index, value, error=True):
...
+
+        >>> spaces.checkpoint()
+        >>> spaces.add(('MNI152NLin6Asym', {'res': '2'}))
+        >>> spaces.get_std_spaces()
+        ['MNI152NLin6Asym', 'fsaverage']
+
+        >>> spaces.get_std_spaces(only_names=False)
+        [Space(name='MNI152NLin6Asym', spec={}),
+         Space(name='fsaverage', spec={'den': '10k'}),
+         Space(name='MNI152NLin6Asym', spec={'res': '2'})]
```
oesteban: I think I prefer the former behavior, which would drop the second instance of ``MNI152NLin6Asym``

mgxd: I understand dropping a duplicate template when looking at `only_names`, but shouldn't we list different specifications of the same templates when looking at all components of the `Space`s? Otherwise, this method is a bit deceiving IMO

oesteban: I guess the problem roots from the name we chose for the `Space` data type. Because name and spec pairs are not just spaces, they are spatial references. A space does not have any notion of the sampling grid. Actually, my original implementation did not have `only_names` flag.

niworkflows/utils/spaces.py
```diff
@@ -508,17 +508,55 @@ def insert(self, index, value, error=True):
...
-        names = []
-        std_spaces = []
-        for s in self._spaces:
-            if s.standard and s.dim in dim and s.fullname not in names:
-                names.append(s.fullname)
-                std_spaces.append(s)
-        if only_names:
-            return names
-        return std_spaces
+    def get_std_spaces(self, only_names=True, dim=(2, 3), cached=False):
```
oesteban: I've been reflecting on this a bit more.

Why not returning a new `SpatialReferences` object when calling ``spaces.cache``? That way, all these methods would operate on the cached version without additional code.

niworkflows/utils/spaces.py
mgxd: ok - three ways to solve this:

1) allow `only_names` to view all `Space` objects, regardless of duplicate templates. This will lighten the load of other queries (ex https://github.com/poldracklab/fmriprep/pull/1955/files#diff-928c5651a2bdc224e2a84acff159a0bcR194)
2) revert this to only list the first `Space`'s specs and send the other(s) into limbo.
3) remove `only_names` and just return template names

anything but 2 is fine with me.

oesteban: I'd go with 3 (remove `only_names`), returning unique names (in the order of first occurrence).

oesteban: For clarity/consistency, we might want to change https://github.com/poldracklab/niworkflows/pull/455/files#diff-fb99c1c1fd9dceadc84eb083f28aede6R564 to:
```Python
def get_nonstd_references(self, only_names=True, dim=(2, 3)):
```
or just
```Python
def get_nonstandard(self, only_names=True, dim=(2, 3)):
```

mgxd: sure, but then let's just drop only_names altogether

oesteban: And also, see suggestion on line 559

niworkflows/utils/spaces.py
```diff
@@ -509,16 +512,49 @@ def insert(self, index, value, error=True):
...
+        for s in self.spaces:
+            if s.standard and s.dim in dim:
+                if only_names:
+                    if s.fullname not in out:
+                        out.append(s.fullname)
+                    continue
+                out.append(s)
+        return out
 
     def get_templates(self, dim=(2, 3)):
```
oesteban:
```suggestion
    def get_standard(self, dim=(2, 3)):
```

mgxd:
let's get this as is - we can touch up a few things in a new PR

PR #457, opened 2020-02-05
ENH: Revision of ``spaces`` module for consistency

oesteban:


niworkflows/utils/spaces.py
```diff
@@ -457,115 +453,128 @@ def __str__(self):
...
+                and s.dim in dim
+            ):
+                out.append(s.fullname)
         return out
 
-    def get_templates(self, dim=(2, 3)):
-        """Return output spaces."""
-        return [s for s in self._spaces
-                if s.standard and hasspec(s.spec, ('res', 'den')) and s.dim in dim]
+    def get_standard(self, full_spec=False, dim=(2, 3)):
```
mgxd: we should consider changing this to plural, as it's supposed to be an iterable anyways
```suggestion
    def get_standard(self, full_spec=False, dims=(2, 3)):
```

niworkflows/utils/spaces.py
```diff
@@ -457,115 +453,128 @@ def __str__(self):
...
+
+        """
+        if not full_spec:
+            return [s for s in self.references if s.standard and s.dim in dim]
+
+        return [s for s in self.references
+                if s.standard and s.dim in dim and hasspec('res', s.spec) or hasspec('den', s.spec)]
 
-    def get_nonstd_spaces(self, only_names=True, dim=(2, 3)):
+    def get_nonstandard(self, dim=(2, 3)):
```
mgxd: this should match `get_standard` and include an option to return specs

niworkflows/utils/spaces.py
```diff
@@ -586,24 +595,28 @@ def get_fs_spaces(self):
...
+class OutputReferencesAction(argparse.Action):
     """Parse spatial references."""
 
     def __call__(self, parser, namespace, values, option_string=None):
         """Execute parser."""
         spaces = getattr(namespace, self.dest) or SpatialReferences()
         for val in values:
-            for sp in Space.from_string(val):
+            val = val.rstrip(":")
+            # Should we support some sort of explicit "default" resolution?
```
mgxd: I don't think so, how would this handle volumes/surfaces? Doesn't an empty dict convey the notion of "default" everything?

niworkflows/utils/spaces.py
oesteban: I think this is just a selector... what I want to include is a "tuplefy" method, so that it also accepts ints - e.g., ``dim=3``

niworkflows/utils/spaces.py
oesteban: Okay, but please re-read the utilization of ``full_spec`` above because I feel it does not do what you seem to think...

niworkflows/utils/spaces.py
```diff
@@ -616,6 +629,25 @@ def hasspec(value, specs):
     return False
 
 
+def format_reference(in_tuple):
+    """
+    Format a spatial reference given as a tuple.
+
+    Examples
+    --------
+    >>> format_space(('MNI152Lin', {'res': 1}))
```
mgxd:
```suggestion
    >>> format_reference(('MNI152Lin', {'res': 1}))
```

niworkflows/utils/spaces.py
```diff
@@ -616,6 +629,25 @@ def hasspec(value, specs):
...
+    'MNI152Lin_res-1'
+    >>> format_space(('MNIPediatricAsym:cohort-2', {'res': 2}))
```
mgxd:
```suggestion
    >>> format_reference(('MNIPediatricAsym:cohort-2', {'res': 2}))
```

niworkflows/utils/spaces.py
oesteban: I am not confident this is the place for setting the default, but this would be necessary if we want (and I think we do) "default" and not-set different semantics.

From the user perspective, users should not be allowed to provide under-specified resampling targets. Therefore, setting a default would never harm.

From the point you added the checkpointing feature, these two different meanings of not-set and "default" are not as badly needed. But this could be helpful to only allow underspecified references for internal spaces.

niworkflows/utils/spaces.py
mgxd: sure, but we also want to have support for multiple resolutions of nstd spaces, as referenced in https://github.com/poldracklab/fmriprep/issues/1604#issuecomment-561443156

niworkflows/utils/spaces.py
mgxd: yeah, should probably go in `Reference.__iadd__`.

if you see it making life easier, i'm for it.

niworkflows/utils/spaces.py
mgxd: i think `dims` is more intuitive as it can be more than one.

niworkflows/utils/spaces.py
```diff
@@ -410,7 +406,7 @@ def check_space(space):
                 space = (None, )
 
             space = space[0]
-        return Space(space, spec)
+        return Reference(space, spec)
 
     def __init__(self, spaces=None):
```
mgxd: while updating the fmriprep docs, i realized there isn't a way to initiate a cached `SpatialReference`.  WDYT about adding an additional keyword `checkpoint` to call `self.checkpoint()` if spaces were set?

oesteban: Oh, yes. I also had thought of this. Will do!

niworkflows/utils/spaces.py
oesteban: Having `Reference.__iadd__` set a default will actually block the possibility of inserting spaces without spec (and we want to be able to do that)

niworkflows/utils/spaces.py
oesteban: `full_spec=True` just drops references without full (i.e., having res or den) spec. Nonstandard spaces will generally not have any of those

mgxd: if we plan on implementing the linked example, they would

```
--output-spaces \
    MNI152NLIn2009cAsym:res-1:res-2:res-iso1.2mm:res-bold \
    T1w:res-native:res-bold \
    func:res-native:res-iso2.0mm \
    session:res-iso2.0mm
```

niworkflows/utils/spaces.py
mgxd: yeah, it would probably be in `check_specs`, I'll leave this for you to decide

niworkflows/utils/spaces.py
oesteban: would you change it in the `Reference` object as well?

niworkflows/utils/spaces.py
oesteban: Let's not worry about this for now. I'll leave the comment there for when we come back to it.

niworkflows/utils/spaces.py
oesteban: Okay, added the full spec option

PR #460, opened 2020-02-13
ENH: Downgrade log level for superfluous scans

mgxd:
ref poldracklab/fmriprep#1980


mgxd:
no need to build for this one

PR #461, opened 2020-02-13
ENH: Revise SpatialReference caching for ease of use, accessibility

mgxd:
to add support for poldracklab/fmriprep#1979

This adds a `checkpoint()` call in `OutputReferencesAction` when the flag attached to the action is used with no arguments.

### DIFF
--- a/niworkflows/utils/spaces.py
+++ b/niworkflows/utils/spaces.py
@@ -331,12 +331,16 @@ class SpatialReferences:
      Reference(space='MNI152NLin2009cAsym', spec={'res': 2}),
      Reference(space='MNI152NLin2009cAsym', spec={'res': 1})]
 
+    >>> sp.is_cached()
+    False
     >>> sp.cached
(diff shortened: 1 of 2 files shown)


niworkflows/utils/spaces.py
```diff
@@ -476,13 +490,16 @@ def references(self):
     @property
     def cached(self):
```
effigies: If we're okay with fiddling with the API still, I would suggest changing this to `cache`. Intuitively, `obj.cached` feels like it should be a bool, while `obj.cache` would return the cache.

mgxd: I don't feel too strongly about either way, only minor pain I see is changing all examples / upstream calls. If we decide to go that route, let's do it on a separate PR

effigies: Sure.

mgxd:
I'll merge this in the morning and cut a new release

PR #463, opened 2020-02-19
ENH: Add a pure-python ApplyMask interface, based on NiBabel

oesteban:


niworkflows/interfaces/nibabel.py
```diff
@@ -13,6 +13,45 @@
...
+        if img.dataobj.shape[:3] != msk.shape:
+            raise ValueError("Image and mask sizes do not match.")
+
+        if not np.allclose(img.affine, msknii.affine):
+            raise ValueError("Image and mask affines are not similar enough.")
+
+        if img.dataobj.ndim == msk.ndim + 1:
+            msk = msk[..., np.newaxis]
+
+        masked = img.__class__(img.dataobj * msk, None, img.header)
```
oesteban: This is probably the potentially contended bit of this new interface. I'll leave it here and won't merge until it gets the blessings from @effigies, whenever he's back and can have a look.

effigies: Seems fine. `numpy.ndarray` defines `__rmul__`, so this will act like `np.asarray(img.dataobj) * mask`.

PR #467, opened 2020-02-26
RF: Replace os operations with pathlib, add metadata indentation

mgxd:


niworkflows/interfaces/cifti.py
```diff
@@ -290,8 +284,7 @@ def _get_cifti_variant(surface, volume, density=None):
         grayords = {'32k': '91k', '59k': '170k'}[density]
         out_json['grayordinates'] = grayords
 
-    with open(out_metadata, 'w') as fp:
-        json.dump(out_json, fp)
+    out_metadata.write_text(json.dumps(out_json, indent=1))
```
effigies: I think 2 is more typical, and these aren't deeply nested.

```suggestion
    out_metadata.write_text(json.dumps(out_json, indent=2))
```

PR #468, opened 2020-02-27
MAINT: Transfer to nipreps organization

oesteban:


effigies:
We'll probably need this in `maint/1.1.x` to ensure things work properly for the next release.

Do we have a PyPI auth token in Circle?

oesteban:
> Do we have a PyPI auth token in Circle?

Yes, I just set it up.

effigies:
Merged a couple things to get a test on `master`. Might cause a failure you haven't covered.

PR #477, opened 2020-03-16
BF: Ensure BOLD and label orientations are equal

mgxd:


mgxd:
We were assuming the BOLD image and label image had a shared orientation

https://github.com/nipreps/niworkflows/blob/bfffaf8ea8e6fe1615be4a24089a5d750bd6e024/niworkflows/interfaces/cifti.py#L364-L370

However, this wasn't the case, as the BOLD image is in RAS while the label image is in LAS. This adds a check to ensure the orientations match up, and if they don't, convert to LAS (HCP Pipelines default, although https://www.nitrc.org/forum/attachment.php?attachid=342&group_id=454&forum_id=1955 recommends RAS space).

niworkflows/interfaces/cifti.py
```diff
@@ -422,3 +428,40 @@ def _create_cifti_image(bold_file, label_file, bold_surfs, annotation_files, tr,
...
+    """
+    Coerce an image to a target orientation.
+
+    .. note::
+        Only RAS -> LAS conversion is currently supported
+
+    Parameters
+    ----------
+    img : `spatialimage`
+    orient_img : tuple
```
oesteban:
```suggestion
    orient_img : :obj:`tuple`
```

niworkflows/interfaces/cifti.py
```diff
@@ -422,3 +428,40 @@ def _create_cifti_image(bold_file, label_file, bold_surfs, annotation_files, tr,
...
+        axis direction codes
+    orient_target : tuple
```
oesteban:
```suggestion
    orient_target : :obj:`tuple`
```

niworkflows/interfaces/cifti.py
```diff
@@ -318,6 +318,12 @@ def _create_cifti_image(bold_file, label_file, bold_surfs, annotation_files, tr,
         warnings.warn("Resampling bold volume to match label dimensions")
         bold_img = resample_to_img(bold_img, label_img)
 
+    bold_orient = nb.aff2axcodes(bold_img.affine)
+    label_orient = nb.aff2axcodes(label_img.affine)
+    if bold_orient != label_orient:
+        bold_img = _reorient_image(bold_img, bold_orient, label_orient)
+    assert nb.aff2axcodes(bold_img) == label_orient == tuple('LAS')
```
oesteban: I would do the checks inside `_reorient_image`, so that you don't need to pass the orientations - just `bold_img` and `label_img`.

niworkflows/interfaces/cifti.py
```diff
@@ -422,3 +428,40 @@ def _create_cifti_image(bold_file, label_file, bold_surfs, annotation_files, tr,
...
+def _reorient_image(img, orient_img, orient_target):
+    """
+    Coerce an image to a target orientation.
+
+    .. note::
+        Only RAS -> LAS conversion is currently supported
+
+    Parameters
+    ----------
+    img : `spatialimage`
```
oesteban: I'd be nice to figure out the right reference for this intersphinx source.
```suggestion
    img : :obj:`spatialimage`
```

mgxd:
any lingering concerns?

I'd like to get this out as a 1.11 despite it changing the outputs, since it's more of a bug fix imo (cc @effigies)

effigies:
Seems reasonable. In the long-term a fully-featured reorientation interface would be good to publish in nipype, but no need to go that far in this context.

PR #479, opened 2020-03-18
[FIX] add ceagent to figures.json

jdkent:
addresses poldracklab/fmriprep#2033

### DIFF
--- a/niworkflows/reports/core.py
+++ b/niworkflows/reports/core.py
@@ -234,7 +234,7 @@ class Report(object):
     >>> robj.generate_report()
     0
     >>> len((testdir / 'out' / 'fmriprep' / 'sub-01.html').read_text())
-    36450
+    36540
 
     .. testcleanup::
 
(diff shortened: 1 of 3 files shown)


jdkent:
[dir](https://github.com/nipreps/niworkflows/blob/ba0d68520ed8ca31ea4f33177c147b3e99743185/niworkflows/reports/figures.json#L32) should also be [direction](https://github.com/bids-standard/pybids/blob/6e11b396aa25c03439e8fb0a6f3caa375456d492/bids/layout/config/bids.json#L32)

oesteban:
> [dir](https://github.com/nipreps/niworkflows/blob/ba0d68520ed8ca31ea4f33177c147b3e99743185/niworkflows/reports/figures.json#L32) should also be [direction](https://github.com/bids-standard/pybids/blob/6e11b396aa25c03439e8fb0a6f3caa375456d492/bids/layout/config/bids.json#L32)

Do you want to address that in this PR?

PR #482, opened 2020-03-18
[FIX] Add CE agent to output figure filename templates

jdkent:
supercedes #479 
addresses poldracklab/fmriprep#2033

### DIFF
--- a/niworkflows/reports/figures.json
+++ b/niworkflows/reports/figures.json
@@ -21,7 +21,7 @@
             "pattern": "[_/\\\\]acq-([a-zA-Z0-9]+)"
         },
         {
-            "name": "ce",
+            "name": "ceagent",
             "pattern": "[_/\\\\]ce-([a-zA-Z0-9]+)"
         },
         {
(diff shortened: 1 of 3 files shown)


niworkflows/reports/figures.json
```diff
@@ -21,15 +21,15 @@
             "pattern": "[_/\\\\]acq-([a-zA-Z0-9]+)"
         },
         {
-            "name": "ce",
+            "name": "ceagent",
```
effigies: Does this have any API effect?

jdkent: No, I do not believe so. This does not impact the recently added --bids-filter, and makes `figures.json` naming consistent with [`fmriprep.yml`](https://github.com/nipreps/niworkflows/blob/bfffaf8ea8e6fe1615be4a24089a5d750bd6e024/niworkflows/reports/fmriprep.yml#L36).

effigies: Okay. @oesteban Is this consistent with your understanding as well?

effigies:
Actually, since #479 isn't merged, let's just make this the main PR, and we'll merge into `master` after the next release.

niworkflows/reports/figures.json
oesteban: Yup, I also think this should not affect the API

effigies:
I've tested with a synthetic dataset where I switched from `run-*` to `ce-*`, and fmriprep 20.0.4 + this PR works as expected.

Given @bbfrederick's comments, I'm going to go ahead and merge.

Thanks!

bbfrederick:
Works for me.  I'll wait until the dust settles, sync up with the patched release version of niworkflows, and then put in a PR for the BIDS_NAME thing.  AFAIK it doesn't actually cause any problems at the moment, but it is wrong and it might bite somebody later if they decide to use it, so it's not currently urgent.

PR #483, opened 2020-03-19
[FIX] update naming patterns in figures.json

jdkent:
two items that I thought should be changed:
- contrast -> ceagent in the anatomical `default_path_patterns`
- remove _bold from echo pattern, (I'm not sure what the original intention of the pattern was, but the double backslash does not [appear to match a reasonable string](https://regex101.com/r/UQROec/1), and echo is not required to be the last key-value pair in the [`default_path_patterns`](https://github.com/nipreps/niworkflows/blob/bfffaf8ea8e6fe1615be4a24089a5d750bd6e024/niworkflows/reports/figures.json#L116))

### DIFF
--- a/niworkflows/reports/figures.json
+++ b/niworkflows/reports/figures.json
@@ -47,7 +47,7 @@
         },
         {
             "name": "echo",
-            "pattern": "[_/\\\\]echo-([0-9]+)\\_bold."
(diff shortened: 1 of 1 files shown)


effigies:
I think it was probably supposed to be `_bold\\.`, but I agree that it doesn't make sense in this context.

PR #484, opened 2020-03-19
ENH: Ensure subcortical CIFTI is in LAS orientation

mgxd:
Go one step beyond #477 and ensures volumetric output in LAS orientation.

### DIFF
--- a/niworkflows/interfaces/cifti.py
+++ b/niworkflows/interfaces/cifti.py
@@ -318,7 +318,9 @@ def _create_cifti_image(bold_file, label_file, bold_surfs, annotation_files, tr,
         warnings.warn("Resampling bold volume to match label dimensions")
         bold_img = resample_to_img(bold_img, label_img)
 
-    bold_img = _reorient_image(bold_img, target_img=label_img)
+    # ensure images match HCP orientation (LAS)
+    bold_img = _reorient_image(bold_img, orientation='LAS')
+    label_img = _reorient_image(label_img, orientation='LAS')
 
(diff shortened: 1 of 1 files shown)


niworkflows/interfaces/cifti.py
```diff
@@ -318,7 +318,9 @@ def _create_cifti_image(bold_file, label_file, bold_surfs, annotation_files, tr,
         warnings.warn("Resampling bold volume to match label dimensions")
         bold_img = resample_to_img(bold_img, label_img)
 
-    bold_img = _reorient_image(bold_img, target_img=label_img)
+    # ensure images match HCP orientation (LAS)
+    bold_img = _reorient_image(bold_img, orientation='LAS')
+    label_img = _reorient_image(label_img, orientation='LAS')
```
effigies: Is this going to be the case for all label files? Should we have a table mapping space to orientation?

mgxd: yes, since the final output should always be in LAS we want to ensure both BOLD / labels are as well prior to setting the IJK voxels

effigies: I guess my reading of Tim's response was that LAS was an accident of the orientation of a specific template. Are we guaranteed that all templates HCP ever uses will be in LAS?

mgxd: no, we aren't - but that seems excessive for us to worry about. I'd rather mirror the outputs generated from HCP Pipelines, which happen to include the volume in LAS orientation

effigies:
Ok. We can make this more complicated when they do. Merge when you like.

Issue #476, opened 2020-03-09
EPI masking: minimize run-to-run difference using a soft mask


oesteban:
Add a soft mask to `MNI152NLin2009cAsym` to replace the hard mask here: https://github.com/nipreps/niworkflows/blob/76f831f1f8fbf6c277fdf92535639bbf11d84527/niworkflows/func/util.py#L291-L292

Having a probabilistic mask that can be binarized after resampling will certainly reduce the differences making our reference run-to-run unreliable.

oesteban:
https://github.com/templateflow/tpl-MNI152NLin2009cAsym/commit/d6753e2ea2d09a59ee541a8822a8400b8a4243ec

PR #485, opened 2020-03-21
ENH: Numerical stability of EPI brain-masks using probabilistic prior

oesteban:
Replaces the template's binary mask with a probabilistic map.

The workflow for EPI brain-extraction utilizes an atlas-based approach in its initialization. Before this PR, the template's mask is binary. The intention is to minimize numerical instability of edge voxels due to the resampling of the binary mask (i.e., when _moving the atlas' brain mask on to the subject space_). With this PR, the binarization occurs later and the risk of having a voxel whose value varies between runs (due to the minimal but existing variability introduced by the ants registration process) is much lower.

See #476

### DIFF
--- a/niworkflows/func/util.py
+++ b/niworkflows/func/util.py
@@ -29,6 +29,7 @@ DEFAULT_MEMORY_MIN_GB = 0.01
 def init_bold_reference_wf(
     omp_nthreads,
     bold_file=None,
+    brainmask_thresh=0.5,
     pre_mask=False,
     name="bold_reference_wf",
     gen_report=False,
@@ -50,10 +51,16 @@ def init_bold_reference_wf(
 
     Parameters
     ----------
-    bold_file : str
-        BOLD series NIfTI file
     omp_nthreads : int
         Maximum number of threads an individual process may use
+    bold_file : str
+        BOLD series NIfTI file
+    brainmask_thresh: :obj:`float`
+        Lower threshold for the probabilistic brainmask to obtain
+        the final binary mask (default: 0.5).
+    pre_mask : bool
+        Indicates whether the ``pre_mask`` input will be set (and thus, step 1
+        should be skipped).
     name : str
         Name of workflow (default: ``bold_reference_wf``)
     gen_report : bool
@@ -133,7 +140,9 @@ using a custom methodology of *fMRIPrep*.
         EstimateReferenceImage(), name="gen_ref", mem_gb=1
     )  # OE: 128x128x128x50 * 64 / 8 ~ 900MB.
     enhance_and_skullstrip_bold_wf = init_enhance_and_skullstrip_bold_wf(
-        omp_nthreads=omp_nthreads, pre_mask=pre_mask
+        brainmask_thresh=brainmask_thresh,
+        omp_nthreads=omp_nthreads,
+        pre_mask=pre_mask,
     )
 
     calc_dummy_scans = pe.Node(
@@ -188,7 +197,10 @@ using a custom methodology of *fMRIPrep*.
 
 
 def init_enhance_and_skullstrip_bold_wf(
-    name="enhance_and_skullstrip_bold_wf", pre_mask=False, omp_nthreads=1
+    brainmask_thresh=0.5,
+    name="enhance_and_skullstrip_bold_wf",
+    omp_nthreads=1,
+    pre_mask=False,
 ):
     """
     Enhance and run brain extraction on a BOLD EPI image.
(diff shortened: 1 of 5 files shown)


oesteban:
Waiting on #488 so that I can run some validation tests on TACC.

niworkflows/func/util.py
```diff
@@ -316,10 +317,15 @@ def init_enhance_and_skullstrip_bold_wf(
...
-            ApplyTransforms(interpolation='MultiLabel', float=True, input_image=str(brain_mask)),
+        map_brainmask = pe.Node(ApplyTransforms(
+            interpolation='BSpline',
+            float=True,
+            # Use the higher resolution and probseg for numerical stability in rounding
+            input_image=str(get_template(
+                'MNI152NLin2009cAsym', resolution=1, desc='brain', suffix='probseg'))),
             name='map_brainmask'
         )
+        binarize_mask = pe.Node(Binarize(thresh_low=0.75), name='binarize_mask')
```
effigies: I would say the default interpretation of binarizing probability is to threshold on >50%.

```suggestion
        binarize_mask = pe.Node(Binarize(thresh_low=0.5), name='binarize_mask')
```

Is there some extenuating situation here?

niworkflows/func/util.py
```diff
@@ -316,10 +317,15 @@ def init_enhance_and_skullstrip_bold_wf(
...
             n_procs=omp_nthreads)
         norm.inputs.fixed_image = str(bold_template)
-        map_brainmask = pe.Node(
-            ApplyTransforms(interpolation='MultiLabel', float=True, input_image=str(brain_mask)),
+        map_brainmask = pe.Node(ApplyTransforms(
+            interpolation='BSpline',
+            float=True,
+            # Use the higher resolution and probseg for numerical stability in rounding
+            input_image=str(get_template(
+                'MNI152NLin2009cAsym', resolution=1, desc='brain', suffix='probseg'))),
```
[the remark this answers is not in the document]
oesteban: correct

oesteban:
I haven't tested this yet on TACC on a larger scale but it should be ready to go. One little detail I might have overseen from @effigies comment is that the `desc-brain_probseg` is not valid, it should be `label-brain_probseg`. I probably did not understand his comment at the moment. Do we want to address that detail?

mgxd:
yup - would you have time to get that in today? I'm planning on cutting a new rc today

oesteban:
okay, doing it right now

niworkflows/func/util.py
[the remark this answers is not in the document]
oesteban: was this comment referring to the following: to stick with BIDS closer, it should not be a `desc` but a `label` here (i.e., `label-brain_probseg`)?

(this has been addressed)

oesteban:
I'm finding it impossible to test this separately before merging. Should we just go ahead or keep waiting until I can do this (or someone else)?

effigies:
Do we have an fMRIPrep branch pinned against this to at least get one CI run we can look at artifacts?

oesteban:
I think we can go ahead with this one - we probably will need to revise the brain mask down the line, but that's more of a TemplateFlow's problem. WDYT?

oesteban:
@effigies @mgxd any issues with merging this one?

Bottom line is: unless we want to spend more time in generating a really probabilistic brainmask of MNI152NLin2009cAsym, this will not improve (we can move things around a little playing with the threshold, that doesn't give a lot of search space).

I think this is sufficient as is to address the problem, and setting the threshold at .85 or a bit higher seems to yield results very very close to those of the original binary mask.

If at some point we embark in generating probabilistic masks, I would basically extract the brain mask for many OpenNeuro datasets with OASIS (default sMRIPrep), keep using sMRIPrep to calculate the normalization to MNI152NLin2009cAsym, and finally average the resulting masks after projection back to MNI152NLin2009cAsym. This probabilistic mask would be interesting for two reasons:

1. The potential improvement here (BOLD masking)
2. Potential for T1w masking (e.g., https://github.com/poldracklab/smriprep/wiki/Brain-Extraction-from-T1w---very-hard-cases) and for switching the default template of brain extraction to the MNI. This would give the whole workflow a better consistency, and initialization from brain extraction could be reused to shave off some minutes of runtime.

(point 2 would be of an extremely low priority and I don't even have any expectation that it will be ever done).

oesteban:
Ref: templateflow/tpl-MNI152NLin2009cAsym#10

niworkflows/func/util.py
```diff
@@ -238,13 +250,16 @@ def init_enhance_and_skullstrip_bold_wf(
 
     Parameters
     ----------
+    brainmask_thresh: :obj:`float`
+        Lower threshold for the probabilistic brainmask to obtain
+        the final binary mask (default: 0.5).
```
mgxd, line 255: > setting the threshold at .85 or a bit higher seems to yield results very very close to those of the original binary mask

shouldn't this be .85 by default then?

oesteban: Sure, no problem

effigies:
On ds210, the mask is calculated on a pre-masked brain, and really doesn't match the boundary. It's hard to tell whether it would fit better if the underlying brain were unmasked.

https://14340-53175327-gh.circle-artifacts.com/0/tmp/ds210/derivatives/fmriprep/sub-02.html#datatype-figures_desc-summary_echo-None_run-1_suffix-bold_task-cuedSGT

oesteban:
> On ds210, the mask is calculated on a pre-masked brain, and really doesn't match the boundary. It's hard to tell whether it would fit better if the underlying brain were unmasked.

Maybe not the best issue tittle I have chosen: poldracklab/fmriprep#2162

Because of that difficulty, I have opened the reportlet before and after this PR and checked visually the contour. Except for minimal deviations, with 0.85 the contour is almost always matched.

PR #487, opened 2020-03-21
MAINT: Drop Python 3.5

### LINKED ISSUE #486: MAINT: remove PY35 workarounds

oesteban:
Addresses #486.

Some mentions to python 3.5 still remain:

```
$ find . -name "*.py" -exec grep -H "35" {} \;
./niworkflows/reports/core.py:        # PY35: Sorted config dict for consistent behavior
./niworkflows/reports/core.py:                # PY35: Coerce to str to pacify os.* functions that don't take Paths until 3.6
./niworkflows/utils/misc.py:    for f in os.scandir(path):  # switch to context manager when py35 dropped
```

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -353,7 +353,7 @@ jobs:
     working_directory: /tmp/src/niworkflows
     steps:
       - checkout
-      - run: pyenv local 3.5.2
+      - run: pyenv local 3.7.0
       - run:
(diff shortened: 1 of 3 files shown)


mgxd:
do you want to refactor those references within this PR?

oesteban:
I was considering to leave them there for now, wdyt?

PR #488, opened 2020-03-23
ENH: Transfer brain mask script from *fMRIPrep*

oesteban:
Adds a scaffold to allow quick implementation of command-lines wrapping the internal workflows.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -47,6 +47,10 @@ test_requires =
 packages = find:
 include_package_data = True
 
+[options.entry_points]
+console_scripts =
+    niworkflows=niworkflows.cli.run:main
+
 [options.package_data]
 niworkflows =
     data/*.txt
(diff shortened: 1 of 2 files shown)


oesteban:
Hi @eilidhmacnicol please have a look into this draft - we want something along these lines for nirodents :).

setup.cfg
```diff
@@ -47,6 +47,10 @@ test_requires =
 packages = find:
 include_package_data = True
 
+[options.entry_points]
+console_scripts =
+    niworkflows=niworkflows.cli.run:main
```
effigies: Seems weird to give a specific tool such primacy of place in the namespace, even if we don't have others yet.

```suggestion
    niworkflows-boldref=niworkflows.cli.boldref:main
```

effigies:
Not sure accepting that suggestion will help without renaming the CLI file...

PR #489, opened 2020-03-24
ENH: Add nibabel-based split and merge interfaces

dPys:
per suggestion [here](https://github.com/nipreps/dmriprep/pull/77)

### DIFF
--- a/niworkflows/interfaces/nibabel.py
+++ b/niworkflows/interfaces/nibabel.py
@@ -96,7 +96,7 @@ class _FourToThreeInputSpec(BaseInterfaceInputSpec):
 
 
 class _FourToThreeOutputSpec(TraitedSpec):
-    out_files = OutputMultiPath(File(exists=True),
+    out_files = OutputMultiObject(File(exists=True),
                                      desc='output list of 3d images')
 
 
--- a/niworkflows/interfaces/nibabel.py
+++ b/niworkflows/interfaces/nibabel.py
@@ -113,7 +113,7 @@ class FourToThree(SimpleInterface):
             raise RuntimeError('Input image (%s) is not 4D.' % filenii)
 
         files_3d = nb.four_to_three(filenii)
-        out_files = []
+        self._results['out_files'] = []
         for i, file_3d in enumerate(files_3d):
             out_file = fname_presuffix(in_file, suffix=f"_idx-{i:03}")
             file_3d.to_filename(out_file)
--- a/niworkflows/interfaces/nibabel.py
+++ b/niworkflows/interfaces/nibabel.py
@@ -140,7 +140,7 @@ class Save3Dto4D(SimpleInterface):
 
     def _run_interface(self, runtime):
         nii_list = []
-        for i, f in enumerate(self.inputs.in_files):
+        for f in self.inputs.in_files:
             filenii = nb.load(f)
             filenii = nb.squeeze_image(filenii)
             if len(filenii.shape) != 3:
--- a/niworkflows/interfaces/nibabel.py
+++ b/niworkflows/interfaces/nibabel.py
@@ -7,7 +7,7 @@ from nipype import logging
 from nipype.utils.filemanip import fname_presuffix
 from nipype.interfaces.base import (
     traits, TraitedSpec, BaseInterfaceInputSpec, File,
-    SimpleInterface
+    SimpleInterface, OutputMultiPath, InputMultiPath
 )
 
 IFLOGGER = logging.getLogger('nipype.interface')
@@ -88,4 +88,68 @@ class Binarize(SimpleInterface):
                                 img.header)
         maskimg.to_filename(self._results['out_mask'])
 
+        return runtime
+
+class _Save4Dto3DInputSpec(BaseInterfaceInputSpec):
+    in_file = File(exists=True, mandatory=True, desc='input 4d image')
+
+
+class _Save4Dto3DOutputSpec(TraitedSpec):
+    out_files = OutputMultiPath(File(exists=True),
+                                     desc='output list of 3d images')
+
+
+class Save4Dto3D(SimpleInterface):
+    """Split a 4D dataset along the last dimension
+    into multiple 3D volumes."""
+
+    input_spec = _Save4Dto3DInputSpec
+    output_spec = _Save4Dto3DOutputSpec
+
+    def _run_interface(self, runtime):
+        filenii = nb.load(self.inputs.in_file)
+        if len(filenii.shape) != 4:
+            raise RuntimeError('Input image (%s) is not 4D.' % filenii)
+
+        files_3d = nb.four_to_three(filenii)
+        out_files = []
+        for i, file_3d in enumerate(files_3d):
+            out_file = fname_presuffix(in_file, suffix="_tmp_{}".format(i))
+            file_3d.to_filename(out_file)
+            out_files.append(out_file)
+
+        self._results['out_files'] = out_files
+        return runtime
+
+
+class _Save3Dto4DInputSpec(BaseInterfaceInputSpec):
+    in_files = InputMultiPath(File(exists=True, mandatory=True,
+                                   desc='input list of 3d images'))
+
+
+class _Save3Dto4DOutputSpec(TraitedSpec):
+    out_file = File(exists=True, desc='output list of 3d images')
+
+
+class Save3Dto4D(SimpleInterface):
+    """Merge a list of 3D volumes along the last dimension into a single
+    4D image."""
+
+    input_spec = _Save4Dto3DInputSpec
+    output_spec = _Save4Dto3DOutputSpec
+
+    def _run_interface(self, runtime):
+        nii_list = []
+        for i, f in enumerate(self.inputs.in_files):
+            filenii = nb.load(f)
+            filenii = nb.squeeze_image(filenii)
+            if len(filenii.shape) != 3:
+                raise RuntimeError('Input image (%s) is not 3D.' % f)
+            else:
+                nii_list.append(filenii)
+        img_4d = nb.funcs.concat_images(nii_list)
+        out_file = fname_presuffix(self.inputs.in_files[0], suffix="_merged")
+        img_4d.to_filename(out_file)
+
+        self._results['out_file'] = out_file
         return runtime
--- a/niworkflows/interfaces/nibabel.py
+++ b/niworkflows/interfaces/nibabel.py
@@ -100,7 +100,7 @@ class _FourToThreeOutputSpec(TraitedSpec):
                                      desc='output list of 3d images')
 
 
-class FourToThree(SimpleInterface):
+class SplitSeries(SimpleInterface):
     """Split a 4D dataset along the last dimension
     into a series of 3D volumes."""
 
--- a/niworkflows/interfaces/nibabel.py
+++ b/niworkflows/interfaces/nibabel.py
@@ -1,6 +1,7 @@
 # emacs: -*- mode: python; py-indent-offset: 4; indent-tabs-mode: nil -*-
 # vi: set ft=python sts=4 ts=4 sw=4 et:
 """Nibabel-based interfaces."""
+from pathlib import Path
 import numpy as np
 import nibabel as nb
 from nipype import logging
@@ -98,12 +99,11 @@ class _FourToThreeInputSpec(BaseInterfaceInputSpec):
 
 class _FourToThreeOutputSpec(TraitedSpec):
     out_files = OutputMultiObject(File(exists=True),
-                                     desc='output list of 3d images')
+                                  desc='output list of 3d images')
 
 
 class SplitSeries(SimpleInterface):
-    """Split a 4D dataset along the last dimension
-    into a series of 3D volumes."""
+    """Split a 4D dataset along the last dimension into a series of 3D volumes."""
 
     input_spec = _FourToThreeInputSpec
     output_spec = _FourToThreeOutputSpec
@@ -119,7 +119,8 @@ class SplitSeries(SimpleInterface):
                 self._results['out_files'] = out_file
                 filenii.to_filename(out_file)
                 return runtime
-            raise RuntimeError(f"Input image image is {ndim}D.")
+            raise RuntimeError(
+                f"Input image image is {ndim}D ({'x'.join(['%d' % s for s in filenii.shape])}).")
 
         files_3d = nb.four_to_three(filenii)
         self._results['out_files'] = []
@@ -137,7 +138,8 @@ class SplitSeries(SimpleInterface):
 class _MergeSeriesInputSpec(BaseInterfaceInputSpec):
     in_files = InputMultiObject(File(exists=True, mandatory=True,
                                      desc='input list of 3d images'))
-    allow_4D = traits.Bool(True, usedefault=True, desc='whether 4D images are allowed to be concatenated')
+    allow_4D = traits.Bool(True, usedefault=True,
+                           desc='whether 4D images are allowed to be concatenated')
 
 
 class _MergeSeriesOutputSpec(TraitedSpec):
@@ -153,15 +155,18 @@ class MergeSeries(SimpleInterface):
     def _run_interface(self, runtime):
         nii_list = []
         for f in self.inputs.in_files:
-            filenii = nb.load(f)
-            filenii = nb.squeeze_image(filenii)
-            if filenii.dataobj.ndim == 3:
+            filenii = nb.squeeze_image(nb.load(f))
+            ndim = filenii.dataobj.ndim
+            if ndim == 3:
                 nii_list.append(filenii)
-            elif self.inputs.allow_4D and filenii.dataobj.ndim == 4:
+                continue
+            elif self.inputs.allow_4D and ndim == 4:
                 nii_list += nb.four_to_three(filenii)
+                continue
             else:
                 raise ValueError("Input image has an incorrect number of dimensions"
-                                 f" ({filenii.dataobj.ndim}).")
+                                 f" ({ndim}).")
+
         img_4d = nb.concat_images(nii_list)
         out_file = fname_presuffix(self.inputs.in_files[0], suffix="_merged")
         img_4d.to_filename(out_file)
--- a/niworkflows/interfaces/nibabel.py
+++ b/niworkflows/interfaces/nibabel.py
@@ -99,22 +99,22 @@ class Binarize(SimpleInterface):
         return runtime
 
 
-class _FourToThreeInputSpec(BaseInterfaceInputSpec):
+class _SplitSeriesInputSpec(BaseInterfaceInputSpec):
     in_file = File(exists=True, mandatory=True, desc="input 4d image")
     allow_3D = traits.Bool(
         False, usedefault=True, desc="do not fail if a 3D volume is passed in"
     )
 
 
-class _FourToThreeOutputSpec(TraitedSpec):
+class _SplitSeriesOutputSpec(TraitedSpec):
     out_files = OutputMultiObject(File(exists=True), desc="output list of 3d images")
 
 
 class SplitSeries(SimpleInterface):
     """Split a 4D dataset along the last dimension into a series of 3D volumes."""
 
-    input_spec = _FourToThreeInputSpec
-    output_spec = _FourToThreeOutputSpec
+    input_spec = _SplitSeriesInputSpec
+    output_spec = _SplitSeriesOutputSpec
 
     def _run_interface(self, runtime):
         filenii = nb.squeeze_image(nb.load(self.inputs.in_file))
--- a/niworkflows/interfaces/nibabel.py
+++ b/niworkflows/interfaces/nibabel.py
@@ -99,7 +99,7 @@ class _Save4Dto3DOutputSpec(TraitedSpec):
                                      desc='output list of 3d images')
 
 
-class Save4Dto3D(SimpleInterface):
+class FourToThree(SimpleInterface):
     """Split a 4D dataset along the last dimension
     into multiple 3D volumes."""
 
--- a/niworkflows/interfaces/nibabel.py
+++ b/niworkflows/interfaces/nibabel.py
@@ -115,7 +115,7 @@ class FourToThree(SimpleInterface):
         files_3d = nb.four_to_three(filenii)
         out_files = []
         for i, file_3d in enumerate(files_3d):
-            out_file = fname_presuffix(in_file, suffix="_tmp_{}".format(i))
+            out_file = fname_presuffix(in_file, suffix=f"_idx-{i:03}")
             file_3d.to_filename(out_file)
             out_files.append(out_file)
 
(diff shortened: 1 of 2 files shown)


niworkflows/interfaces/nibabel.py
```diff
@@ -89,3 +89,67 @@ def _run_interface(self, runtime):
...
+class _Save4Dto3DInputSpec(BaseInterfaceInputSpec):
+    in_file = File(exists=True, mandatory=True, desc='input 4d image')
+
+
+class _Save4Dto3DOutputSpec(TraitedSpec):
+    out_files = OutputMultiPath(File(exists=True),
+                                     desc='output list of 3d images')
+
+
+class Save4Dto3D(SimpleInterface):
```
effigies: I would say either call it by the nibabel name:

```suggestion
class FourToThree(SimpleInterface):
```

Or something more standard:

```suggestion
class Split(SimpleInterface):
```

`Save` sounds like a datasink.

niworkflows/interfaces/nibabel.py
```diff
@@ -89,3 +89,67 @@ def _run_interface(self, runtime):
...
+class _Save3Dto4DInputSpec(BaseInterfaceInputSpec):
+    in_files = InputMultiPath(File(exists=True, mandatory=True,
+                                   desc='input list of 3d images'))
+
+
+class _Save3Dto4DOutputSpec(TraitedSpec):
+    out_file = File(exists=True, desc='output list of 3d images')
+
+
+class Save3Dto4D(SimpleInterface):
```
effigies: Here something like "Concatenate" or "Merge" would make sense to me. "ConcatImages" would also be fine, if trying to follow nibabel naming.

niworkflows/interfaces/nibabel.py
```diff
@@ -89,3 +89,67 @@ def _run_interface(self, runtime):
...
+    """Merge a list of 3D volumes along the last dimension into a single
+    4D image."""
+
+    input_spec = _Save4Dto3DInputSpec
+    output_spec = _Save4Dto3DOutputSpec
+
+    def _run_interface(self, runtime):
+        nii_list = []
+        for i, f in enumerate(self.inputs.in_files):
```
effigies: You don't seem to use the `i` for anything.

```suggestion
        for f in self.inputs.in_files:
```

niworkflows/interfaces/nibabel.py
```diff
@@ -7,7 +7,7 @@
 from nipype.utils.filemanip import fname_presuffix
 from nipype.interfaces.base import (
     traits, TraitedSpec, BaseInterfaceInputSpec, File,
-    SimpleInterface
+    SimpleInterface, OutputMultiPath, InputMultiPath
```
effigies: Deprecated names.

```suggestion
    SimpleInterface, OutputMultiObject, InputMultiObject
```

niworkflows/interfaces/nibabel.py
```diff
@@ -89,3 +89,67 @@ def _run_interface(self, runtime):
         maskimg.to_filename(self._results['out_mask'])
 
         return runtime
+
```
effigies:
```suggestion


```

niworkflows/interfaces/nibabel.py
dPys: I like `ConcatImages`

niworkflows/interfaces/nibabel.py
```diff
@@ -89,3 +89,68 @@ def _run_interface(self, runtime):
...
+
+class _FourToThreeInputSpec(BaseInterfaceInputSpec):
+    in_file = File(exists=True, mandatory=True, desc='input 4d image')
+
+
+class _FourToThreeOutputSpec(TraitedSpec):
+    out_files = OutputMultiPath(File(exists=True),
```
effigies:
```suggestion
    out_files = OutputMultiObject(File(exists=True),
```

niworkflows/interfaces/nibabel.py
```diff
@@ -89,3 +89,68 @@ def _run_interface(self, runtime):
...
+            out_file = fname_presuffix(in_file, suffix="_tmp_{}".format(i))
+            file_3d.to_filename(out_file)
+            out_files.append(out_file)
+
+        self._results['out_files'] = out_files
+        return runtime
+
+
+class _ConcatImagesInputSpec(BaseInterfaceInputSpec):
+    in_files = InputMultiPath(File(exists=True, mandatory=True,
```
effigies:
```suggestion
    in_files = InputMultiObject(File(exists=True, mandatory=True,
```

niworkflows/interfaces/nibabel.py
```diff
@@ -89,3 +89,68 @@ def _run_interface(self, runtime):
...
+                                   desc='input list of 3d images'))
+
+
+class _ConcatImagesOutputSpec(TraitedSpec):
+    out_file = File(exists=True, desc='output list of 3d images')
+
+
+class ConcatImages(SimpleInterface):
```
effigies: Thinking about this, I'm not sure `ConcatImages` is a good name, since you're doing some preprocessing before passing to `concat_images`. `concat_images` is perfectly happy to concatenate along any axis and along that axis there is no requirement that the images have the same dimension.

dPys: hmm maybe something like `4dMergeImages` would be better?

dPys: or `FourDMerge`, `MergeImages`, `MergeThreeToFour` ...

effigies: `Merge` seems fine. Or `MergeImages` or `MergeSeries` if that seems non-specific.

dPys: `MergeSeries`!

niworkflows/interfaces/nibabel.py
```diff
@@ -89,3 +89,68 @@ def _run_interface(self, runtime):
...
+
+    def _run_interface(self, runtime):
+        filenii = nb.load(self.inputs.in_file)
+        if len(filenii.shape) != 4:
+            raise RuntimeError('Input image (%s) is not 4D.' % filenii)
+
+        files_3d = nb.four_to_three(filenii)
+        out_files = []
+        for i, file_3d in enumerate(files_3d):
+            out_file = fname_presuffix(in_file, suffix="_tmp_{}".format(i))
```
oesteban, line 118: We are Python +3.6, let's use f-strings and make it a bit bids-y:
```suggestion
            out_file = fname_presuffix(in_file, suffix=f"_idx-{i:03}")
```

niworkflows/interfaces/nibabel.py
```diff
@@ -89,3 +89,68 @@ def _run_interface(self, runtime):
...
+    input_spec = _FourToThreeInputSpec
+    output_spec = _FourToThreeOutputSpec
+
+    def _run_interface(self, runtime):
+        filenii = nb.load(self.inputs.in_file)
+        if len(filenii.shape) != 4:
+            raise RuntimeError('Input image (%s) is not 4D.' % filenii)
+
+        files_3d = nb.four_to_three(filenii)
+        out_files = []
```
oesteban, line 116:
```suggestion
        self._results['out_files'] = []
```

niworkflows/interfaces/nibabel.py
```diff
@@ -89,3 +89,68 @@ def _run_interface(self, runtime):
...
+        for i, file_3d in enumerate(files_3d):
+            out_file = fname_presuffix(in_file, suffix="_tmp_{}".format(i))
+            file_3d.to_filename(out_file)
+            out_files.append(out_file)
+
+        self._results['out_files'] = out_files
```
oesteban, line 122:
```suggestion
            self._results['out_files'].append(out_file)

```

niworkflows/interfaces/nibabel.py
```diff
@@ -89,3 +89,68 @@ def _run_interface(self, runtime):
...
+        return runtime
+
+
+class _MergeSeriesInputSpec(BaseInterfaceInputSpec):
+    in_files = InputMultiObject(File(exists=True, mandatory=True,
+                                   desc='input list of 3d images'))
```
oesteban, line 128:
```suggestion
    in_files = InputMultiObject(File(exists=True, mandatory=True,
                                     desc='input list of 3d images'))
```

niworkflows/interfaces/nibabel.py
```diff
@@ -89,3 +89,68 @@ def _run_interface(self, runtime):
...
+
+
+class _MergeSeriesOutputSpec(TraitedSpec):
+    out_file = File(exists=True, desc='output 4d image')
+
+
+class MergeSeries(SimpleInterface):
+    """Merge a series of 3D volumes along the last dimension into a single
+    4D image."""
```
oesteban, line 137:
```suggestion
    """Merge a series of 3D volumes along the last dimension into a single 4D image."""
```

niworkflows/interfaces/nibabel.py
```diff
@@ -89,3 +89,68 @@ def _run_interface(self, runtime):
...
+class _FourToThreeInputSpec(BaseInterfaceInputSpec):
+    in_file = File(exists=True, mandatory=True, desc='input 4d image')
+
+
+class _FourToThreeOutputSpec(TraitedSpec):
+    out_files = OutputMultiObject(File(exists=True),
+                                     desc='output list of 3d images')
+
+
+class FourToThree(SimpleInterface):
```
oesteban, line 103: I would stick with other existing interfaces conventions (i.e., ``fsl.Split``) or with the convention of ``MergeSeries`` below:
```suggestion
class SplitSeries(SimpleInterface):
```

niworkflows/interfaces/nibabel.py
```diff
@@ -89,3 +89,68 @@ def _run_interface(self, runtime):
...
+    """Split a 4D dataset along the last dimension
+    into a series of 3D volumes."""
+
+    input_spec = _FourToThreeInputSpec
+    output_spec = _FourToThreeOutputSpec
+
+    def _run_interface(self, runtime):
+        filenii = nb.load(self.inputs.in_file)
```
oesteban, line 111:
```suggestion
        filenii = nb.squeeze_image(nb.load(self.inputs.in_file))
```
I would like to hear from @effigies about this one

niworkflows/interfaces/nibabel.py
```diff
@@ -89,3 +89,68 @@ def _run_interface(self, runtime):
...
+        if len(filenii.shape) != 4:
+            raise RuntimeError('Input image (%s) is not 4D.' % filenii)
```
oesteban, line 113: Perhaps just return if it is already 3D - only fail otherwise?

niworkflows/interfaces/nibabel.py
```diff
@@ -89,3 +89,68 @@ def _run_interface(self, runtime):
...
+
+    def _run_interface(self, runtime):
+        nii_list = []
+        for f in self.inputs.in_files:
+            filenii = nb.load(f)
+            filenii = nb.squeeze_image(filenii)
+            if len(filenii.shape) != 3:
+                raise RuntimeError('Input image (%s) is not 3D.' % f)
+            else:
+                nii_list.append(filenii)
```
oesteban, line 150:
```suggestion
            if filenii.dataobj.ndim == 3:
                nii_list.append(filenii)
            elif filenii.dataobj.ndim == 4:
                nii_list += nb.four_to_three(filenii)
            else:
                raise ValueError("Input image has an incorrect number of dimensions"
                                 f" ({filenii.dataobj.ndim}).")
```

niworkflows/interfaces/nibabel.py
```diff
@@ -89,3 +89,68 @@ def _run_interface(self, runtime):
...
+        img_4d = nb.funcs.concat_images(nii_list)
```
oesteban, line 151:
```suggestion
        img_4d = nb.concat_images(nii_list)
```
is this right, @effigies?

niworkflows/interfaces/nibabel.py
dPys: sounds good

niworkflows/interfaces/nibabel.py
dPys: I support this change, but does it change the affine as well as the shape? i.e. is there ever a case where this could corrupt the image?

niworkflows/interfaces/nibabel.py
dPys: What about a case where the image should be 4D, but a 3D image is supplied by mistake? if we went with the `return`, maybe we could still retain a user warning?

niworkflows/interfaces/nibabel.py
dPys: I wonder whether here, too, there should be a user warning accompanying a rogue 4D image when the typical use-case for this interface probably assumes a list of 3D-only inputs (or 4D-only if we wish to generalize it further)?

niworkflows/interfaces/nibabel.py
dPys: both seem to work here, so whatever @effigies prefers?

niworkflows/interfaces/nibabel.py
oesteban: Nope, this won't have any effect on the affine.

niworkflows/interfaces/nibabel.py
```diff
@@ -89,3 +89,66 @@ def _run_interface(self, runtime):
         maskimg.to_filename(self._results['out_mask'])
 
         return runtime
+
+
+class _FourToThreeInputSpec(BaseInterfaceInputSpec):
+    in_file = File(exists=True, mandatory=True, desc='input 4d image')
+
```
oesteban, line 96:
```suggestion
    accept_3D = traits.Bool(False, usedefault=True, desc='do not fail if a 3D volume is passed in')

```

niworkflows/interfaces/nibabel.py
oesteban: Two issues: better check of dimensions, and `filenii` is not a path that the string interpolation can take in. 

I'd suggest this:
(assuming the nb.squeeze_image and the new input above are accepted)
```suggestion
        ndim = filenii.dataobj.ndim
        if ndim != 4:
            if self.inputs.accept_3D and ndim == 3:
                out_file = str(
                    Path(fname_presuffix(self.inputs.in_file, suffix=f"_idx-000")).absolute()
                )
                self._results['out_files'] = out_file
                filenii.to_filename(out_file)
                return runtime
            raise RuntimeError(f"Input image image is {ndim}D.")
```

niworkflows/interfaces/nibabel.py
```diff
@@ -89,3 +89,66 @@ def _run_interface(self, runtime):
...
+        filenii = nb.load(self.inputs.in_file)
+        if len(filenii.shape) != 4:
+            raise RuntimeError('Input image (%s) is not 4D.' % filenii)
+
+        files_3d = nb.four_to_three(filenii)
+        self._results['out_files'] = []
+        for i, file_3d in enumerate(files_3d):
+            out_file = fname_presuffix(in_file, suffix=f"_idx-{i:03}")
+            file_3d.to_filename(out_file)
+            self._results['out_files'].append(out_file)
```
oesteban, line 120: in_file not defined

```suggestion
        in_file = self.inputs.in_file
        for i, file_3d in enumerate(files_3d):
            out_file = str(
                Path(fname_presuffix(in_file, suffix=f"_idx-{i:03}")).absolute()
            )
            file_3d.to_filename(out_file)
            self._results['out_files'].append(out_file)
```

niworkflows/interfaces/nibabel.py
```diff
@@ -89,3 +89,66 @@ def _run_interface(self, runtime):
...
+
+        return runtime
+
+
+class _MergeSeriesInputSpec(BaseInterfaceInputSpec):
+    in_files = InputMultiObject(File(exists=True, mandatory=True,
+                                     desc='input list of 3d images'))
+
```
oesteban, line 128:
```suggestion
    allow_4D = traits.Bool(True, usedefault=True, desc='whether 4D images are allowed to be concatenated')

```

niworkflows/interfaces/nibabel.py
oesteban: added a switch to allow the strict behavior, see my comment above

niworkflows/interfaces/nibabel.py
```diff
@@ -7,7 +7,7 @@
 from nipype.utils.filemanip import fname_presuffix
```
oesteban, line 7: Please add
```Python
from pathlib import Path
```
before all other imports.

oesteban:
BTW, in the interest of your time, I don't mind taking over the thrust in this PR, @dPys - let me know if you are fine with me editing your code

dPys:
Go for it @oesteban :-)

oesteban:
@effigies, should `squeeze_image` not remove axes from data? I'm creating a test with data of shape (20, 20, 20, 1, 3) and I would expect that would be squeezed to (20, 20, 20, 3) but it doesn't seem to be the case:

```Python
        # Test splitting ANTs warpfields
        data = np.ones((20, 20, 20, 1, 3), dtype=float)
        in_file = tmp_path / 'warpfield.nii.gz'
        nb.Nifti1Image(data, np.eye(4), None).to_filename(str(in_file))
    
>       split = SplitSeries(in_file=str(in_file)).run()
/home/travis/build/nipreps/niworkflows/niworkflows/interfaces/tests/test_nibabel.py:124: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/tmp/venv/lib/python3.6/site-packages/nipype/interfaces/base/core.py:397: in run
    runtime = self._run_interface(runtime)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
self = <niworkflows.interfaces.nibabel.SplitSeries object at 0x7fd69a6c6cc0>
runtime = Bunch(cwd='/tmp/pytest-of-travis/pytest-0/test_SplitSeries0', duration=0.003802, endTime='2020-04-08T20:10:06.709906',...s 5D (20x20x20x1x3).\nAn exception of type RuntimeError occurred while running interface SplitSeries.',), version=None)
        filenii = nb.squeeze_image(nb.load(self.inputs.in_file))
        ndim = filenii.dataobj.ndim
        if ndim != 4:
            if self.inputs.accept_3D and ndim == 3:
                out_file = str(
                    Path(fname_presuffix(self.inputs.in_file, suffix=f"_idx-000")).absolute()
                )
                self._results['out_files'] = out_file
                filenii.to_filename(out_file)
                return runtime
            raise RuntimeError(
>               f"Input image image is {ndim}D ({'x'.join(['%d' % s for s in filenii.shape])}).")
E           RuntimeError: Input image image is 5D (20x20x20x1x3).
/home/travis/build/nipreps/niworkflows/niworkflows/interfaces/nibabel.py:123: RuntimeError
```

effigies:
It removes final 1-dimensions. Check the docstring.

niworkflows/interfaces/nibabel.py
```diff
@@ -72,20 +79,115 @@ class Binarize(SimpleInterface):
...
+    """Split a 4D dataset along the last dimension into a series of 3D volumes."""
+
+    input_spec = _SplitSeriesInputSpec
+    output_spec = _SplitSeriesOutputSpec
+
+    def _run_interface(self, runtime):
+        filenii = nb.squeeze_image(nb.load(self.inputs.in_file))
+        filenii = filenii.__class__(
+            np.squeeze(filenii.dataobj), filenii.affine, filenii.header
+        )
```
effigies, line 123: The reason that non-terminal 1-dimensions are left in place by `squeeze_image` is to preserve the meaning of each dimension. For example, if I have a time series of a single slice `(64, 64, 1, 48)`, `squeeze_image` will preserve the meaning of `i, j, k, n`, but `np.squeeze` will recast `n` as `k`.

What's the use case that you're taking care of here?

oesteban: Yes, the use case I'm contemplating is separating out the three components of deformation fields and other model-based nonlinear transforms. There is one example of this in the tests.

effigies: Found it. I guess I would think splitting along the fifth dimension is a different task from splitting along the fourth, but I see that it's convenient to have a single interface. I'm not sure that risking dropping meaningful dimensions is a good idea.

What about forcing to 4D like:

```Python
extra_dims = tuple(dim for dim in img.shape[3:] if dim > 1) or (1,)
if len(extra_dims) != 1:
    raise ValueError("Invalid shape")
img = img.__class__(img.dataobj.reshape(img.shape[:3] + extra_dims),
                    img.affine, img.header)
```

This coerces a 3D image to (x, y, z, 1) and a 4+D image to (x, y, z, n) assuming that dimensions 4-7 are all 1, n or absent.

niworkflows/interfaces/nibabel.py
```diff
@@ -72,20 +79,115 @@ class Binarize(SimpleInterface):
...
+        ndim = filenii.dataobj.ndim
+        if ndim != 4:
+            if self.inputs.allow_3D and ndim == 3:
```
effigies, line 126: This is odd, as the above will coerce a valid 4D `(x, y, z, 1)` image to 3D `(x, y, z)`, requiring you to then `allow_3D`.

oesteban: why is that odd? It is indeed a 3D volume, right?

oesteban: for instance, there are a fair number of T1w images in OpenNeuro with (x, y, z, 1) dimensions.

effigies: It is a 4D series. You should be able to split it into one 3D volume without special casing it.

niworkflows/interfaces/nibabel.py
oesteban: Is it guaranteed that the spatial dimensions will not be affected on that reshape? If so, I'm down with this solution.

niworkflows/interfaces/nibabel.py
oesteban: I guess the above snippet you wrote would make this particular use-case a standard one?

effigies:
> How much of this you think would be left uncovered after I insert your snippet for the split?

I think it should work fine, but the point was more that we should probably do this in a test-driven way.

> Sure, is there any practical consequence of this?

No. I actually meant to delete it. I was going to say "If the idea is to be making nibabel functions available as interfaces, we've definitely strayed from that goal." But then I decided it didn't matter and failed to remove it.

niworkflows/interfaces/nibabel.py
effigies: I would say checking for 3D volumes should happen before reshaping the image.

niworkflows/interfaces/nibabel.py
effigies: Trivial dimensions have no effect on indexing, whether it's C- or Fortran ordered. If you don't change the order, it's fine.

Issue #334, opened 2019-04-09
DerivativesDataSink: new input ``dismiss_entities``


oesteban:
Specifies a list of entities that should not be propagated from ``in_source`` to the final filename.

  - [ ] Deprecates the ``keep_dtype`` input.

PR #494, opened 2020-04-03
ENH: Make default native resolution explicit

mgxd:
Sets the default value of volumetric references resolution to native.

Originally discussed in https://github.com/nipreps/niworkflows/pull/457#discussion_r375510227, looks it is needed.

Addresses poldracklab/fmriprep#2060 ~~https://github.com/poldracklab/fmriprep/issues/2000~~

### DIFF
--- a/niworkflows/interfaces/bids.py
+++ b/niworkflows/interfaces/bids.py
@@ -454,6 +454,10 @@ desc-preproc_bold.json'
         if len(self.inputs.in_file) > 1 and not isdefined(self.inputs.extra_values):
             formatstr = formatbase + '{suffix}{i:04d}{dtype}{ext}'
 
+        if self.inputs.space and '_res-native' in self.inputs.space:
+            # strip native tag
+            self.inputs.space = self.inputs.space.split('_')[0]
+
         space = '_space-{}'.format(self.inputs.space) if self.inputs.space else ''
         desc = '_desc-{}'.format(self.inputs.desc) if self.inputs.desc else ''
         suffix = '_{}'.format(self.inputs.suffix) if self.inputs.suffix else ''
--- a/niworkflows/interfaces/bids.py
+++ b/niworkflows/interfaces/bids.py
@@ -342,11 +342,12 @@ desc-preproc_bold.json'
     >>> dsink.inputs.source_file = str(tricky_source)
     >>> dsink.inputs.keep_dtype = True
     >>> dsink.inputs.desc = 'preproc'
+    >>> dsink.inputs.space = 'MNI152NLin6Asym_res-01'
     >>> dsink.inputs.RepetitionTime = 0.75
     >>> res = dsink.run()
     >>> res.outputs.out_meta  # doctest: +ELLIPSIS
     '.../niworkflows/sub-02/ses-noanat/func/sub-02_ses-noanat_task-rest_run-01_\
-desc-preproc_bold.json'
+space-MNI152NLin6Asym_res-01_desc-preproc_bold.json'
 
     >>> lines = Path(res.outputs.out_meta).read_text().splitlines()
     >>> lines[1]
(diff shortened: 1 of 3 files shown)


mgxd:
With this change applied, the default BOLD outputs will now have the "res-native" identifier included in filenames.

I'm inclined to edit `DerivativesDataSink` to ignore this identifier, but open to any thoughts/concerns (cc @oesteban @effigies)

effigies:
This looks fine, but I'm missing where it changes anything but filenames.

niworkflows/utils/spaces.py
```diff
@@ -660,10 +660,14 @@ def __call__(self, parser, namespace, values, option_string=None):
...
-            # if ":res-" not in val or ":resolution-" not in val:
-            #     val = ":".join((val, "res-default"))
+            if (
+                val not in NONSTANDARD_REFERENCES
+                and not val.split(':')[0].startswith('fs')
+                and ":res-" not in val
+                and ":resolution-" not in val
+            ):
+                # by default, explicitly set volumetric resolution to native
+                val = ":".join((val, "res-native"))
```
mgxd, line 670: @effigies res-native is already handled correctly, so this change just appends it to the output spaces by default if no resolution is specified

oesteban:
> I'm inclined to edit `DerivativesDataSink` to ignore this identifier, but open to any thoughts/concerns

Check #334

mgxd:
@oesteban actually this shouldn't have much side effects, since the only time`OutputReferencesAction` is used is for the user requested output spaces. 

When we are adding internal spaces, we use the References

> Check #334

that looks useful, but don't see how we can leverage it here - `res-native` isn't included in the source file, but rather in the provided space.

niworkflows/interfaces/bids.py
```diff
@@ -454,6 +456,10 @@ def _run_interface(self, runtime):
         if len(self.inputs.in_file) > 1 and not isdefined(self.inputs.extra_values):
             formatstr = formatbase + '{suffix}{i:04d}{dtype}{ext}'
 
+        if self.inputs.space and '_res-native' in self.inputs.space:
+            # strip native tag
+            self.inputs.space = self.inputs.space.split('_')[0]
+
```
oesteban, line 462: Can't we just drop the resolution entity in the connection? Or the resolution entity is currently crammed into the space input in fMRIPrep and you are working around that hacky solution (honestly, I can't remember)?

niworkflows/utils/spaces.py
```diff
@@ -660,10 +660,14 @@ def __call__(self, parser, namespace, values, option_string=None):
             spaces.checkpoint()
         for val in values:
             val = val.rstrip(":")
-            # Should we support some sort of explicit "default" resolution?
-            # https://github.com/nipreps/niworkflows/pull/457#discussion_r375510227
```
oesteban, line 664: Can we keep this reference and add a reference to this PR as comments?

niworkflows/interfaces/bids.py
mgxd: the resolution is appended to the space input, a hack we implemented to allow multiple resolutions of the same template

niworkflows/utils/spaces.py
mgxd: sure

mgxd:
> Also, I'm not convinced we should always drop the res-native bit. If there are other resolutions we might as well keep it. DYT?

I went back and forth here - ultimately I went with the route of least changes to the documentation. Do we want to make this explicit now (its been implicit for a long time) at the cost of renaming the expected output files? 

If we have other resolutions, those others will have the `res-` fields included. It seems off to produce different filenames based on how many output spaces were requested

oesteban:
I'm also bouncing back and forth on this one. WDYT @effigies ?

effigies:
I would tend to keeping it invisible.

res/den are only required to distinguish from another file. The goal of BIDS is not to put all relevant metadata in the file name but to help people quickly distinguish files. All other resolutions will have a `res-*` entity, so that seems fine.

mgxd:
great, thanks for the reviews!

PR #503, opened 2020-04-22
FIX: Merge/SplitSeries write to path of input image, instead of cwd

oesteban:
Addresses an oversight in #489 .

### DIFF
--- a/niworkflows/interfaces/nibabel.py
+++ b/niworkflows/interfaces/nibabel.py
@@ -1,7 +1,6 @@
 # emacs: -*- mode: python; py-indent-offset: 4; indent-tabs-mode: nil -*-
 # vi: set ft=python sts=4 ts=4 sw=4 et:
 """Nibabel-based interfaces."""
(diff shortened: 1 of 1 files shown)


niworkflows/interfaces/nibabel.py
```diff
@@ -125,7 +125,8 @@ def _run_interface(self, runtime):
         self._results["out_files"] = []
         for i, img_3d in enumerate(nb.four_to_three(img)):
             out_file = str(
-                Path(fname_presuffix(in_file, suffix=f"_idx-{i:03}")).absolute()
+                Path(fname_presuffix(in_file, suffix=f"_idx-{i:03}",
+                                     newpath=runtime.cwd)).absolute()
```
mgxd, line 129: runtime.cwd should be sufficient i believe?
```suggestion
                                     newpath=runtime.cwd))
```

oesteban: Removed `Path` altogether

PR #506, opened 2020-04-27
MAINT: Finalize upstreaming of ants interfaces to nipype

oesteban:
After nipy/nipype#3210 the ``niworkflows/interfaces/ants.py`` can go in whole.

@mgxd we need to cut an RC (talking about *fMRIPrep*) and exercise this particular change in a good range of datasets to make sure all use-cases are covered. Happy to talk about more targetted tests. This integration could be disastrous in the long term.


oesteban:
With [this perspective](https://github.com/nipreps/niworkflows/pull/507#issuecomment-621821539), this PR can be postponed until after the RC.

PR #511, opened 2020-05-05
ENH: Add a pure-Python interface to resample to specific resolutions

oesteban:
This is a replacement for ANTs' ``ResampleImageBySpacing`` with a more
rigorous handling of NIfTI x-form headers.

cc/ @eilidhmacnicol

This may also be useful to @mgxd in the adaptation of antsBrainExtraction to infants.

@effigies' pair of eyes will be very much appreciated, as this new interface plays quite a bit with nibabel's image objects.

### DIFF
--- a/niworkflows/utils/images.py
+++ b/niworkflows/utils/images.py
@@ -179,7 +179,7 @@ def resample_by_spacing(in_file, zooms, order=3, clip=True):
         prefilter=True,
     ).reshape(new_size)
     if clip:
-        resampled = np.clip(resampled, min=data.min(), max=data.max())
+        resampled = np.clip(resampled, a_min=data.min(), a_max=data.max())
 
     # Prepare output x-forms
     sform, scode = hdr.get_sform(coded=True)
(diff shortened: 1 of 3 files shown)


effigies:
Have you looked at nibabel.processing? @kaczmarj just added a conform function in 3.1 that might do a lot of this.

I'll try to have a look tomorrow. I know I'm saying that a lot these days...

oesteban:
> Have you looked at nibabel.processing? @kaczmarj just added a conform function in 3.1 that might do a lot of this.

Thanks, I wasn't aware of that new module. I guess I could use some of the sanity checks he introduced there, but overall this feels more complementary than overlapping. Let me know what you think whenever you find the time :) - no rush.

niworkflows/utils/images.py
```diff
@@ -127,3 +127,83 @@ def dseg_label(in_seg, label, newpath=None):
     new.set_data_dtype(np.uint8)
     new.to_filename(out_file)
     return out_file
+
+
+def resample_by_spacing(in_file, zooms, order=3, clip=True):
```
mgxd, line 132: is this function compatible with 4D images? if not, we might want to add a check JIC

oesteban: I think it is, but the test would be great.

oesteban: Okay, it is not compatible with 4D. NiTransforms needs some updates.

oesteban:
I'm going to go ahead and hit merge. Please note that this change will break compatibility with `rel/1.2.x` and therefore a direct merge cannot be done to backport bugfixes.

PR #513, opened 2020-05-06
MAINT: Revise ``setup.py``/``.cfg``, setuptools pin and PyBIDS pin

oesteban:
Following #507 as these changes were out-of-scope there. In this PR:

  * [x] Updated the PyBIDS pin to ``>=0.10.2`` to include the
    ``build_path`` bug-fixes.
  * [x] Updated setuptools to ``>=38.3.0``, because a fair amount of
    metadata had new support in the ``setup.cfg``
  * [x] Added a list of interesting changes in setuptools, included
    as a comment in ``setup.py``.
  * [x] Moved metadata ``name`` to ``setup.cfg``, as it seems something
    have done in ``setuptools`` themselves. Happy to roll back if GitHub
    won't like this, although a quick search did not reveal anything
    relevant.
  * [x] Minimized ``setup.py``, using ``setup_requires`` from
    ``setup.cfg``. The addition of ``wheel`` for ``bdist_wheel`` is
    maintained. If we finally transition from versioneer to
    setuptools-scm, this will leave us with a very minimal ``setup.py``.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,21 +1,29 @@
 #!/usr/bin/env python
-# -*- coding: utf-8 -*-
-""" niworkflows setup script """
+"""NiWorkflows setup script."""
 import sys
 from setuptools import setup
 import versioneer
 
+setupargs = {
+    "version": versioneer.get_version(),
+    "cmdclass": versioneer.get_cmdclass(),
+}
+
 # Give setuptools a hint to complain if it's too old a version
 # 30.3.0 allows us to put most metadata in setup.cfg
 # 30.4.0 gives us options.packages.find
-# Should match pyproject.toml
-SETUP_REQUIRES = ['setuptools >= 30.4.0']
-# This enables setuptools to install wheel on-the-fly
-SETUP_REQUIRES += ['wheel'] if 'bdist_wheel' in sys.argv else []
+# 34.0.2 ensure extras are honored (e.g., for console_scripts)
+# 36.3 Several files within file: directive in metadata.long_description
+# 36.4 Description-Content-Type medatada, misc fixes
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -1,21 +1,29 @@
...
+# 36.4 Description-Content-Type medatada, misc fixes
+# 36.7 Support setup_requires in setup.cfg files.
+# 38.0 More deterministic builds
+# 38.2.2 fix handling of namespace packages when installing from a wheel.
+# 38.3 Support for long_description_type in setup.cfg, PEP 345 Project-URL metadata
+# Should match pyproject.toml and setup.cfg
+
+if "bdist_wheel" in sys.argv:
+    # This enables setuptools to install wheel on-the-fly
+    setupargs["setup_requires"] = ["setuptools >= 38.3.0", "wheel"]
```
effigies, line 26: Only setting this for bdist_wheel means that if someone has setuptools < 30.3, they won't get a warning about why their build is failing. That's the reason for having `SETUP_REQUIRES` in `setup.py`.

setup.py
```diff
@@ -1,21 +1,29 @@
...
 
-if __name__ == '__main__':
-    setup(name='niworkflows',
-          version=versioneer.get_version(),
-          cmdclass=versioneer.get_cmdclass(),
-          setup_requires=SETUP_REQUIRES,
-          )
+if __name__ == "__main__":
+    setup(**setupargs)
```
effigies, line 29: I don't think expanding a dictionary is clearer than using keyword arguments. It's only four things.

And again, moving the name to `setup.cfg` will break GitHub's dependency graph.

PR #514, opened 2020-05-06
FIX: Two minor amends to #507

oesteban:
Quickfix for two problems:

- [x] Pin PyBIDS >=0.10.2 which includes bugfixes to ``build_path`` (this did not jump out in pinned smriprep, fmriprep, etc. because those projects are already pinned that way, and because we might not have reproduced the bugs).
- [x] Improved stripping of ``.gz`` extensions. We were using ``rstrip`` with the following caveat:

  ```Python
  >>> "some.file.svg".rstrip(".gz")
  'some.file.sv'
  ```

  This PR uses a regexp to ensure this is not occurring. Also, it ~~fixes ``niworkflows.utils.misc.splitext`` for the same issue.~~ just adds a couple of doctests to better cover that function.


oesteban:
These changes should be noncontroversial, and since I'm holding poldracklab/smriprep#183 off for this, I'll go ahead and merge.

PR #520, opened 2020-05-09
CI: Fix skimage 0.17-induced failures

effigies:
imagecodecs has kind of forced our hand with regard to testing lower version pip/setuptools. To handle manylinux2010, we need pip 19+.

Also cleaned up the `.travis.yml` based on config validation, adding an `os` entry, dropping the deprecated `sudo`, and renaming `matrix` to `jobs`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -70,7 +70,6 @@ before_script:
             tar xvzf - -C $HOME/.cache/stanford-crn
         curl -sSL "https://files.osf.io/v1/resources/fvuh8/providers/osfstorage/57f328f6b83f6901ef94cf70" |
             tar xvzf - -C $HOME/.cache/stanford-crn
-        python -c "import skimage"      # XXX: temporary hack; remove when Pooch 1.1.1 or higher is released
     fi
 
 script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 # vim ft=yaml
+os: linux
 dist: xenial
-sudo: true
 language: python
 
 cache:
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -17,25 +17,17 @@ env:
...
     - INSTALL_DEPENDS="pip setuptools"
-  matrix:
+  jobs:
     - CHECK_TYPE="style"
     - CHECK_TYPE="test"
     - INSTALL_TYPE="sdist"
     - INSTALL_TYPE="wheel"
-    - INSTALL_DEPENDS="pip==18.1 setuptools==30.2.1"
-    - INSTALL_DEPENDS="pip==10.0.1 setuptools==30.4.0"
+    - INSTALL_DEPENDS="pip==19.0 setuptools==30.2.1"
```
mgxd, line 25: this should do it?
```suggestion
    - INSTALL_DEPENDS="pip==19.0 setuptools==40.8.0"
```

.travis.yml
```diff
@@ -78,6 +70,7 @@ before_script:
             tar xvzf - -C $HOME/.cache/stanford-crn
         curl -sSL "https://files.osf.io/v1/resources/fvuh8/providers/osfstorage/57f328f6b83f6901ef94cf70" |
             tar xvzf - -C $HOME/.cache/stanford-crn
+        python -c "import skimage"      # XXX: temporary hack; remove when Pooch 1.1.1 or higher is released
```
mgxd, line 73:
```suggestion
```

.travis.yml
effigies: Oh, I see. Yeah, the issue here was verifying that `pip` should be able to install the necessary setuptools to perform the build. If `imagecodecs` is making this impossible, then we can just remove this check.

Alternately, we can change this to `BUILD_DEPENDS`, make sure the wheel builds (dependencies only need to be built on install, not build), and then install it with latest pip/setuptools.

oesteban:
While this is figured out, do we want to pin scikit-image<0.17 not to clog progress of other maintenance PRs?

effigies:
This should work. I'm okay merging this, and then can send a cleanup PR when I get time. I'll just make an issue.

PR #527, opened 2020-05-26
FIX: Ensure dataobj datatype is altered

mgxd:
brought to attention in https://github.com/poldracklab/fmriprep/issues/2142

### DIFF
--- a/niworkflows/interfaces/bids.py
+++ b/niworkflows/interfaces/bids.py
@@ -596,8 +596,16 @@ space-MNI152NLin6Asym_desc-preproc_bold.json'
                         # load updated NIfTI
                         nii = nb.load(out_file, mmap=False)
                     data_dtype = np.dtype(data_dtype)
-                    if nii.get_data_dtype() != data_dtype:
+                    orig_dtype = nii.get_data_dtype()
+                    if orig_dtype != data_dtype:
+                        LOGGER.warning(
+                            f"Changing {out_file} dtype from {orig_dtype} to {data_dtype}"
+                        )
+                        # coerce dataobj to new data dtype
+                        new_data = np.asanyarray(nii.dataobj).astype(data_dtype)
+                        # and set header to match
                         nii.set_data_dtype(data_dtype)
+                        nii = nii.__class__(new_data, nii.affine, nii.header)
                         nii.to_filename(out_file)
 
         if len(self._results["out_file"]) == 1:
--- a/niworkflows/interfaces/bids.py
+++ b/niworkflows/interfaces/bids.py
@@ -602,8 +602,8 @@ space-MNI152NLin6Asym_desc-preproc_bold.json'
                             f"Changing {out_file} dtype from {orig_dtype} to {data_dtype}"
                         )
                         # coerce dataobj to new data dtype
-                        if issubclass(data_dtype, np.integer):
-                            new_data = np.rint(nii.dataobj, dtype=data_dtype)
+                        if np.issubdtype(data_dtype, np.integer):
+                            new_data = np.rint(nii.dataobj).astype(data_dtype)
                         else:
                             new_data = data_dtype(nii.dataobj)
                         # and set header to match
--- a/niworkflows/interfaces/bids.py
+++ b/niworkflows/interfaces/bids.py
@@ -602,7 +602,10 @@ space-MNI152NLin6Asym_desc-preproc_bold.json'
                             f"Changing {out_file} dtype from {orig_dtype} to {data_dtype}"
                         )
                         # coerce dataobj to new data dtype
-                        new_data = np.asanyarray(nii.dataobj, dtype=data_dtype)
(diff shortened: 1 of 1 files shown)


effigies:
This will have no effect. `img.set_data_dtype()` is just a proxy:

https://github.com/nipy/nibabel/blob/966a01e63fd7ce7e58d5971184e31b3305d350f6/nibabel/analyze.py#L931-L932

effigies:
I don't fully understand the `data_dtype` parameter. Is that strictly for the on-disk format? Or is it supposed to coerce the data array, as well?

mgxd:
It was meant to only effect the on-disk format. I'm a bit hesitant to coerce the data within the datasink interface, as there is already a lot going on there. If we want to go that route, I'd feel more comfortable splitting that into its own Node

niworkflows/interfaces/bids.py
```diff
@@ -596,8 +596,16 @@ def _run_interface(self, runtime):
...
                         nii = nb.load(out_file, mmap=False)
                     data_dtype = np.dtype(data_dtype)
-                    if nii.get_data_dtype() != data_dtype:
+                    orig_dtype = nii.get_data_dtype()
+                    if orig_dtype != data_dtype:
+                        LOGGER.warning(
+                            f"Changing {out_file} dtype from {orig_dtype} to {data_dtype}"
+                        )
+                        # coerce dataobj to new data dtype
+                        new_data = np.asanyarray(nii.dataobj).astype(data_dtype)
```
oesteban, line 605:
```suggestion
                        new_data = np.asanyarray(nii.dataobj, dtype=data_dtype)
```

effigies: Is this going to cast BOLD data to int16?

mgxd: i think https://github.com/poldracklab/smriprep/issues/195 is part of the problem, but there is some further step downstream that is upcasting to float. But this is also necessary for any conversion to match `source_file` (currently just `boldref`).

mgxd: @effigies this should only affect data suffixes:
```
        ("mask", "uint8"),
        ("dseg", "int16"),
        ("probseg", "float32"),
        ("boldref", "source")
```
unless the `data_dtype` flag is explicitly checked, which I don't believe it ever is.

oesteban: > i think [poldracklab/smriprep#195](https://github.com/poldracklab/smriprep/issues/195) is part of the problem, but there is some further step downstream that is upcasting to float.

commented there, yes there is a resampling downstream in this case - moving to standard space.

niworkflows/interfaces/bids.py
```diff
@@ -596,8 +596,16 @@ def _run_interface(self, runtime):
...
                         nii = nb.load(out_file, mmap=False)
                     data_dtype = np.dtype(data_dtype)
-                    if nii.get_data_dtype() != data_dtype:
+                    orig_dtype = nii.get_data_dtype()
+                    if orig_dtype != data_dtype:
+                        LOGGER.warning(
+                            f"Changing {out_file} dtype from {orig_dtype} to {data_dtype}"
+                        )
+                        # coerce dataobj to new data dtype
+                        new_data = np.asanyarray(nii.dataobj, dtype=data_dtype)
```
effigies, line 605: Can't just cast to int without rounding.

```Python
>>> np.asanyarray(50.99999, np.int16)
array(50, dtype=int16)
```

What about:
```suggestion
                        if issubclass(data_dtype, np.integer):
                            new_data = np.rint(nii.dataobj, dtype=data_dtype)
                        else:
                            new_data = data_dtype(nii.dataobj)
```

mgxd: nice catch!

niworkflows/interfaces/bids.py
```diff
@@ -596,8 +596,19 @@ def _run_interface(self, runtime):
...
                         nii = nb.load(out_file, mmap=False)
                     data_dtype = np.dtype(data_dtype)
-                    if nii.get_data_dtype() != data_dtype:
+                    orig_dtype = nii.get_data_dtype()
+                    if orig_dtype != data_dtype:
+                        LOGGER.warning(
+                            f"Changing {out_file} dtype from {orig_dtype} to {data_dtype}"
+                        )
+                        # coerce dataobj to new data dtype
+                        if issubclass(data_dtype, np.integer):
```
oesteban, line 605:
```suggestion
                        if np.issubdtype(data_dtype, np.integer):
```

mgxd:
I kept running into:
```
TypeError: No loop matching the specified signature and casting was found for ufunc rint
```
when trying to set `dtype` in the `rint()` call. 

Logically, it would have to round first and then apply the dtype to the array anyways.

mgxd:
i'll merge this in now and cut a new release, thank you both for the reviews

PR #528, opened 2020-05-26
FIX: Non-integer data coercion initialization

mgxd:
Supplements #527; Non-integer data initialization wasn't being exercised by the tests, and was leading to:
```
TypeError: 'numpy.dtype' object is not callable
```


mgxd:
travis is not updating, but the build is passing: 

https://travis-ci.org/github/nipreps/niworkflows/builds/691510156

👍 ?

PR #529, opened 2020-05-27
FIX: Do not break figure-datatype derivatives by sessions

oesteban:
This PR amends #507 so that the reportlets are not broken into sessions. The session is kept after the subject key though, so these files are still identifiable.


effigies:
Looks reasonable.

PR #532, opened 2020-06-02
FIX: Avoid diverting CIFTI dtype from original BOLD

mgxd:
Since outputs would change, this PR targets `1.3.0`

### DIFF
--- a/niworkflows/interfaces/cifti.py
+++ b/niworkflows/interfaces/cifti.py
@@ -330,7 +330,6 @@ def _create_cifti_image(
         BOLD data saved as CIFTI dtseries
     """
     bold_img = nb.load(bold_file)
-    bold_dtype = bold_img.get_data_dtype()
     label_img = nb.load(label_file)
     if label_img.shape != bold_img.shape[:3]:
         warnings.warn("Resampling bold volume to match label dimensions")
@@ -340,14 +339,14 @@ def _create_cifti_image(
     bold_img = _reorient_image(bold_img, orientation="LAS")
     label_img = _reorient_image(label_img, orientation="LAS")
 
-    bold_data = bold_img.get_fdata()
+    bold_data = bold_img.get_fdata(dtype="float32")
     timepoints = bold_img.shape[3]
     label_data = np.asanyarray(label_img.dataobj).astype("int16")
 
     # Create brain models
     idx_offset = 0
     brainmodels = []
-    bm_ts = np.empty((timepoints, 0))
+    bm_ts = np.empty((timepoints, 0), dtype="float32")
 
     for structure, labels in CIFTI_STRUCT_WITH_LABELS.items():
         if labels is None:  # surface model
@@ -355,8 +354,8 @@ def _create_cifti_image(
             # use the corresponding annotation
             hemi = structure.split("_")[-1]
             # currently only supports L/R cortex
-            surf = nb.load(bold_surfs[hemi == "RIGHT"])
-            surf_verts = len(surf.agg_data()[0])
+            surf_ts = nb.load(bold_surfs[hemi == "RIGHT"])
+            surf_verts = len(surf_ts.darrays[0].data)
             if annotation_files[0].endswith(".annot"):
                 annot = nb.freesurfer.read_annot(annotation_files[hemi == "RIGHT"])
                 # remove medial wall
@@ -365,7 +364,7 @@ def _create_cifti_image(
                 annot = nb.load(annotation_files[hemi == "RIGHT"])
                 medial = np.nonzero(annot.darrays[0].data)[0]
             # extract values across volumes
-            ts = np.array([tsarr.data[medial] for tsarr in surf.darrays])
+            ts = np.array([tsarr.data[medial] for tsarr in surf_ts.darrays])
 
             vert_idx = ci.Cifti2VertexIndices(medial)
             bm = ci.Cifti2BrainModel(
@@ -404,7 +403,7 @@ def _create_cifti_image(
                 voxel_indices_ijk=vox,
             )
             idx_offset += len(vox)
-            bm_ts = np.column_stack((bm_ts, ts.T)).astype(bold_dtype)
+            bm_ts = np.column_stack((bm_ts, ts.T))
         # add each brain structure to list
         brainmodels.append(bm)
 
@@ -440,7 +439,8 @@ def _create_cifti_image(
     matrix.append(geometry_map)
     matrix.metadata = ci.Cifti2MetaData(meta)
     hdr = ci.Cifti2Header(matrix)
-    img = ci.Cifti2Image(dataobj=bm_ts, header=hdr, nifti_header=bold_img.header)
+    img = ci.Cifti2Image(dataobj=bm_ts, header=hdr)
+    img.set_data_dtype(bold_img.get_data_dtype())
     img.nifti_header.set_intent("NIFTI_INTENT_CONNECTIVITY_DENSE_SERIES")
 
     out_file = "{}.dtseries.nii".format(split_filename(bold_file)[1])
--- a/niworkflows/interfaces/cifti.py
+++ b/niworkflows/interfaces/cifti.py
@@ -330,6 +330,7 @@ def _create_cifti_image(
         BOLD data saved as CIFTI dtseries
     """
     bold_img = nb.load(bold_file)
+    bold_dtype = bold_img.get_data_dtype()
     label_img = nb.load(label_file)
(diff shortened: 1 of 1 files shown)


niworkflows/interfaces/cifti.py
```diff
@@ -355,7 +356,7 @@ def _create_cifti_image(
             hemi = structure.split("_")[-1]
             # currently only supports L/R cortex
             surf = nb.load(bold_surfs[hemi == "RIGHT"])
-            surf_verts = len(surf.darrays[0].data)
+            surf_verts = len(surf.agg_data()[0])
```
effigies, line 359: Do not assume order. Try something like:

```suggestion
            surf_verts = len(img.agg_data('pointset'))
```

mgxd: order shouldn't matter here, it's just polling the number of vertices of a given data array

niworkflows/interfaces/cifti.py
```diff
@@ -330,6 +330,7 @@ def _create_cifti_image(
         BOLD data saved as CIFTI dtseries
     """
     bold_img = nb.load(bold_file)
+    bold_dtype = bold_img.get_data_dtype()
```
mgxd, line 333: from my testing, this is still honoring the original NIfTI data type

niworkflows/interfaces/cifti.py
```diff
@@ -391,7 +392,7 @@ def _create_cifti_image(
                     else np.concatenate((ts, bold_data[ijk]))
                 )
                 vox += [
-                    [ijk[0][ix], ijk[1][ix], ijk[2][ix]] for ix, row in enumerate(ts)
+                    [ijk[0][idx], ijk[1][idx], ijk[2][idx]] for idx in range(len(ts))
```
mgxd, line 395: replacing enumerate with range

niworkflows/interfaces/cifti.py
```diff
@@ -403,7 +404,7 @@ def _create_cifti_image(
                 voxel_indices_ijk=vox,
             )
             idx_offset += len(vox)
-            bm_ts = np.column_stack((bm_ts, ts.T))
+            bm_ts = np.column_stack((bm_ts, ts.T)).astype(bold_dtype)
```
mgxd, line 407: this will convert cortical + subcortical values to the bold_dtype - basically, is this fine to do?

niworkflows/interfaces/cifti.py
effigies: You're using `[0]`. A surface file is a pointset and a triangles. These don't have the same shape.

niworkflows/interfaces/cifti.py
effigies: If the original has an on-disk dtype of int16, this is going to induce rounding. Why not just `set_data_dtype` at the end?

niworkflows/interfaces/cifti.py
```diff
@@ -339,7 +340,7 @@ def _create_cifti_image(
     bold_img = _reorient_image(bold_img, orientation="LAS")
     label_img = _reorient_image(label_img, orientation="LAS")
 
-    bold_data = bold_img.get_fdata(dtype="float32")
+    bold_data = bold_img.get_fdata()
```
effigies, line 343: Is there a good reason to upcast to float64?

niworkflows/interfaces/cifti.py
mgxd: These are functional (`func.gii`) GIFTIs, which I believe don't have any triangles included?

(Looking back, `surf` probably wasn't the best variable name).

niworkflows/interfaces/cifti.py
mgxd: nope, can revert that

niworkflows/interfaces/cifti.py
oesteban: If that is the case (faces are not included), what is the reason for doing so? If we are dropping the triangles, I believe we should re-add them. There's no value in dropping them.

effigies: Not sure this change is worth making, then. It makes a copy of all the data just for the purpose of taking the length of the first dimension.

effigies: Oscar, this is a time series. They don't have coordinates and faces.

niworkflows/interfaces/cifti.py
```diff
@@ -439,7 +440,7 @@ def _create_cifti_image(
     matrix.append(geometry_map)
     matrix.metadata = ci.Cifti2MetaData(meta)
     hdr = ci.Cifti2Header(matrix)
-    img = ci.Cifti2Image(bm_ts, hdr)
+    img = ci.Cifti2Image(dataobj=bm_ts, header=hdr, nifti_header=bold_img.header)
```
effigies, line 443: You probably don't want to reused the NIfTI header. All you want to do is set the dtype.

```suggestion
    img = ci.Cifti2Image(dataobj=bm_ts, header=hdr)
    img.set_data_dtype(bold_img.get_data_dtype())
```

I think you could replace the whole PR with this. Maybe changing L349 to:

```Diff
-    bm_ts = np.empty((timepoints, 0))
+    bm_ts = np.empty((timepoints, 0), dtype=np.float32)
```

to save a little memory.

niworkflows/interfaces/cifti.py
oesteban: GIFTI does not impose what you can't embed on the file. NIfTI is great in the sense that, you can access the data and reconstruct the sampling structure. If vertices and faces are not stored along with the time-series, it is definitely not self-contained and also prone to failures. Whether the visualization software does not check whether the surface is packaged within the CIFTI, that's a different issue. But I believe we should try to generate self-contained derivatives whenever that's possible.

effigies: Do any tools accept this?

niworkflows/interfaces/cifti.py
```diff
@@ -364,7 +364,7 @@ def _create_cifti_image(
                 annot = nb.load(annotation_files[hemi == "RIGHT"])
                 medial = np.nonzero(annot.darrays[0].data)[0]
             # extract values across volumes
-            ts = np.array([tsarr.data[medial] for tsarr in surf.darrays])
+            ts = np.array([tsarr.data[medial] for tsarr in surf_ts.darrays])
```
effigies, line 367: Oh, here would be a good place to use `agg_data()`.

```suggestion
            ts = surf_ts.agg_data()[medial, :]
```

niworkflows/interfaces/cifti.py
mgxd: FWIW, CIFTIs are not intended to store surface information - generally surface geometry information is supplemental

niworkflows/interfaces/cifti.py
mgxd: doesn't `agg_data()` return a tuple?

effigies: If the image has multiple types of data array, or you request multiple intents, then no.

If you want to be safe (e.g., if we start putting surfaces inside functional files...):

```suggestion
            ts = surf_ts.agg_data("time series")[medial, :]
```

effigies: FWIW this case is explicitly covered in the docs: https://nipy.org/nibabel/reference/nibabel.gifti.html#nibabel.gifti.gifti.GiftiImage.agg_data

mgxd: running fmriprep with this change leads to:
```
   ts = surf_ts.agg_data("time series")[medial, :]
TypeError: tuple indices must be integers or slices, not tuple
```

I think keeping it as it was is simplest ATM

effigies: Sure. That seems really weird, though. If you can `pdb` it at some point, I would be interested in the details.

effigies: Oh, if there are no darrays, you do get an empty tuple.

niworkflows/interfaces/cifti.py
oesteban: > Do any tools accept this?

Okay, after reflecting a bit on this - there's the problem that one of these timeseries may map on to several surfaces (midthickness, inflated, sphere...). I guess, in the extreme, you'd want to keep the information of triangles. But then you'll always need the pointset so it's kind of pointless (did you see what I did there?). I guess there's no option but just not storing surface information.



> CIFTIs are not intended to store surface information

Yup, but we are talking about GIFTI here, right?

PR #533, opened 2020-06-02
ENH: Improve FS license checking

mgxd:
This improves the check for a valid FS license. New additions:

- ~~if the given license `lic` is a directory, return `False`~~
- in some cases, i.e. license file exists but is not valid / older format, "ERROR" was sent to `stdout`, not `stderr`. This combines both streams for easier parsing.
- `mri_convert` command was simplified


niworkflows/utils/misc.py
```diff
@@ -313,8 +313,9 @@ def check_valid_fs_license(lic=None):
 
     if lic is not None:
         import os
-
         os.environ["FS_LICENSE"] = str(Path(lic).absolute())
```
oesteban, line 316: To me, this is still the biggest issue. Attempting to influence the environment from a helper function is not explicit, and it is likely to end up having unexpected effects. In other words: we do have a canary but we are placing it on a different coal mine.

All the environment manipulations and checks should happen in the config module. That's the only way to be sure that environment variables will have the contents we expect. I haven't checked properly, but I bet we are running this canary somewhere in the setup of fMRIPrep, very likely within a `multiprocessing.Process`. In other words, when the Process ends, this setting of `FS_LICENSE` has no effect outside the process (i.e., a different coal mine).

The canary should just call mri_convert and check the output. Just as any other nipype interface in the workflow. The resolution of the license path and setting up the environment should happen elsewhere.

mgxd: we aren't, it should be happening on the main thread prior to spawning a subprocess to build the workflow
https://github.com/poldracklab/fmriprep/blob/eb3a40b827d277ce661e31c27f6ace40ef3c2be6/fmriprep/cli/parser.py#L550-L570

When I implemented the `lic`, the idea was to try and encapsulate all licensing logic within the function. But sure, we can remove the implicit setting of the envvar. The function will just have to assume the license has been located / handled appropriately.

oesteban: The design problem is that you can't predict the result and side-effects of the function unless you are told the context in which it is run.

mgxd: Alright, so far I tested this with the follow:

- Valid license at $FREESURFER_HOME/license.txt, `True`
- Valid license at $FS_LICENSE, `True`
- No license, `False`
- Directory as license, `False`
- Invalid license, `False`

niworkflows/utils/tests/test_misc.py
```diff
@@ -17,17 +16,17 @@ def test_pass_dummy_scans(algo_dummy_scans, dummy_scans, expected_out):
...
-        (True, "custom/license.txt", b""),
+        (b"Successful command", 0, True),
+        (b"", 0, True),
+        (b"ERROR: FreeSurfer license file /made/up/license.txt not found", 1, False),
+        (b"Failed output", 1, False),
+        (b"ERROR: Systems running GNU glibc version greater than 2.15", 0, False),
     ],
 )
-def test_fs_license_check(valid, lic, stderr):
+def test_fs_license_check(tmp_path, stdout, rc, valid):
```
mgxd, line 28:
```suggestion
def test_fs_license_check(stdout, rc, valid):
```

effigies:
Can we revert the `lic=None` bit, and just mock the environment?

e.g.:

```
from unittest import mock
with mock.patch.dict(os.environ, {'FS_LICENSE': lic}):
    ....
```

This was a mid-bug-fix API change.

PR #535, opened 2020-06-04
FIX: sentinel path

mgxd:


niworkflows/utils/tests/test_misc.py
```diff
@@ -39,6 +39,7 @@ def test_fs_license_check2(monkeypatch):
     assert check_valid_fs_license() is True
 
 
+@pytest.mark.skipif(not os.getenv("FREESURFER_HOME"), reason="FreeSurfer not installed")
```
oesteban, line 42:
```suggestion
```
Shouldn't this test always pass (regardless of whether you have access to an fs_license or not?

The only skipif you could have is when `mri_convert` is not present in the environment

mgxd: yep, i was using `FREESURFER_HOME` as an indication of that. I guess using `shutil.which` would be more precise

oesteban: in the end I think there won't be a big difference, let's just keep the skipif.

mgxd: i did it anyways for the sake of best practices

oesteban:
```
niworkflows/utils/tests/test_misc.py::test_fs_license_check2 SKIPPED
```
this shouldn't be happening...

.circleci/config.yml
```diff
@@ -214,7 +214,9 @@ jobs:
             docker run -u $( id -u ) -it --rm=false -w /src/niworkflows \
               -e COVERAGE_FILE=/tmp/summaries/.pytest.coverage \
               -e TEST_DATA_HOME=/data -v /tmp/data:/data \
+              -e FS_LICENSE=/etc/fslicense.txt \
               -v ${PWD}:/tmp niworkflows:latest \
+              -v /tmp/fslicense/license.txt:/etc/fslicense.txt:ro \
```
oesteban, line 219:
```suggestion
              -e FS_LICENSE=/etc/fslicense.txt \
              -v /tmp/fslicense/license.txt:/etc/fslicense.txt:ro \
              -v ${PWD}:/tmp niworkflows:latest \
```

mgxd:
looking good now. i'm gonna merge this in and release so 20.1.1 isn't breaking anymore
```
niworkflows/utils/tests/test_misc.py::test_fs_license_check2 PASSED
 
niworkflows/utils/tests/test_misc.py::test_fs_license_check3 PASSED
 ```

PR #505, opened 2020-04-23
ENH: Add utility function to quickly check FS license

mgxd:
Closes #502

PR #538, opened 2020-06-07
FIX: Restore and deprecate license argument to check_valid_fs_license

effigies:
#533 broke API mid-series. This restores it while also not modifying the environment.

Target: 1.2.6.

### DIFF
--- a/niworkflows/utils/misc.py
+++ b/niworkflows/utils/misc.py
@@ -307,6 +307,10 @@ def check_valid_fs_license(lic=None):
 
     env = os.environ.copy()
     if lic is not None:
+        import warnings
+        warnings.warn("The license argument has been deprecated. Please set the environment "
+                      "if needed before calling this function without arguments.",
+                      DeprecationWarning)
         env["FS_LICENSE"] = os.path.abspath(lic)
 
     with TemporaryDirectory() as tmpdir:
--- a/niworkflows/utils/misc.py
+++ b/niworkflows/utils/misc.py
@@ -289,7 +289,7 @@ def pass_dummy_scans(algo_dummy_scans, dummy_scans=None):
     return dummy_scans
 
 
-def check_valid_fs_license():
+def check_valid_fs_license(lic=None):
     """
     Run ``mri_convert`` to assess FreeSurfer access to a license.
 
@@ -299,10 +299,15 @@ def check_valid_fs_license():
         FreeSurfer successfully executed (valid license)
 
     """
+    import os
     from pathlib import Path
     import subprocess as sp
     from tempfile import TemporaryDirectory
     from pkg_resources import resource_filename
+
+    env = os.environ.copy()
+    if lic is not None:
+        env["FS_LICENSE"] = os.path.abspath(lic)
 
     with TemporaryDirectory() as tmpdir:
         # quick FreeSurfer command
@@ -311,7 +316,7 @@ def check_valid_fs_license():
             resource_filename("niworkflows", "data/sentinel.nii.gz"),
             str(Path(tmpdir) / "out.mgz"),
         )
-        proc = sp.run(_cmd, stdout=sp.PIPE, stderr=sp.STDOUT)
+        proc = sp.run(_cmd, stdout=sp.PIPE, stderr=sp.STDOUT, env=env)
     return proc.returncode == 0 and "ERROR:" not in proc.stdout.decode()
 
 
--- a/niworkflows/utils/misc.py
+++ b/niworkflows/utils/misc.py
@@ -308,7 +308,7 @@ def check_valid_fs_license(lic=None):
     env = os.environ.copy()
     if lic is not None:
         import warnings
-        warnings.warn("The license argument has been deprecated. Please set the environment "
+        warnings.warn("The license argument has been deprecated and will be removed in 1.3.0. Please set the environment "
                       "if needed before calling this function without arguments.",
                       DeprecationWarning)
         env["FS_LICENSE"] = os.path.abspath(lic)
--- a/niworkflows/utils/misc.py
+++ b/niworkflows/utils/misc.py
@@ -299,20 +299,17 @@ def check_valid_fs_license(lic=None):
         FreeSurfer successfully executed (valid license)
 
     """
-    import os
     from pathlib import Path
     import subprocess as sp
     from tempfile import TemporaryDirectory
     from pkg_resources import resource_filename
 
-    env = os.environ.copy()
     if lic is not None:
         import warnings
-        warnings.warn("The license argument has been deprecated and will be removed in 1.3.0. "
-                      "Please set the environment if needed before calling this function "
-                      "without arguments.",
+        warnings.warn("The license argument has been deprecated, has no effect, and will be "
+                      "removed in 1.3.0. Please set the environment if needed before calling "
+                      "this function without arguments.",
                       DeprecationWarning)
-        env["FS_LICENSE"] = os.path.abspath(lic)
 
     with TemporaryDirectory() as tmpdir:
         # quick FreeSurfer command
@@ -321,7 +318,7 @@ def check_valid_fs_license(lic=None):
             resource_filename("niworkflows", "data/sentinel.nii.gz"),
             str(Path(tmpdir) / "out.mgz"),
         )
-        proc = sp.run(_cmd, stdout=sp.PIPE, stderr=sp.STDOUT, env=env)
+        proc = sp.run(_cmd, stdout=sp.PIPE, stderr=sp.STDOUT)
     return proc.returncode == 0 and "ERROR:" not in proc.stdout.decode()
 
 
--- a/niworkflows/utils/misc.py
+++ b/niworkflows/utils/misc.py
@@ -308,8 +308,9 @@ def check_valid_fs_license(lic=None):
     env = os.environ.copy()
     if lic is not None:
         import warnings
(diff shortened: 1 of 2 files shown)


niworkflows/utils/misc.py
```diff
@@ -299,19 +299,24 @@ def check_valid_fs_license():
...
     """
+    import os
     from pathlib import Path
     import subprocess as sp
     from tempfile import TemporaryDirectory
     from pkg_resources import resource_filename
 
+    env = os.environ.copy()
+    if lic is not None:
+        env["FS_LICENSE"] = os.path.abspath(lic)
```
oesteban, line 310: I would not bring this back. This makes the behavior of the function unpredictable.

effigies: Can you elaborate? It seems very predictable to me. Unless you're concerned somebody's using a pre-6.0.1 version of FreeSurfer?

oesteban: IMHO, the canary should not alter the environment. Otherwise, it stops being usable in every context because you need to understand the environment.

Let's put the following example:

1. sMRIPrep runs in multiproc mode.
2. In the main process, but after the worker pool has been initialized, you check the license with the canary (with `lic` not `None`), `FS_LICENSE` was not set in the environment and the license is not under `FREESURFER_HOME`.
3. The canary succeeds because it accommodated its environment.
4. The first worker that needs a license fails, because the environment is different.

So, if the canary modifies the environment, you need to run the canary in every context it may be pertinent and always be careful of the order of operations.

oesteban: It is a problem of expectations: the canary should be a smoke test on the environment not a unit test of the license.

effigies: I see. Then the purpose of it has changed. I would suggest that we deprecate the unit test capacity instead of removing it and making the test fail for downstream tools that still expect the result to be a meaningful assertion about a passed argument.

Would this be an acceptable compromise? It can be a very short deprecation window (i.e., remove completely in 1.3.0).

oesteban:
I've been also thinking about whether this is really and API-breaking change justifying a minor bump rather than a bugfix bump.
- If interpreted as deserving a minor bump, then this PR should be based on `maint/1.2.x` not master (i.e., to be released on 1.3.0)
- But, beyond that, I don't think the canary change should force the minor bump. As per the documentation:
  > **API compatibility.** *Workflow-generating functions, workflow input- and outputnode fields must not change. As an end-user application, this may seem overly strict, but the odds of introducing a bug are much higher in these cases.*

  #533 did not change any workflow-generating function or workflow API. I don't think it changes the user interface either (and the other two reasons to permit just patch-bumps are also met, IMHO).

effigies:
Agreed. I meant to target 1.2.6.

I think breaking the build of a downstream tool that depends on this minor series is a problem, even if it's not a workflow-generating function.

effigies:
Went ahead and added the deprecation while rebasing.

oesteban:
I'll merge conversations in one if you don't mind:

> Then the purpose of it has changed (https://github.com/nipreps/niworkflows/pull/538#discussion_r436368703).

Not sure I follow. After re-reading poldracklab/smriprep#43 I still think the canary was not meant to interact with the coal mine at all. I'm afraid that we (and more precisely, myself) just made a bad call when accepting the first PR, which implicitly did change the purpose of it by including the environment modification. I can remember I raised some concerns, but with the information I had at the moment, I chose not to stop the merge. Let's not make my mistake again.

> I think breaking the build of a downstream tool that depends on this minor series is a problem, even if it's not a workflow-generating function.

We should definitely consider this to be included in https://www.nipreps.org/devs/releases/ - although I see it as an additional checkpoint, rather than part of the API changes item.

That said, and following the above conversation, I think we should actually use that little flexibility we allowed ourselves unknowingly to make maintenance easier. After all, we have already broken sMRIPrep - maintaining 1.2.x becomes more costly if we bring the environment setting back. To me, the anti-pattern of breaking downstream builds is much less relevant than that of interacting with the global scope (or the environment in this case).

effigies:
Is it making maintenance easier to break downstream? The point of having these rules is that we can make changes upstream without immediately searching all downstream projects for uses that need to change. All of that can be done at minor version bumps.

Also, this particular change is not costly. It's one code path that is currently used. I've added a test to ensure it works as expected, and deprecated that particular usage. We can move away, but we don't have to break smriprep 0.6.0 and 0.6.1.

niworkflows/utils/misc.py
```diff
@@ -299,19 +299,28 @@ def check_valid_fs_license():
...
+    import os
     from pathlib import Path
     import subprocess as sp
     from tempfile import TemporaryDirectory
     from pkg_resources import resource_filename
 
+    env = os.environ.copy()
+    if lic is not None:
+        import warnings
+        warnings.warn("The license argument has been deprecated. Please set the environment "
```
mgxd, line 311: this is most likely too long but let's note the quick removal.

```suggestion
        warnings.warn("The license argument has been deprecated and will be removed in 1.3.0. Please set the environment "
```

niworkflows/utils/misc.py
```diff
@@ -299,19 +299,28 @@ def check_valid_fs_license():
...
     from pathlib import Path
     import subprocess as sp
     from tempfile import TemporaryDirectory
     from pkg_resources import resource_filename
 
+    env = os.environ.copy()
+    if lic is not None:
+        import warnings
+        warnings.warn("The license argument has been deprecated and will be removed in 1.3.0. Please set the environment "
+                      "if needed before calling this function without arguments.",
```
effigies, line 312:
```suggestion
        warnings.warn("The license argument has been deprecated and will be removed in 1.3.0. "
                      "Please set the environment if needed before calling this function "
                      "without arguments.",
```

oesteban:
Sorry for my push-back - it was my oversight accepting #510 without further thought. I had a couple of comments drafted that I wanted to share:

- We are not _breaking 0.6.0 and 0.6.1_. We are breaking sMRIPrep's CLI under some particular settings. In addition, the problem can be worked around just by setting the `FS_LICENSE` env variable. Also, fMRIPrep is unaffected.
- We didn't actually change the purpose of the canary: even the initial attempt to implement the canary did not modify the environment (#505), then #510 kicked in.

Would it be a good compromise that we:
- Bring back the `lic=None` argument to the canary's signature for the 0.6.x series.
- Do not bring back modifying the environment of the canary
- Do not modify tests (should be unaffected if we do not modify the implementation of the canary)
- Cut niworkflows-1.2.6
- Prohibit niworkflows 1.2.4 and 1.2.5 as smriprep's dependencies.

I think these actions plus poldracklab/smriprep#205 should put the mess #533 surfaced under control.

WDYT?

niworkflows/utils/misc.py
```diff
@@ -304,6 +304,13 @@ def check_valid_fs_license():
     from tempfile import TemporaryDirectory
     from pkg_resources import resource_filename
 
+    if lic is not None:
+        import warnings
+        warnings.warn("The license argument has been deprecated, has no effect, and will be "
+                      "removed in 1.3.0. Please set the environment if needed before calling "
+                      "this function without arguments.",
+                      DeprecationWarning)
```
effigies, line 312: Now purely deprecation.

niworkflows/utils/tests/test_misc.py
```diff
@@ -34,8 +34,9 @@ def test_fs_license_check(stdout, rc, valid):
         assert check_valid_fs_license() is valid
 
 
-@pytest.mark.skipif(not os.getenv("FS_LICENSE"), reason="No FS license found")
-def test_fs_license_check2(monkeypatch):
+@pytest.mark.skipif(shutil.which('mri_convert') is None, reason="FreeSurfer not installed")
+@pytest.mark.xfail(not os.getenv("FS_LICENSE"), reason="No FS license found")
+def test_fs_license_check2():
```
effigies, line 39: I think this is a more reasonable marking, since I was already fiddling with this file. Can remove it, if you disagree.

niworkflows/utils/tests/test_misc.py
```diff
@@ -46,3 +47,12 @@ def test_fs_license_check3(monkeypatch):
...
         assert check_valid_fs_license() is False
+
+
+def test_fs_license_check_deprecated_arg():
+    """Execute the canary with passed license."""
+    with mock.patch("subprocess.run") as mocked_run:
+        mocked_run.return_value.stdout = b""
+        mocked_run.return_value.returncode = 0
+        with pytest.deprecated_call():
+            check_valid_fs_license(lic="Any non-None value")
```
effigies, line 58: Kept deprecation test.

oesteban:
I'm holding merging this one for after #534 because after merging this, a direct merge of `maint/1.2.x` into `master` will not work. Makes sense?

effigies:
I don't think I follow.

oesteban:
I mean, to keep `master` up to date with `main/1.2.x` we can just merge `maint/1.2.x` into `master`. But after we merge this PR into `main/1.2.x` then a direct merge will not be possible anymore.

effigies:
Merge master into maint/1.2.x, merge this PR, merge maint/1.2.x into master using `--strategy ours` to ensure that these changes stay out of master with future merges.

effigies:
Actually, reverse that first one.

oesteban:
I'll have to study the `--strategy ours` - this would be the first time I use it...

PR #541, opened 2020-06-11
RF: update to py36+

### LINKED ISSUE #486: MAINT: remove PY35 workarounds

mgxd:
Closes #486 

context manager change context: https://bugs.python.org/issue25994

### DIFF
--- a/niworkflows/reports/core.py
+++ b/niworkflows/reports/core.py
@@ -137,9 +137,8 @@ class Reportlet(Element):
         if not config:
             raise RuntimeError("Reportlet must have a config object")
 
-        # PY35: Sorted config dict for consistent behavior
         self.name = config.get(
-            "name", "_".join("%s-%s" % i for i in sorted(config["bids"].items()))
+            "name", "_".join("%s-%s" % i for i in config["bids"].items())
         )
         self.title = config.get("title")
         self.subtitle = config.get("subtitle")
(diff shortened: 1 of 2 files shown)


niworkflows/reports/core.py
```diff
@@ -137,9 +137,8 @@ def __init__(self, layout, out_dir, config=None):
         if not config:
             raise RuntimeError("Reportlet must have a config object")
 
-        # PY35: Sorted config dict for consistent behavior
         self.name = config.get(
-            "name", "_".join("%s-%s" % i for i in sorted(config["bids"].items()))
+            "name", "_".join("%s-%s" % i for i in config["bids"].items())
```
oesteban, line 141: Not sure dropping the `sorted` here is a good move. Given the size of the dictionary to be sorted, I would probably leave it (just remove the comment above) just not to break compatibility with previous runs. WDTY?

niworkflows/utils/misc.py
```diff
@@ -253,17 +253,18 @@ def clean_directory(path):
...
+            if f.is_file() or f.is_symlink():
+                try:
+                    os.unlink(f.path)
+                except OSError:
+                    return False
+            elif f.is_dir():
+                try:
+                    shutil.rmtree(f.path)
+                except OSError:
+                    return False
```
oesteban, line 267: Should we move this to `pathlib.Path` which I think gives a much more succinct, regular and dependable interface?

niworkflows/reports/core.py
mgxd: This is targeting a new minor, so it's fine to break compatibility no? I don't feel strongly about this though, so I can revert.

Issue #281, opened 2018-05-31
Contours in space


effigies:
https://s3.amazonaws.com/openneuro.outputs/921294bd5b869b1852ab3ce886583795/4dd151e3-52d1-4fa2-9591-27c16520331c/fmriprep/sub-001.html#Functional

See top-right of **ROIs in BOLD space**:

![bold1](https://user-images.githubusercontent.com/83442/40788307-b9958cea-64bd-11e8-91dd-7c3f8092d350.png)

PR #543, opened 2020-06-20
FIX: Use ``numpy.linspace`` to calculate mosaic plots' cutting planes

oesteban:
References: #281.


oesteban:
Reportlets generated by tests look good to me.

Issue #542, opened 2020-06-19
Artifacts in mosaic plots when intersection of masks and visualization planes is minimal


oesteban:
This problem has been around for a long while. When the intersection of the mask with the visualization plane only contains a fraction of a voxel, nilearn plotting gets very confused:

<img src="https://5825-53608443-gh.circle-artifacts.com/0/tmp/ds054/derivatives/fmriprep/sub-100185/figures/sub-100185_task-machinegame_run-2_desc-rois_bold.svg" />

With poldracklab/fmriprep#2130 the problem is exacerbated (not so much for fully sampled datasets, but definitely for our test data).

I believe this problem would completely be fixed if we replaced the affine information of the input images to be aligned with the cardinal axes. To do so, I'd propose:

1. `nibabel.as_closest_canonical` to all volumetric inputs of plots.
2. Generate a new affine with: image_zooms x identity
3. Perhaps re-center the volume for visualization purposes (so reported coordinates are w.r.t. the center of the volume).

This way, visualization will cut through full voxels always and the marching cubes algorithm will not go bananas.

In general, this approach would also make our reports more transparent because:
1. We would be able to safely drop interpolations by nilearn
2. The user will have access to the visualization of the naked data array, instead of a reoriented version.

Opinions?

oesteban:
This is another manifestation of #281 which escapes the solution in #304 because the glitch does not only happen at views on the edge of the brain.

effigies:
Probably better to comment here than on #544. So the issue, to be clear, is that nilearn can't handle images that do not have identical affine and FoV? Or can't handle oblique images?

oesteban:
The second, nilearn trips when these three happen together:

- Plotting with `add_contour()`
- Contour (mask) has sparse voxels
- Oblique affine

PR #544, opened 2020-06-20
FIX: Segmentation plots aligned with cardinal axes

oesteban:
While nilearn/nilearn#2525 is considered, this PR advances the generation of mosaics after aligning the affine matrix of the data with the cardinal axes. This should overcome the glitches when nilearn is requested to plot contours and the mask for the ROI has voxels that only partially intersect with the visualization plane.

Resolves: #542.
References: #281, #304.
Depends: #543.

### DIFF
--- a/niworkflows/utils/images.py
+++ b/niworkflows/utils/images.py
@@ -3,6 +3,16 @@ import nibabel as nb
 import numpy as np
 
 
+def as_canonical(img):
+    """Drop rotation w.r.t. cardinal axes of input image."""
+    img = nb.as_closest_canonical(img)
+    zooms = list(img.header.get_zooms()[:3])
+    newaff = np.diag(zooms + [1])
+    rot = newaff[:3, :3].dot(np.linalg.inv(img.affine[:3, :3]))
+    newaff[:3, 3] = rot.dot(img.affine[:3, 3])
+    return nb.Nifti1Image(img.dataobj, newaff, img.header)
+
+
 def unsafe_write_nifti_header_and_data(fname, header, data):
     """Write header and data without any consistency checks or data munging
 
--- a/niworkflows/utils/images.py
+++ b/niworkflows/utils/images.py
@@ -3,14 +3,26 @@ import nibabel as nb
 import numpy as np
 
 
-def as_canonical(img):
-    """Drop rotation w.r.t. cardinal axes of input image."""
+def rotation2canonical(img):
+    """Calculate the rotation w.r.t. cardinal axes of input image."""
     img = nb.as_closest_canonical(img)
-    zooms = list(img.header.get_zooms()[:3])
-    newaff = np.diag(zooms + [1])
-    rot = newaff[:3, :3].dot(np.linalg.inv(img.affine[:3, :3]))
-    newaff[:3, 3] = rot.dot(img.affine[:3, 3])
-    return nb.Nifti1Image(img.dataobj, newaff, img.header)
+    newaff = np.diag(img.header.get_zooms()[:3])
+    r = newaff @ np.linalg.inv(img.affine[:3, :3])
+    if np.allclose(r, np.eye(3)):
+        return None
+    return r
+
+
+def rotate_affine(img, rot=None):
+    """Rewrite the affine of a spatial image."""
+    if rot is None:
+        return img
+
+    img = nb.as_closest_canonical(img)
+    affine = np.eye(4)
+    affine[:3, :3] = rot @ img.affine[:3, :3]
+    affine[:3, 3] = rot @ img.affine[:3, 3]
+    return img.__class__(img.dataobj, affine, img.header)
 
 
 def unsafe_write_nifti_header_and_data(fname, header, data):
--- a/niworkflows/utils/images.py
+++ b/niworkflows/utils/images.py
@@ -7,7 +7,7 @@ def rotation2canonical(img):
     """Calculate the rotation w.r.t. cardinal axes of input image."""
     img = nb.as_closest_canonical(img)
     newaff = np.diag(img.header.get_zooms()[:3])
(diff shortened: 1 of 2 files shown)


oesteban:
Before:



After:



(there should not be any difference here because all our tests are on images aligned with cardinal axes, I know the reports look good for fMRIPrep with oblique datasets).

niworkflows/utils/images.py
```diff
@@ -3,6 +3,16 @@
 import numpy as np
 
 
+def as_canonical(img):
+    """Drop rotation w.r.t. cardinal axes of input image."""
+    img = nb.as_closest_canonical(img)
+    zooms = list(img.header.get_zooms()[:3])
+    newaff = np.diag(zooms + [1])
+    rot = newaff[:3, :3].dot(np.linalg.inv(img.affine[:3, :3]))
+    newaff[:3, 3] = rot.dot(img.affine[:3, 3])
```
effigies, line 12: This can be simplified. Rendering in a more readable form:

```Python
rot = newaff[:3, :3] @ (img.affine[:3, :3] ** -1)
newaff[:3, 3] = rot @ img.affine[:3, 3]
```

Substituting:

```Python
newaff[:3, 3] = newaff[:3, :3] @ (img.affine[:3, :3] ** -1) @ img.affine[:3, 3]
newaff[:3, 3] = newaff[:3, :3]
```

```suggestion
```

niworkflows/viz/utils.py
```diff
@@ -222,19 +223,27 @@ def plot_segs(
...
+    Plot segmentation as contours over the image (e.g. anatomical).
     seg_niis should be a list of files. mask_nii helps determine the cut
     coordinates. plot_params will be passed on to nilearn plot_* functions. If
     seg_niis is a list of size one, it behaves as if it was plotting the mask.
     """
     plot_params = {} if plot_params is None else plot_params
 
-    image_nii = _3d_in_file(image_nii)
+    image_nii = as_canonical(_3d_in_file(image_nii))
+    seg_niis = [as_canonical(_3d_in_file(f)) for f in seg_niis]
```
effigies, line 237: These are guaranteed to have the same affines? If not, the implicit rotation by removing the rotational components of the affine my produce a misalignment.

Might be worth having an affine check before calling `as_canonical`.

niworkflows/viz/utils.py
```diff
@@ -222,19 +223,27 @@ def plot_segs(
...
     data = image_nii.get_fdata()
 
     plot_params = robust_set_limits(data, plot_params)
 
-    bbox_nii = nb.load(image_nii if bbox_nii is None else bbox_nii)
+    bbox_nii = (
+        image_nii if bbox_nii is None
+        else as_canonical(_3d_in_file(bbox_nii))
+    )
```
effigies, line 245: I'm not entirely sure what bbox is supposed to be, so it's hard to say if reorienting it can be potentially problematic.

niworkflows/viz/utils.py
oesteban: In principle, right now it should allow you to stick in a NIfTI clipping some area (e.g., a face mask given the circumstances) with different voxel size and FoV. Not sure we are ever using this.

What if we run `as_canonical` only once and apply the same rotation matrix to all other inputs?

niworkflows/viz/utils.py
oesteban: bbox is a mask from which the visible extent of the plot will be calculated. Typically, you pass a brain mask and the mosaic will be zoomed in into the bounding box of that brain mask (thus, making sure you don't render blank planes).

oesteban:
I think an alternative approach would be to densely resample the masks so that you make sure some voxel center will fall close enough of the cutting plane so that the marching cubes algorithm doesn't trip. This seemed too heuristic to me, as there's no easy way to calculate the upsampling factor you need to ensure the contour extraction will work.

Dropping the rotation, however, does ensure the intersection of visualization planes will not be oblique.

In addition, this _deobliquing_ for visualization would be also useful, e.g., to plot SDC.

niworkflows/utils/images.py
oesteban: Tried the `img.affine[:3, :3] ** -1` but SVD does not converge

oesteban:
> Since the original NIfTI affine is preserved in the outputs, I don't see a huge problem with altering it in the context of a plot, though perhaps I'm missing something?

Some of these concerns are being discussed in nilearn/nilearn#2525 - although, while I see them crucial for nilearn (as a general-purpose library), I don't think they make any difference in the context of niworkflows.

effigies:
Okay, I think I finally properly understand the situation, and that what we want is to see the data array, with properly scaled voxels and orientation information, but the plotter should not be given obliquity (rotation and shear) information that would cause it to rotate the voxels relative to the pixels in the display grid.

Assuming this is right, I will check the code from this perspective.

effigies:
Question: is it the case that the two images should always have the same affine/FoV? I would assume so, but it doesn't seem to be documented.

oesteban:
> what we want is to see the data array, with properly scaled voxels and orientation information

Correct

> plotter should not be given obliquity (rotation and shear) information that would cause it to rotate the voxels relative to the pixels in the display grid

If "display grid" = the current plane being plotted, yes this is correct.


> is it the case that the two images should always have the same affine/FoV? I would assume so, but it doesn't seem to be documented.

The two images should have the same rotation w.r.t. the cardinal axes. In theory, they could have different zooms because the rotation is extracted from the reference and then applied to the masks. However, I don't know (as you said, it is not documented anywhere) whether nilearn imposes having the same affine. Either way, this PR only requires the rotation w.r.t. cardinal to be the same.

effigies:
@oesteban Do you want to rerun the test? I can't see any reason it should suddenly start failing.

effigies:
Tests pass locally. Merging.

PR #545, opened 2020-06-26
fix: skip t1w file existence check if anat_derivatives are provided

bpinsard:
Changes to allow processing dataset without T1w series if `--anat_derivatives` are provided.
See https://github.com/poldracklab/fmriprep/pull/2201

### DIFF
--- a/niworkflows/interfaces/bids.py
+++ b/niworkflows/interfaces/bids.py
@@ -222,7 +222,7 @@ class BIDSDataGrabber(SimpleInterface):
 
     def __init__(self, *args, **kwargs):
         anat_only = kwargs.pop("anat_only")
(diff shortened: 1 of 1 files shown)


effigies:
@mgxd Should we target 1.3.0 or 1.2.7 with this?

mgxd:
I think this is fine for either, but since the behavior will not be included into the 20.1.x series of fmriprep, I'm leaning towards 1.3.0 - wdyt?

effigies:
Sure. I just haven't kept up with the current status of dependencies. Merging into master.

mgxd:
oh, nevermind - it's in 😄

effigies:
Suggestion LGTM. Feel free to patch on master.

PR #547, opened 2020-08-11
RF: Enable DerivativesDataSink to take multiple source files to derive entities

effigies:
This change permits DerivativesDataSink to take multiple files in order to extract entities. It keeps common entities and discards differing entities.

Doctest included to demonstrate.

Alternative to nipreps/smriprep#223.


niworkflows/interfaces/bids.py
```diff
@@ -279,7 +279,8 @@ class _DerivativesDataSinkInputSpec(DynamicTraitedSpec, BaseInterfaceInputSpec):
         File(exists=True), mandatory=True, desc="the object to be saved"
     )
     meta_dict = traits.DictStrAny(desc="an input dictionary containing metadata")
-    source_file = File(exists=False, mandatory=True, desc="the input func file")
+    source_file = InputMultiObject(
```
mgxd, line 282: one thing to note: we might want to deprecate `source_file` and replace it with `source_files`

PR #549, opened 2020-08-12
ENH: Add ``smooth`` input to ``RegridToZooms``

oesteban:
Resampling (up- or down-) typically requires a gussian smoothing to be
run before, so that outliers are not introduced (esp. when
downsampling).

This PR adds this option, relying on scipy.

This is the first element of a series of PRs deconstructing #506 into digestible, independent bits.

### DIFF
--- a/niworkflows/interfaces/images.py
+++ b/niworkflows/interfaces/images.py
@@ -42,6 +42,13 @@ class _RegridToZoomsInputSpec(BaseInterfaceInputSpec):
         usedefault=True,
         desc="clip the data array within the original image's range",
     )
+    smooth = traits.Either(
+        False,
+        traits.Bool(),
+        traits.Float(),
+        usedefault=True,
+        desc="apply gaussian smoothing before resampling"
+    )
 
 
 class _RegridToZoomsOutputSpec(TraitedSpec):
@@ -65,6 +72,7 @@ class RegridToZooms(SimpleInterface):
             self.inputs.zooms,
             order=self.inputs.order,
             clip=self.inputs.clip,
+            smooth=self.inputs.smooth,
         ).to_filename(self._results["out_file"])
         return runtime
 
--- a/niworkflows/interfaces/images.py
+++ b/niworkflows/interfaces/images.py
@@ -43,9 +43,9 @@ class _RegridToZoomsInputSpec(BaseInterfaceInputSpec):
         desc="clip the data array within the original image's range",
     )
     smooth = traits.Either(
-        False,
         traits.Bool(),
         traits.Float(),
+        default=False,
         usedefault=True,
         desc="apply gaussian smoothing before resampling"
     )
--- a/niworkflows/utils/images.py
+++ b/niworkflows/utils/images.py
@@ -148,7 +148,7 @@ def dseg_label(in_seg, label, newpath=None):
     return out_file
 
 
-def resample_by_spacing(in_file, zooms, order=3, clip=True):
+def resample_by_spacing(in_file, zooms, order=3, clip=True, smooth=False):
     """Regrid the input image to match the new zooms."""
     from pathlib import Path
     import numpy as np
@@ -164,7 +164,6 @@ def resample_by_spacing(in_file, zooms, order=3, clip=True):
 
     hdr = in_file.header.copy()
     dtype = hdr.get_data_dtype()
-    data = np.asanyarray(in_file.dataobj)
     zooms = np.array(zooms)
 
     # Calculate the factors to normalize voxel size to the specific zooms
@@ -194,6 +193,13 @@ def resample_by_spacing(in_file, zooms, order=3, clip=True):
     ijk = np.linalg.inv(card).dot(
         new_card.dot(np.vstack((new_grid, np.ones((1, new_grid.shape[1])))))
     )
+
+    if smooth:
+        from scipy.ndimage import gaussian_filter
+        data = gaussian_filter(in_file.get_fdata(),
+                               2 if smooth is True else smooth).astype(dtype)
+    else:
+        data = np.asanyarray(in_file.dataobj)
 
     # Resample data in the new grid
     resampled = map_coordinates(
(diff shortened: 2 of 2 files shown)


niworkflows/utils/images.py
```diff
@@ -195,6 +194,13 @@ def resample_by_spacing(in_file, zooms, order=3, clip=True):
         new_card.dot(np.vstack((new_grid, np.ones((1, new_grid.shape[1])))))
     )
 
+    if smooth:
+        from scipy.ndimage import gaussian_filter
+        data = gaussian_filter(in_file.get_fdata(),
+                               2 if smooth is True else smooth).astype(dtype)
```
effigies, line 200: Should we default to something besides 2? Scikit-image uses (max downsampling factor - 1) / 2.

https://github.com/scikit-image/scikit-image/blob/1188c33c3defefdc0f672870b5bcc36beaec967f/skimage/transform/_warps.py#L120

niworkflows/interfaces/images.py
```diff
@@ -42,6 +42,13 @@ class _RegridToZoomsInputSpec(BaseInterfaceInputSpec):
         usedefault=True,
         desc="clip the data array within the original image's range",
     )
+    smooth = traits.Either(
+        False,
+        traits.Bool(),
+        traits.Float(),
```
mgxd, line 48: shouldn't the traits go first here?
```suggestion
        traits.Bool(),
        traits.Float(),
        False,
```

niworkflows/utils/images.py
```diff
@@ -195,6 +194,13 @@ def resample_by_spacing(in_file, zooms, order=3, clip=True):
         new_card.dot(np.vstack((new_grid, np.ones((1, new_grid.shape[1])))))
     )
 
+    if smooth:
+        from scipy.ndimage import gaussian_filter
+        data = gaussian_filter(in_file.get_fdata(),
```
mgxd, line 199: why the difference of `in_file.get_fdata()` and `in_file.dataobj`?

oesteban: `gaussian_filter` wants float, and nibabel does it conveniently. Please note the `astype` in the next line

effigies: You'll probably want to round if re-casting to int, otherwise you truncate 2.9 to 2.

niworkflows/interfaces/images.py
oesteban: I think either takes the first argument for default. To be sure:

```suggestion
        traits.Bool(),
        traits.Float(),
        default=False,
```

(please note that `Either` is not regular and doesn't take `default_value`)

oesteban:
@mgxd @effigies please double-check my changes make sense and address the problems identified. thanks!

effigies:
Scaling factors are based on zooms, not dimensions. And I don't understand why you're using get_data_dtype to determine the type of the interpreted data.

oesteban:
> Scaling factors are based on zooms, not dimensions.

Right, thanks for catching this.



> And I don't understand why you're using get_data_dtype to determine the type of the interpreted data.

TBH, because I thought that is the correct way of doing it. Please suggest what would be best.

effigies:
`get_data_dtype` tells you what type things are on disk. If you want to know the type of the interpreted data, you can use `np.asanyarray(img.dataobj).dtype` or `img.dataobj[0, 0, 0].dtype`.

What's the goal of manually setting the dtype? Is there a case where you want to clip your data to integers?

oesteban:
> What's the goal of manually setting the dtype? Is there a case where you want to clip your data to integers?

The output image should be of the same dtype of the input - so I guess this is correct?

effigies:
What should be correct? I'm having trouble following the conversation. Can you provide a concrete case where setting the dtype might help?

oesteban:
I have removed the casting in the latest commit. After thinking about it, there's no need to re-cast the data because `map_coordinates` will ensure the output data type matches that of the input image. I agree that introducing a rounding before `map_coordinates` just wasted accuracy for nothing. Is that what you meant?

Please note that we still need to interrogate the data type on disk because the output of this function should not modify the datatype. Is there any automated operation on the dtype by nibabel? (either casting the data array to the dtype stated on the header or vice-versa, changing the nifti header to reflect the datatype of the data array.

niworkflows/utils/images.py
```diff
@@ -195,11 +193,19 @@ def resample_by_spacing(in_file, zooms, order=3, clip=True):
...
+        data = gaussian_filter(in_file.get_fdata(), smooth)
+    else:
+        data = np.asarray(in_file.dataobj)
+
     # Resample data in the new grid
     resampled = map_coordinates(
         data,
         ijk[:3, :],
-        output=dtype,
+        output=hdr.get_data_dtype(),
```
effigies, line 208: I would just drop this. Therefore `resampled` will be whatever type `data` is.

```suggestion
```

effigies:
> Is there any automated operation on the dtype by nibabel? (either casting the data array to the dtype stated on the header or vice-versa, changing the nifti header to reflect the datatype of the data array.

The data will be cast to the dtype set in the header. If scale factors are needed to make it work, they will be set.

The only time you determine the on-disk dtype from the data object is if you don't pass in a header.

oesteban:
> The data will be cast to the dtype set in the header.

Got it

> If scale factors are needed to make it work, they will be set.

Does this mean the cast does not round? I'm not sure we don't want to round in this case.

effigies:
> Does this mean the cast does not round?

It will round to the precision of your output type. Basically what you get is:

```Python
def rescale_data(data, dtype):
    # assume dtype is int
    dinfo = np.iinfo(dtype)
    slope = (data.max() - data.min()) / (dinfo.max - dinfo.min)
    inter = data.min() - dinfo.min * slope
    new_data = (data - data.min()) / slope + dinfo.min
    return new_data.astype(dtype), slope, inter

def reconstruct_data(data, slope, inter):
    return data * slope + inter

data = np.random.default_rng().normal(1000, 200, size=(5,5,5))
new_data, slope, inter = rescale_data(data, 'int16')
reconstructed = reconstruct_data(new_data, slope, inter)
np.abs(reconstructed - data).max() 
```

> I'm not sure we don't want to round in this case.

Round to what?

oesteban:
> Round to what?

``` Python
if issubclass(input_nii.header.get_data_dtype(), numbers.Integer) and not issubclass(data.dtype, numbers.Integer):
    data = np.round(data).astype(input_nii.header.get_data_dtype())
```

effigies:
That's done implicitly.

Issue #553, opened 2020-08-13
MultiApplyTransforms does not support `identity` transform in list


emdupre:
Discovered in https://github.com/poldracklab/fmriprep/pull/2239

Here's a reproducible example, assuming you have a toy 'affine.txt' file locally:

```python
> from niworkflows.interfaces import itk  

> bold_to_t1_transform = itk.MultiApplyTransforms()
> bold_to_t1_transform.inputs.transforms = ['./affine.txt', './affine.txt']  # OK
> bold_to_t1_transform.inputs.transforms = 'identity'  # OK
> bold_to_t1_transform.inputs.transforms = ['identity', './affine.txt']  # Fails

TraitError: The 'transforms' trait of a _MultiApplyTransformsInputSpec instance must be a list of items which are a pathlike object or string representing an existing file or 'identity', but a value of ['identity', './affine.txt'] <class 'list'> was specified.
```

A transform of `['identity', 'identity']` fails similarly. I can look into this tonight.

effigies:
I'm taking a shot at this now.

effigies:
Annoyingly, the bug here is a Nipype problem. If we pass `identity` to regular `ApplyTransforms`, we may need to patch that, too.

oesteban:
There were concerns that led to not implementing the identity because it's not documented. But yes, this pertains in nipype's proper.

PR #554, opened 2020-08-13
FIX: Patch ApplyTransforms spec to permit identity in a chain

effigies:
First step in addressing #553.

The current trait is:

```Python
    transforms = traits.Either(
        InputMultiPath(File(exists=True)),
        "identity",
        argstr="%s",
        mandatory=True,
        desc="transform files: will be applied in reverse order. For "
        "example, the last specified transform will be applied first.",
    ) 
```

This will deserve its own nipype patch, but let's not stand on ceremony, there.

### DIFF
--- a/niworkflows/interfaces/fixes.py
+++ b/niworkflows/interfaces/fixes.py
@@ -4,9 +4,9 @@ import os
 
 import nibabel as nb
 
-from nipype.interfaces.base import traits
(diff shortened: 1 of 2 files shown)


effigies:
Thanks for the reviews, @emdupre and @mgxd. Merging to smooth the path for tomorrow.

@oesteban Please shout if you have concerns.

PR #558, opened 2020-08-17
FIX: Derive DerivativesDataSink sidecar filenames from the source basename

mgxd:
Encountered during https://github.com/nipreps/smriprep/pull/215

This changes the metadata sidecar filename generation to account for multiple suffixes not covered by https://github.com/nipreps/niworkflows/blob/e3cf33aa7888263b17d8180de93938d1ea42fee4/niworkflows/utils/misc.py#L184-L205

### DIFF
--- a/niworkflows/interfaces/bids.py
+++ b/niworkflows/interfaces/bids.py
@@ -33,7 +33,7 @@ from nipype.interfaces.io import add_traits
 from templateflow.api import templates as _get_template_list
 from ..utils.bids import _init_layout, relative_to_root
 from ..utils.images import overwrite_header
-from ..utils.misc import splitext as _splitext, _copy_any
+from ..utils.misc import _copy_any
 
 regz = re.compile(r"\.gz$")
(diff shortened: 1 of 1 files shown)


oesteban:
I'm concerned about going ahead with this - I share the views of Chris w.r.t. disambiguation. Plus, if we use the T1w metadata for the Euler number, then this can be pushed away for a little while until there's something more specific in BIDS.

mgxd:
> Out of curiosity, why the hidden variable?

There was already an `out_file` define, so I wanted to avoid confusing them. Probably overkill looking back.

> I'm concerned about going ahead with this - I share the views of Chris w.r.t. disambiguation. Plus, if we use the T1w metadata for the Euler number, then this can be pushed away for a little while until there's something more specific in BIDS.

That's fair, this will definitely change fmriprep outputs (CIFTI-2). Though, since `...bold.dtseries.json` files are already being created, we should add this to the spec sooner rather than later.

PR #561, opened 2020-09-03
FIX: Improved control over correlations plot

oesteban:
This PR adds a ``columns`` argument to the interface and underlying function so that _fMRIPrep_ can select what confounds will be used to estimate the correlations.

The PR also minimizes to 20 (it was 70!) the number of signals that are plotted.

References: poldracklab/fmriprep#1746
References: https://github.com/poldracklab/fmriprep/issues/2245#issuecomment-682618147


effigies:
> The PR also minimizes to 20 (it was 70!)

I'd noticed that when looking through this yesterday. I was also going to suggest 20...

oesteban:
This seems to be working as expected - Merge?

oesteban:
This is pretty safe, I think. I'll go ahead, merge and release.

Issue #400, opened 2019-09-24
BIDS-Derivatives compliance revision


oesteban:
Double-check that the current infrastructure fully supports BIDS-Derivatives:

- [x] Revise BIDS entities and default patterns in the [`nipreps.json`](https://github.com/nipreps/niworkflows/blob/master/niworkflows/data/nipreps.json) config.
- [x] AROMA metadata key forces acronyms to CamelCase
  ```
    "aroma_motion_1": {
      "CsfFraction": 0.06,
      "EdgeFraction": 0.44,
      "HighFrequencyContent": 0.09,
      "MaximumRpCorrelation": 0.81,
      "Method": {
        "Name": "ICA-AROMA",
        "Version": "0.4.4-beta"
      },
      "MotionNoise": false
    },
  ```

oesteban:
@rciric any chance you can look into this anytime soon?

oesteban:
I have reused the somewhat anecdotical original issue about aroma

PR #563, opened 2020-09-08
FIX: AROMA metadata ``CsfFraction`` -> ``CSFFraction``

oesteban:
A minimal change to the capitalization of the metadata dictionary key.

Resolves: #400.


oesteban:
This silly patch works: https://15806-53175327-gh.circle-artifacts.com/0/tmp/ds005/derivatives/fmriprep/sub-01/func/sub-01_task-mixedgamblestask_run-1_desc-confounds_regressors.json

Issue #559, opened 2020-08-28
Problems generating FLIRT BBR reportlets


oesteban:
```
Node Name: fmriprep_wf.single_subject_100104_wf.func_preproc_task_machinegame_run_001_wf.bold_reg_wf.fsl_bbr_wf.flt_bbr

/oak/stanford/groups/russpold/data/ds000054/0.0.5/derivatives/fmriprep-20.2.0rc0/fmriprep/sub-100104/log/20200827-094552_164fe55c-bbf2-4682-aaeb-10edf72f3ccb/crash-20200827-110133-oesteban-flt_bbr-7ebe8bf9-77cd-4a80-bd91-cdc1bcbedc63.txt
Working Directory: /lscratch/100104/fmriprep_wf/single_subject_100104_wf/func_preproc_task_machinegame_run_001_wf/bold_reg_wf/fsl_bbr_wf/flt_bbr
Inputs:
angle_rep:
apply_isoxfm:
apply_xfm:
args:
bbrslope:
bbrtype:
bgvalue:
bins:
coarse_search:
compress_report: auto
cost:
cost_func: bbr
datatype:
display_init:
dof: 6
echospacing:
environ: {'FSLOUTPUTTYPE': 'NIFTI_GZ'}
fieldmap:
fieldmapmask:
fine_search:
force_scaling:
in_file:
in_matrix_file:
in_weight:
interp:
min_sampling:
no_clamp:
no_resample:
no_resample_blur:
no_search:
out_file:
out_log:
out_matrix_file:
out_report: report.svg
output_type: NIFTI_GZ
padding_size:
pedir:
ref_weight:
reference:
rigid2D:
save_log:
schedule: /usr/share/fsl/5.0/etc/flirtsch/bbr.sch
searchr_x:
searchr_y:
searchr_z:
sinc_width:
sinc_window:
uses_qform:
verbose:
wm_seg:
wmcoords:
wmnorms:
Traceback (most recent call last):
  File "/usr/local/miniconda/lib/python3.7/site-packages/nipype/pipeline/plugins/multiproc.py", line 67, in run_node
    result["result"] = node.run(updatehash=updatehash)
  File "/usr/local/miniconda/lib/python3.7/site-packages/nipype/pipeline/engine/nodes.py", line 516, in run
    result = self._run_interface(execute=True)
  File "/usr/local/miniconda/lib/python3.7/site-packages/nipype/pipeline/engine/nodes.py", line 635, in _run_interface
[...]

PR #566, opened 2020-09-11
FIX: Reportlets would crash in edge condition

oesteban:
The function that calculates the bounding box (and the cuts) to be
represented in reportlets has default thresholds thought out for brain
masks.

When passed a white-matter mask, these thresholds could yield only one
cut (below the required number of cuts).

This patch checkes that the number of cuts are sufficient in all the
cutting planes.

Resolves: #559.

### DIFF
--- a/niworkflows/viz/utils.py
+++ b/niworkflows/viz/utils.py
@@ -174,7 +174,7 @@ def cuts_from_bbox(mask_nii, cuts=3):
         smin, smax = (0, mask_data.shape[ax] - 1)
 
         B = np.argwhere(c > th)
-        if not B.size or B.size < cuts:  # Threshold too high
+        if B.size < cuts:  # Threshold too high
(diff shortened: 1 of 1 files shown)


oesteban:
Tested locally and works 👍

mgxd:
LGTM

niworkflows/viz/utils.py
```diff
@@ -174,7 +174,7 @@ def cuts_from_bbox(mask_nii, cuts=3):
         smin, smax = (0, mask_data.shape[ax] - 1)
 
         B = np.argwhere(c > th)
-        if not B.size:  # Threshold too high
+        if not B.size or B.size < cuts:  # Threshold too high
```
effigies, line 177: `B.size` is always numeric:

```suggestion
        if B.size < cuts:  # Threshold too high
```

Issue #531, opened 2020-05-29
RFC: Brain extraction workflow


mgxd:
taking improvements from:

- https://github.com/nipreps/nirodents/blob/master/nirodents/workflows/brainextraction.py
- https://github.com/nipreps/nibabies/blob/master/nibabies/workflows/brain_extraction.py

PR #567, opened 2020-09-16
FIX: Revision of ``antsBrainExtraction``, better handling edge cases

oesteban:
This PR introduces changes to better handle the copying of headers within the workflow, which is a potential source for nipreps/smriprep#127, and addresses some problems that surface when the ATROPOS refinement step fails (nipreps/smriprep#125).

This refactor could be considered a preliminary work on #531. 

  - [x] Updates the nodes with pure python interfaces based on nibabel, minimizing the need for the new ``copy_header`` of ANTs' nipype interfaces.
  - [x] Reorganizes the workflow so that the Atropos refinement is completely self-contained.
  - [x] If a WM prior is found within the template's structure, then it is mapped on to the individual's brain and fed as ``weight_image`` in the N4 correction.
  - [x] Adds regularization of the WM posterior estimated with ATROPOS, by multiplying with the WM prior mapped from the template. In the future, an easy check could be implemented here to dismiss the ATROPOS refinement if the prior and the posterior do not overlap sufficiently (i.e., the WM was not correctly labeled or segmented)

### DIFF
--- a/niworkflows/anat/ants.py
+++ b/niworkflows/anat/ants.py
@@ -271,8 +271,10 @@ def init_brain_extraction_wf(
     try:
         init_aff.inputs.search_grid = (40, (0, 40, 40))
     except ValueError:
-        warn("antsAI's option --search-grid was added in ANTS 2.3.0 "
-             f"({init_aff.interface.version} found.)")
+        warn(
+            "antsAI's option --search-grid was added in ANTS 2.3.0 "
+            f"({init_aff.interface.version} found.)"
+        )
 
     # Set up spatial normalization
     settings_file = (
@@ -291,7 +293,9 @@ def init_brain_extraction_wf(
     norm.inputs.float = use_float
     fixed_mask_trait = "fixed_image_mask"
 
-    if norm.interface.version and parseversion(norm.interface.version) >= Version("2.2.0"):
+    if norm.interface.version and parseversion(norm.interface.version) >= Version(
+        "2.2.0"
+    ):
         fixed_mask_trait += "s"
 
     map_brainmask = pe.Node(
@@ -329,9 +333,11 @@ def init_brain_extraction_wf(
     try:
         inu_n4_final.inputs.rescale_intensities = True
     except ValueError:
-        warn("N4BiasFieldCorrection's --rescale-intensities option was added in ANTS 2.1.0 "
-             f"({inu_n4_final.interface.version} found.) Please consider upgrading.",
-             UserWarning)
+        warn(
+            "N4BiasFieldCorrection's --rescale-intensities option was added in ANTS 2.1.0 "
+            f"({inu_n4_final.interface.version} found.) Please consider upgrading.",
+            UserWarning,
+        )
 
     # Apply mask
     apply_mask = pe.MapNode(ApplyMask(), iterfield=["in_file"], name="apply_mask")
@@ -743,9 +749,11 @@ def init_atropos_wf(
     try:
         inu_n4_final.inputs.rescale_intensities = True
     except ValueError:
-        warn("N4BiasFieldCorrection's --rescale-intensities option was added in ANTS 2.1.0 "
-             f"({inu_n4_final.interface.version} found.) Please consider upgrading.",
-             UserWarning)
+        warn(
+            "N4BiasFieldCorrection's --rescale-intensities option was added in ANTS 2.1.0 "
+            f"({inu_n4_final.interface.version} found.) Please consider upgrading.",
+            UserWarning,
+        )
 
     # Apply mask
     apply_mask = pe.MapNode(ApplyMask(), iterfield=["in_file"], name="apply_mask")
@@ -972,9 +980,11 @@ def init_n4_only_wf(
     try:
         inu_n4_final.inputs.rescale_intensities = True
     except ValueError:
-        warn("N4BiasFieldCorrection's --rescale-intensities option was added in ANTS 2.1.0 "
-             f"({inu_n4_final.interface.version} found.) Please consider upgrading.",
-             UserWarning)
+        warn(
+            "N4BiasFieldCorrection's --rescale-intensities option was added in ANTS 2.1.0 "
+            f"({inu_n4_final.interface.version} found.) Please consider upgrading.",
+            UserWarning,
+        )
 
     # fmt: off
     wf.connect([
--- a/niworkflows/anat/ants.py
+++ b/niworkflows/anat/ants.py
@@ -67,7 +67,7 @@ def init_brain_extraction_wf(
     The workflow follows the following structure:
 
       1. Step 1 performs several clerical tasks (preliminary INU correction,
-         calculating the Laplacian of inputs, affine initialization) and the 
+         calculating the Laplacian of inputs, affine initialization) and the
          core spatial normalization.
       2. Maps the brain mask into target space using the normalization
          calculated in 1.
@@ -216,7 +216,9 @@ def init_brain_extraction_wf(
     )
 
     trunc = pe.MapNode(
-        ImageMath(operation="TruncateImageIntensity", op2="0.01 0.999 256", copy_header=True),
+        ImageMath(
+            operation="TruncateImageIntensity", op2="0.01 0.999 256", copy_header=True
+        ),
         name="truncate_images",
         iterfield=["op1"],
     )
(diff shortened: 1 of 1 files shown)


niworkflows/anat/ants.py
```diff
@@ -340,47 +331,61 @@ def init_brain_extraction_wf(
...
-            ("out_file", "out_file"),
-            ("out_mask", "out_mask"),
-            ("bias_corrected", "bias_corrected"),
-            ("bias_image", "bias_image"),
-        ]),
+        (thr_brainmask, apply_mask, [("output_image", "in_mask")]),
+        (thr_brainmask, outputnode, [("output_image", "out_mask")]),
+        (inu_n4_final, outputnode, [("output_image", "bias_corrected"),
+                                    ("bias_image", "bias_image")]),
+        (apply_mask, outputnode, [("out_file", "out_file")]),
```
effigies, line 354: Should probably note that this is no longer an equivalent of `antsBrainExtraction.sh` in the workflow docstring. Ideally indicating the differences.

niworkflows/anat/ants.py
```diff
@@ -549,10 +587,12 @@ def init_atropos_wf(
 
     # massage outputs
     pad_segm = pe.Node(
-        ImageMath(operation="PadImage", op2="%d" % padding), name="02_pad_segm"
+        ImageMath(operation="PadImage", op2=f"{padding}", copy_header=False),
```
effigies, line 590: If ANTs is potentially screwing up the affines, should we just implement padding ourselves?

niworkflows/anat/ants.py
```diff
@@ -340,47 +323,37 @@ def init_brain_extraction_wf(
...
+        (res_target, init_aff, [("out_file", "moving_image")]),
         (init_aff, norm, [("output_transform", "initial_moving_transform")]),
         (norm, map_brainmask, [
             ("reverse_transforms", "transforms"),
             ("reverse_invert_flags", "invert_transform_flags"),
         ]),
         (map_brainmask, thr_brainmask, [("output_image", "input_image")]),
-        (thr_brainmask, dil_brainmask, [("output_image", "op1")]),
-        (dil_brainmask, get_brainmask, [("output_image", "op1")]),
+        (map_brainmask, inu_n4_final, [("output_image", "weight_image")]),
```
effigies, line 348: Having trouble tracing how this originally connected through...

niworkflows/anat/ants.py
oesteban: Yes, yes, yes :)

- #568

oesteban:
It seems this refactor still has some problems:

ds054/sub-100003:
Screen Shot 2020-09-19 at 9 28 47 AM

I might have to roll back the WM prior utilization and change it to using the brain mask from the template. Look at the remaining inhomogeneity around the cerebellum in ds054/sub-100057:
Screen Shot 2020-09-19 at 9 32 08 AM

niworkflows/anat/ants.py
oesteban: Although not ideal, this might actually help -> https://1942-74174878-gh.circle-artifacts.com/0/docs/_build/no_version_html/html/api/niworkflows.anat.ants.html#niworkflows.anat.ants.init_brain_extraction_wf

oesteban:
The latest commit includes changes to make the workflow more reliable overall, fixing the problems (at least for sub-100003) I detected on ds054.

I will re-run a good bunch of datasets, including the two originally identified as failing (ds005/sub-08 and ds051/sub-06) and ds054.

oesteban:
I have visually confirmed that after a05dddc, the posterior most overlapping with the WM prior is correctly selected for ds051/sub-06.

oesteban:
Okay, it seems to be working correctly now.

### ds005
Checked visually and all subjects (incl. 08) look great now.

### ds051
Checked visually and all subjects (incl. 06) look great now.

### ds054/sub-100003
Before:
Screen Shot 2020-09-19 at 9 28 47 AM

After:
Screen Shot 2020-09-22 at 8 43 10 AM

### ds054/sub-100057
Before:
Screen Shot 2020-09-19 at 9 32 08 AM

After:
Screen Shot 2020-09-22 at 8 48 33 AM

### ds054/sub-100042
Before:
Screen Shot 2020-09-22 at 8 51 45 AM

After:
Screen Shot 2020-09-22 at 8 52 09 AM

niworkflows/anat/ants.py
oesteban: I have edited the documentation of both workflows to note this.

oesteban:
This is ready for merge - final comments @effigies @mgxd ?

niworkflows/anat/ants.py
```diff
@@ -41,6 +40,8 @@
     "FLAIR": OrderedDict([("nclasses", 3), ("csf", 1), ("gm", 3), ("wm", 2)]),
 }
 
+_ants_version = Registration().version
```
effigies, line 43: This means `import niworkflows.anat.ants` (and possibly `niworkflows`) will try to run a subprocess. I think we should avoid things like that at import. We could keep track of whether we've checked the version with a flag:

```suggestion
_ants_version_checked = False
_ants_version = None
```

Alternately, you could do a one-time property, which returns the cached value after first call:

```Python
from nibabel.onetime import auto_attr
class _version_check:
    @auto_attr
    def ants(self):
        return Registration().version
```

niworkflows/anat/ants.py
effigies: Okay, so it looks like we're no longer using Atropos to select white matter at all. What purpose does it now serve?

niworkflows/anat/ants.py
oesteban: Resorted to copy Alejandro's pattern (and implicitly Nipypes memoization of versions) in the workflow for masked datasets a few lines below. Please let me know if that works.

niworkflows/anat/ants.py
oesteban: Yes, we still use ATROPOS:
- To refine the binary mask (as originally done by antsBrainExtraction.sh)
- To obtain the WM posterior (i.e., select the WM).

In some sense, now we use it even more :D 

However, it is a bit implicit - see the `disconnect()` call below, to replace the final outputs of the brain-extraction workflow with those coming from the ATROPOS workflow.

niworkflows/anat/ants.py
```diff
@@ -951,3 +1071,47 @@ def _conform_mask(in_mask, in_reference):
...
+def _matchlen(value, reference):
+    return [value] * len(reference)
+
+
+def _imsum(op1, op2, out_file=None):
+    import nibabel as nb
+
+    im1 = nb.load(op1)
+
+    data = im1.get_fdata() + nb.load(op2).get_fdata()
```
mgxd, line 1085: is 64bit precision necessary here?

niworkflows/anat/ants.py
```diff
@@ -951,3 +1071,47 @@ def _conform_mask(in_mask, in_reference):
...
+    nii.to_filename(out_file)
+    return out_file
+
+
+def _improd(op1, op2, in_mask, out_file=None):
+    import nibabel as nb
+
+    im1 = nb.load(op1)
+
+    data = im1.get_fdata() * nb.load(op2).get_fdata()
```
mgxd, line 1103: same as above

mgxd (commented):
This looks pretty good (looking at the before/after pictures)

niworkflows/anat/ants.py
oesteban: I would reverse the question and say, do we need to cast back to float32? after all these are 3D maps.

niworkflows/anat/ants.py
```diff
@@ -60,20 +58,31 @@ def init_brain_extraction_wf(
...
          calculated in 1.
-      3. Superstep 1b: smart binarization of the brain mask
-      4. Superstep 6: apply ATROPOS and massage its outputs
-      5. Superstep 7: use results from 4 to refine the brain mask
+      3. Superstep 1b: binarization of the brain mask
+      4. Maps the WM probability map from the template, if such prior exists.
+         Combines the BS (brainstem) probability map before mapping if the WM
+         and BS are given separately (as it is the case for ``OASIS30ANTs``.)
+      5. Run a second N4 INU correction round, using the prior mapped into
+         individual step in step 4.
```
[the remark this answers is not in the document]
oesteban: the brainmask is used https://github.com/nipreps/niworkflows/pull/567/files#diff-f15ba4e65fd25eb717a54bbd8f6759c2R392

niworkflows/anat/ants.py
```diff
@@ -979,11 +972,9 @@ def init_n4_only_wf(
...
         inu_n4_final.inputs.rescale_intensities = True
     except ValueError:
-        warn(
-            "The installed ANTs version too old. Please consider upgrading to "
-            "2.1.0 or greater.",
-            DeprecationWarning,
-        )
+        warn("N4BiasFieldCorrection's --rescale-intensities option was added in ANTS 2.1.0 "
+             f"({inu_n4_final.interface.version} found.) Please consider upgrading.",
+             DeprecationWarning)
```
effigies, line 977: `DeprecationWarning`s are for things we plan to deprecate. I think this should be a `UserWarning`.

oesteban:
Good to go?

oesteban:
Hey @bpinsard, I'd like to squish this into an RC2 before the LTS is cut (and preferably, cut the RC as soon as possible). Do you want to take a look into it before we go ahead? We would like to hear your opinion before merging since you'll be maintaining the LTS.

effigies (approved):
I think I roughly follow what happened here. No obvious problems.

oesteban:
I'll take @bpinsard's silence as no-opposition. The sooner we start testing this, the better.

PR #582, opened 2020-11-11
FIX: Allow omission of <res> for template normalization

mgxd:
Since templateflow templates are not required to have `<res>` identifiers, there should be a way to disable searching for one.

EDIT: This is a bit of a hacky solution, especially since `template_resolution` is deprecated. Open to other suggestions.


mgxd:
This is minor enough - merging

PR #584, opened 2020-11-12
Include _T2starw _MTw and _TSE-suffixes in build_path options

Gilles86:
Me again...

For my project I would like to store `_MTw`-, `_TSE`- and `T2starw`-data in my derivatives, using the `DerivativesDatasink`. however, `niworkflows` currently doesn't support that. Would it be possible to include these suffixes (waiting for BEP-001 to be finally accepted)?


effigies:
I can't see any harm in this. Let's do it.

PR #593, opened 2020-12-15
MNT: Drop Python 3.6, test setuptools builds, pip installations 

effigies:
As noted in other discussions, setuptools is aiming to be an installation backend, not an installer, and has no plans to implement various installation PEPs, such as respecting the `PythonRequires` metadata exposed by PyPI.

Therefore I think the recommended installation method needs to be pip. Since we distribute sdists and wheels, we should continue to test those, and those are built with `setuptools` and `wheel`, but we do not need to test against old versions.

We may want to test against old versions of `pip` again, but I think after a year or two of using `pyproject.toml`, we're pretty comfortable that pip can fetch installation dependencies as needed for this project and its dependencies.

### DIFF
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -23,16 +23,13 @@ jobs:
         
     steps:
     - name: Set up Python ${{ matrix.python-version }}
-      uses: actions/setup-python@v1
+      uses: actions/setup-python@v2
       with:
         python-version: ${{ matrix.python-version }}
     - uses: actions/checkout@v2
       with:
         ssh-key: "${{ secrets.NIPREPS_DEPLOY }}"
-    - name: Fetch all tags (for versioneer to work)
-      if: "!startsWith(github.ref, 'refs/tags/')"
-      run: |
-        /usr/bin/git -c protocol.version=2 fetch --tags --prune --unshallow origin
+        fetch-depth: 0
     - name: Build in confined, updated environment and interpolate version
       run: |
         python -m venv /tmp/buildenv
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -16,7 +16,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.6, 3.7, 3.8]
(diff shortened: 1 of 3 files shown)


oesteban:
Since Numpy is dropping Python 3.6, should we follow suit and make our lives easier?

(BTW, the failed test is unrelated to these changes, obviously - I'm getting a similar error on sdcflows proper)

effigies:
Yup, I'm okay with us following numpy's lead in [NEP29](https://numpy.org/neps/nep-0029-deprecation_policy.html).

For reference NiBabel and Nipype are on NEP29 + 1 year, and PyBIDS is just going with Python EOL at this point. But as mostly user-facing tools with containerized installations, there's no strong need for nipreps to outlast numpy.

oesteban:
The failing test is because scikit-image. Docker build log:

```
ERROR: scikit-image 0.18.0 has requirement numpy>=1.16.5, but you'll have numpy 1.15.4 which is incompatible.
```

Testing an update on nipreps/sdcflows#158

effigies:
Yeah, let's update the numpy/scipy/scikit-learn dependencies.

Should I use the versions you're using over there?

oesteban:
> Should I use the versions you're using over there?

I think so. I'm trying to find the least annoying way of installing scikit-image. It seems that pip will be the fastest option (conda-forge also attempts to change many package's versions/builds).

I have also bumped setuptools and pip inside the Dockerimage - we probably want to unpin setuptools and use the latest at the time is cleaned up.

oesteban:
I'm also thinking that we could be more flexible with pinning and do not force patch releases. That should make the installation less breakable (assuming packages do not break dependency tables on patch release changes, which seems fairly safe to assume).

Then, when releasing, we could clean up the conda layer to force a fresh installation (or have a cron job do this).

oesteban:
Okay, done on SDCFlows. Have the dependency be resolved by pip was the best option. Pinned:
numpy-1.19
scikit-learn-0.19
scipy-1.5

(https://github.com/nipreps/sdcflows/blob/7c81f25e90c79453570c27e9d00cac96f59d4f72/Dockerfile#L83-L102)

oesteban:
Do you want to include that in this PR?

PR #596, opened 2021-01-06
PIN: svgutils to previous versions

mgxd:
See https://github.com/nipreps/niworkflows/issues/595 for context

### DIFF
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -16,22 +16,17 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.6, 3.7, 3.8]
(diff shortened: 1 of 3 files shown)


effigies:
In retrospect, maybe just pinning !=0.3.2 would be safer. An == is fragile for a dependency instead of a user facing app.

mgxd:
Yeah, definitely more room for error. Do you feel strongly about changing it, or shall we just wait for a new svgutils release?

effigies:
I don't think it's urgent to do until they release, but I don't think we should release a library with an `==` dependency, so any niworkflows release process should include updating to `!=0.3.2`. (Since they're fixing now, I think we can safely ignore the possibility of a broken 0.3.3 or 0.4.0.)

Issue #600, opened 2021-01-22
Make ``EstimateReference`` modality-agnostic


oesteban:
Revise the interface so that it can be seamlessly used with BOLD or DWIs (or any other modality):

1. Split the interface into two steps: one interface for extracting/selecting the volumes (DWI orientations or BOLD timepoints). This interface has an input (`ref_list` or some name along the lines) that may be left undefined:
    - when `ref_list` is set, a list of bools will determine the volumes in the 4D file that will be extracted out as references (for DWI this is basically something like ``bvals < 100``, and for BOLD it could be something like ``np.zeros(shape4, dtype=bool)[:5] = True``)
    - when `ref_list` is `Undefined` then the nonsteady states detector we have is run and the list is generated internally.

2. Run the rest of ``EstimateReference`` on the short 4D file generated in 1), only that now the HMC step will be available to all branches (at this moment if the nonsteady states are to be detected, HMC cannot be run).
    - The interface outputs both the averaged volume and the aligned 4D files.
    - the interface outputs the list of affine matrices, and
    - the interface compensates for signal drifts/decay.

Assigning Mathias this issue.

Requesting feedback from @effigies, @mattcieslak, @eilidhmacnicol, @josephmje, @dPys.

PR #602, opened 2021-01-27
ENH: Add a ``RobustAverage`` interface and split volume selection

oesteban:
Resolves: #600.

### DIFF
--- a/niworkflows/interfaces/images.py
+++ b/niworkflows/interfaces/images.py
@@ -259,7 +259,7 @@ class RobustAverage(SimpleInterface):
                 f"Image length ({img_len} timepoints) unmatched by mask ({len(t_mask)})"
             )
 
-        n_volumes = np.sum(self.inputs.t_mask)
+        n_volumes = np.sum(t_mask)
         if n_volumes < 1:
             raise ValueError("At least one volume should be selected for slicing")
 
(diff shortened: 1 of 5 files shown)


oesteban:
I'll work on the interface that generates `t_list` tomorrow.

oesteban:
Okay, this is looking better. Please LMKWYT, @mgxd .

niworkflows/interfaces/images.py
```diff
@@ -188,6 +191,134 @@ def _run_interface(self, runtime):
...
+        t_mask = (
+            self.inputs.t_mask if isdefined(self.inputs.t_mask) else [True] * img_len
+        )
+
+        if len(t_mask) != img_len:
+            raise ValueError(
+                f"Image length ({img_len} timepoints) unmatched by mask ({len(t_mask)})"
+            )
+
+        n_volumes = np.sum(self.inputs.t_mask)
```
mgxd, line 262: what if this is undefined?
```suggestion
        n_volumes = np.sum(t_mask)
```

niworkflows/interfaces/images.py
```diff
@@ -188,6 +191,134 @@ def _run_interface(self, runtime):
...
+        if n_volumes < 1:
+            raise ValueError("At least one volume should be selected for slicing")
+
+        self._results["out_file"] = fname(suffix="_average")
+        self._results["out_volumes"] = fname(suffix="_sliced")
+
+        sliced = nb.concat_images(
+            i for i, t in zip(nb.four_to_three(img), self.inputs.t_mask) if t
```
mgxd, line 270:
```suggestion
            i for i, t in zip(nb.four_to_three(img), t_mask) if t
```

mgxd (approved):
Few comments, but this LGTM

niworkflows/interfaces/images.py
oesteban: Yup, this was the idea - thanks for catching it.

PR #617, opened 2021-02-09
ENH: carpet plot tweaks

notZaki:
This PR is a mix of fixes and enhancements to `viz.plot_carpet`. Most of the enhancements are based on personal preference so they can be dropped if desired.

**Fixes**

- Legend should now show up if `legend = True`
  + The legend argument was not getting passed from `carpet_plot()` to `_carpet()`
- Title should now show up if `title` argument is specified

**Enhancements**

- The `nskip` argument should now work for NIfTI inputs
  + Looks like it's trivial to implement this for CIFTI as well, but I don't use them, so I didn't touch it
- Tick formatting
  + If the xticks are frame numbers, then they will be printed as integers instead of decimals
  + Xticks now use the default font size
     + This prevents a clash with user-defined matplotlib configuration
- The input data can now be a nibabel-image. A path (as a string) is still acceptable.
  + This is convenient because I can pre-process the data and then plot it directly without the i/o overhead

### DIFF
--- a/niworkflows/viz/plots.py
+++ b/niworkflows/viz/plots.py
@@ -360,6 +360,8 @@ def _carpet(
     ax1.spines["bottom"].set_visible(False)
     ax1.spines["left"].set_color("none")
     ax1.spines["left"].set_visible(False)
+    if title:
+        ax1.set_title(title)
 
     ax2 = None
     if legend:
--- a/niworkflows/viz/plots.py
+++ b/niworkflows/viz/plots.py
@@ -151,7 +151,7 @@ def plot_carpet(
     ----------
 
         func : string or nibabel-image object
-            Path to NIfTI or CIFTI BOLD image, or a nibabel-image object 
+            Path to NIfTI or CIFTI BOLD image, or a nibabel-image object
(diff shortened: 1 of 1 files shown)


oesteban:
Sorry for our slow turn around @notZaki - I'll go through this next week.

niworkflows/viz/plots.py
```diff
@@ -150,16 +150,16 @@ def plot_carpet(
...
+            Path to NIfTI or CIFTI BOLD image, or a nibabel-image object
         atlaslabels: ndarray, optional
             A 3D array of integer labels from an atlas, resampled into ``img`` space.
             Required if ``func`` is a NIfTI image.
         detrend : boolean, optional
             Detrend and standardize the data prior to plotting.
         nskip : int, optional
             Number of volumes at the beginning of the scan marked as nonsteady state.
-            Not used.
+            Only used by volumetric NIfTI.
```
oesteban, line 162: @mgxd I'm realizing of an important issue with CIFTIs - are we dropping nonsteady states before sampling?

PR #620, opened 2021-02-11
FIX: Update svgutils after breaking API changes

mgxd:
Originally #619 (see https://github.com/nipreps/niworkflows/pull/619#issuecomment-777772466)

### DIFF
--- a/niworkflows/viz/utils.py
+++ b/niworkflows/viz/utils.py
@@ -367,6 +367,7 @@ def compose_view(bg_svgs, fg_svgs, ref=0, out_file="report.svg"):
 
 
 def _compose_view(bg_svgs, fg_svgs, ref=0):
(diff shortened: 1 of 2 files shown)


effigies:
The SVGs from before have expired. When they regenerate, could you post them here as attachments to the comment, or screenshots?

mgxd:
Essentially it seems to now be fixed, instead of dynamic based on window size.

master
this patch

effigies:
Hmm. Not ideal.

mgxd:
When comparing new/current images, it seemed the viewBox element was not being saved. Setting the height / width through svgutils' `Unit` fixes this, and images are correctly sized again. This should be good now.

effigies (approved):
LGTM. Thanks for tracking this down.

PR #623, opened 2021-02-27
FIX: TypeError - ``moveto()`` got an unexpected keyword argument ``'scale_x'``

oesteban:
Accounts for this small change in svgutils' interface.

### DIFF
--- a/niworkflows/viz/utils.py
+++ b/niworkflows/viz/utils.py
@@ -399,7 +399,7 @@ def _compose_view(bg_svgs, fg_svgs, ref=0):
 
     yoffset = 0
     for i, r in enumerate(roots):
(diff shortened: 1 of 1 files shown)


effigies:
This seems to be a reversion. What version are you seeing the issue on?

mgxd (commented):
`scale`has been replaced with `scale_x` and `scale_y` in later versions. 
https://github.com/btel/svg_utils/blob/7b3c9ef33b916173209ba23252c35d7cd7128bdd/src/svgutils/transform.py#L24

I don't think this is necessary since we now require `svgutils >= 0.3.4`

oesteban:
I'll double check, it is very possible I'm using an outdated svgutils

PR #626, opened 2021-04-09
FIX: Purge removed `--disable` flag from svgo call

mgxd:
svgo 2.x dropped support for CLI plugin alteration.
We could generate a JSON config with this option, but
the changes are so miniscule we can likely just drop this flag.

This PR also pins the svgo version in the Dockerfile to the 2.x series.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -100,7 +100,7 @@ WORKDIR /home/niworkflows
 ENV HOME="/home/niworkflows"
 
 # Installing SVGO
-RUN npm install -g svgo~2
+RUN npm install -g svgo@^2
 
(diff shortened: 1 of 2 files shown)


mgxd:
This is fairly straightforward, but would love a review before merging. I've compiled an svg generated via the current method and after this new PR at https://gist.github.com/mgxd/224df751db6670c682ae2c96dbce94b6, if it helps

effigies (approved):
LGTM. I can't see a visible difference. Have you checked that they embed the same in reports?

mgxd:
Just checked, nothing jumps out. Thanks for the review 👍

PR #629, opened 2021-04-14
FIX: Limit memory usage in EstimateReferenceImage

### LINKED ISSUE #628: MemoryError in `init_bold_reference_wf` `gen_ref`
@ilyaveer and I have encountered a MemoryError in `init_bold_reference_wf` with version 1.3.2, when the bold file is very large (multiband data).
```
MemoryError: Unable to allocate 4.23 GiB for an array with shape (97, 115, 97, 525) and data type float64
```
[crash-20210409-151708-yansch-gen_ref.a0-ee823608-df22-40e4-84d8-fb225777e3d8.txt](https://github.com/nipreps/niworkflows/files/6308806/crash-20210409-151708-yansch-gen_ref.a0-ee823608-df22-40e4-84d8-fb225777e3d8.txt)

After looking at the code, I believe the problem may be that `gen_ref` (1.3.2) and its successor `gen_avg` (1.4.0rc5) are coded to have a memory usage of at most one gigabyte. 
https://github.com/nipreps/niworkflows/blob/dafdb2e21242337527a2fd91a8e358f6a37a564c/niworkflows/func/util.py#L156-L158 https://github.com/nipreps/niworkflows/blob/616d2bca69c023a4144afe3d1117b9c3ef5ebbcd/niworkflows/func/util.py#L168

However, when running, the interface `RobustAverage` (and its predecessor `EstimateReferenceImage`) both load the entire bold_file into memory, which may in some cases use much more than one gigabyte.
https://github.com/nipreps/niworkflows/blob/616d2bca69c023a4144afe3d1117b9c3ef5ebbcd/niworkflows/interfaces/images.py#L227

As such, I was wondering whether it would be possible to make the `mem_gb` adaptive as is done in other places, so that memory errors can be avoided by the nipype scheduler. This would likely entail adding a parameter to `init_bold_reference_wf`. I would be happy to contribute code.

effigies:
Fixes #628. At least partially. If we're splitting and concatenating 900MB of data, it seems plausible we could double our usage to 1.8GB. Maybe should call it 2GB to be safe?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -100,7 +100,7 @@ WORKDIR /home/niworkflows
 ENV HOME="/home/niworkflows"
 
 # Installing SVGO
-RUN npm install -g svgo
+RUN npm install -g svgo@^2
 
 # Installing WEBP tools
 RUN curl -sSLO "http://downloads.webmproject.org/releases/webp/libwebp-0.5.2-linux-x86-64.tar.gz" && \
--- a/niworkflows/interfaces/registration.py
+++ b/niworkflows/interfaces/registration.py
@@ -484,7 +484,7 @@ class EstimateReferenceImage(SimpleInterface):
         # Build the nibabel spatial image we will work with
         ref_im = []
         for im_i in ref_input:
-            max_new_volumes = 40 - len(ref_im)
+            max_new_volumes = 50 - len(ref_im)
             if max_new_volumes <= 0:
                 break
             nib_i = nb.squeeze_image(nb.load(im_i))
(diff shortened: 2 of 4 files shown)


HippocampusGirl:
Awesome, thank you!

HippocampusGirl:
I'm not sure if you're already on that, but you may want to add the same check here for 1.4.0
https://github.com/nipreps/niworkflows/blob/616d2bca69c023a4144afe3d1117b9c3ef5ebbcd/niworkflows/interfaces/images.py#L269-L271

HippocampusGirl:
Although I guess there you will need the full image, so I may be wrong :-)

effigies:
Yeah, I don't know that we can avoid loading all the data, but we definitely don't need to be making so many copies.

effigies:
So this is only a smoke test, but it's more than we had. I'm going to go ahead and merge.

effigies:
1.3.3 released. Should hit PyPI in an hour or two.

Issue #611, opened 2021-02-05
Revise all N4 nodes and add adaptations for rodents


oesteban:
This conversation stems from https://github.com/nipreps/niworkflows/pull/610#discussion_r570443035, but a general pass through the whole codebase, identifying N4 nodes would be appropriate.

A more generalistic approach would be to extend Nipype's N4 interface and provide it with the extra logic to calculate the bspline distances.

eilidhmacnicol:
I've had a look at the codebase, and surprisingly there may only be one application of N4 that will be pulled for rodents (`init_epi_reference_wf`). Other examples of N4 correction are used in workflows unlikely to be used with rodents (`ants.init_brain_extraction_wf`, `ants.init_atropos_wf`, `ants.init_n4_only_wf`, `skullstrip.afni_wf`, and `util.init_bold_reference_wf`) because there are other changes that would be required for these workflows, and I would suggest that if these workflows are to be adapted that they should be located in nirodents. I will submit a PR with the change to the reference workflow, but interested to hear what @oesteban  thinks about whether there is any utility in adaption of these other workflows for rodent images.

PR #630, opened 2021-04-19
ENH: add rodent-specific adaption for B-spline

eilidhmacnicol:
Addressing #611, this PR 1) adds a function to determine bspline grid using a ratio of the number of slices in each dimension, and 2) adds logic to various workflows so that this might be used with rodents by updating the N4 parameters.  Not to say that the function can't be used with non-rodent images, but it hasn't been tested as such. 

There are other workflows that call `N4BiasFieldCorrection`, but they are unlikely to be used by NiRodents/fprodents because there are already rodent alternatives or because they would require more changes and would probably benefit from a rodent-specific workflow.

### DIFF
--- a/niworkflows/anat/ants.py
+++ b/niworkflows/anat/ants.py
@@ -507,8 +507,9 @@ def init_atropos_wf(
     padding=10,
     in_segmentation_model=tuple(ATROPOS_MODELS["T1w"].values()),
     bspline_fitting_distance=200,
+    use_bspline_grid=False,
+    min_n4_iter=False,
     wm_prior=False,
-    rodent=False,
 ):
     """
     Create an ANTs' ATROPOS workflow for brain tissue segmentation.
@@ -557,6 +558,12 @@ def init_atropos_wf(
         ``(4,4,2,3)`` uses K=4, CSF=4, GM=2, WM=3.
     bspline_fitting_distance : float
         The size of the b-spline mesh grid elements, in mm (default: 200)
+    use_bspline_grid : :obj:`bool`
+        If true, defines the number of b-spline mesh grid elements in each dimension rather
+        than using the isotropic distance given in ``bspline_fitting_distance``.
+    min_n4_iter : :obj:`bool`
+        If ``True``, reduces the number of b-spline iterations from 5 to 4. This use-case
+        is intended for rodents and other non-human/non-adult cases.
     wm_prior : :obj:`bool`
         Whether the WM posterior obtained with ATROPOS should be regularized with a prior
         map (typically, mapped from the template). When ``wm_prior`` is ``True`` the input
@@ -763,6 +770,7 @@ def init_atropos_wf(
         name="inu_n4_final",
         iterfield=["input_image"],
     )
+    inu_n4_final.inputs.n_iterations = [50] * (5 - min_n4_iter)
 
     try:
         inu_n4_final.inputs.rescale_intensities = True
@@ -875,13 +883,9 @@ def init_atropos_wf(
         ])
         # fmt: on
 
-    # rodent-specific N4 settings
-    if rodent:
+    if use_bspline_grid:
+        # set INU bspline grid based on image shape
         from ..utils.images import _bspline_grid
-        # fewer iterations of N4
-        inu_n4_final.inputs.n_iterations = [50] * 4
-
-        # set INU bspline grid based on voxel size
         bspline_grid = pe.MapNode(niu.Function(function=_bspline_grid), name="bspline_grid")
 
         # fmt:off
@@ -891,7 +895,7 @@ def init_atropos_wf(
         ])
         # fmt:on
     else:
-        inu_n4_final.inputs.n_iterations = [50] * 5
+        # set INU bspline grid based on isotropic distance
         inu_n4_final.inputs.bspline_fitting_distance = bspline_fitting_distance
 
     return wf
@@ -902,7 +906,8 @@ def init_n4_only_wf(
     atropos_refine=True,
     atropos_use_random_seed=True,
     bids_suffix="T1w",
-    rodent=False,
+    use_bspline_grid=False,
+    min_n4_iter=False,
     mem_gb=3.0,
     name="n4_only_wf",
     omp_nthreads=None,
@@ -943,6 +948,11 @@ def init_n4_only_wf(
     atropos_model : tuple or None
         Allows to specify a particular segmentation model, overwriting
         the defaults based on ``bids_suffix``
+    use_bspline_grid : :obj:`bool`
+        If true, defines the number of b-spline mesh grid elements in each dimension.
+    min_n4_iter : :obj:`bool`
+        If ``True``, reduces the number of b-spline iterations from 5 to 4. This use-case
+        is intended for rodents and other non-human/non-adult cases.
     name : str, optional
         Workflow name (default: ``'n4_only_wf'``).
 
@@ -1005,15 +1015,14 @@ def init_n4_only_wf(
             dimension=3,
             save_bias=True,
             copy_header=True,
-            n_iterations=[50] * 5,
             convergence_threshold=1e-7,
             shrink_factor=4,
-            bspline_fitting_distance=200,
         ),
         n_procs=omp_nthreads,
         name="inu_n4_final",
         iterfield=["input_image"],
     )
+    inu_n4_final.inputs.n_iterations = [50] * (5 - min_n4_iter)
 
     # Check ANTs version
     try:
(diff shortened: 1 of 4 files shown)


niworkflows/utils/images.py
```diff
@@ -297,3 +297,20 @@ def nii_ones_like(in_file, value, dtype, newpath=None):
     nii.to_filename(out_file)
 
     return out_file
+
+
+def _bspline_grid(in_file):
```
eilidhmacnicol, line 302: Inspiration for using slice number to determine grid sparsity: 
https://itk.org/pipermail/community/2014-February/005036.html

niworkflows/anat/ants.py
```diff
@@ -863,6 +886,8 @@ def init_n4_only_wf(
     atropos_refine=True,
     atropos_use_random_seed=True,
     bids_suffix="T1w",
+    use_bspline_grid=False,
+    min_n4_iter=False,
```
oesteban, line 890: Do we use this workflow in nirodents? I don't think so, and I'd honestly avoid changing it at this moment.

niworkflows/utils/images.py
oesteban: Please add the link to the function's documentation

niworkflows/utils/images.py
```diff
@@ -297,3 +297,20 @@ def nii_ones_like(in_file, value, dtype, newpath=None):
...
+    """Estimate B-Spline fitting distance grid using the number of slices of ``in_file``."""
+    import nibabel as nb
+    import numpy as np
+    import math
+
+    img = nb.load(in_file)
+    zooms = img.header.get_zooms()[:3]
+    # find extent of each dimension wrt voxel sizes
+    extent = (np.array(img.shape[:3]) - 1) * zooms
```
oesteban, line 311:
```suggestion
    extent = np.array(img.shape[:3]) * zooms
```

oesteban: extent is clear in definition - doesn't suffer from the fencepost problem

oesteban (commented):
A few suggestions

eilidhmacnicol:
Okay, tested the N4 in the reference workflow on two sets of multi-run EPIs from a subject, and it works well.
Although I was going to develop some tests for the `bspline_grid` function, I would quite like to push on with integrating the reference workflow into fmriprep-rodents. 

@oesteban what do you think?

eilidhmacnicol:
`test_pytest` seems to be failing because of the pandas versioning rather than anything to do with the code, FYI

oesteban:
I'll have a look ASAP

niworkflows/anat/ants.py
```diff
@@ -735,15 +743,14 @@ def init_atropos_wf(
...
-            n_iterations=[50] * 5,
             convergence_threshold=1e-7,
             shrink_factor=4,
-            bspline_fitting_distance=bspline_fitting_distance,
         ),
         n_procs=omp_nthreads,
         name="inu_n4_final",
         iterfield=["input_image"],
     )
+    inu_n4_final.inputs.n_iterations = [50] * n4_iter
```
oesteban, line 753: why split this setting, instead of writing it above?

niworkflows/anat/ants.py
```diff
@@ -855,6 +862,22 @@ def _argmax(in_dice):
...
+
+        # fmt:off
+        wf.connect([
+            (inputnode, bspline_grid, [(("in_files", _pop), "in_file")]),
+            (bspline_grid, inu_n4_final, [("out", "args")])
+        ])
+        # fmt:on
+    else:
+        # set INU bspline grid based on isotropic distance
+        inu_n4_final.inputs.bspline_fitting_distance = bspline_fitting_distance
```
oesteban, line 879:
```suggestion
```

You can restore line 741 to pick `bspline_fitting_distance` as default, and the connection will override the value when `adaptive_bspline_grid` is true.

niworkflows/utils/images.py
```diff
@@ -297,3 +297,21 @@ def nii_ones_like(in_file, value, dtype, newpath=None):
     nii.to_filename(out_file)
 
     return out_file
+
+
+def _bspline_grid(in_file):
+    """Estimate B-Spline fitting distance grid using the number of slices of ``in_file``.
```
oesteban, line 303:
```suggestion
    """
    Estimate B-Spline fitting distance grid using the number of slices of ``in_file``.

```

niworkflows/utils/images.py
```diff
@@ -297,3 +297,21 @@ def nii_ones_like(in_file, value, dtype, newpath=None):
...
+    Using slice number to determine grid sparsity is inspired by the conversation found at
+    https://itk.org/pipermail/community/2014-February/005036.html"""
```
oesteban, line 305:
```suggestion
    https://itk.org/pipermail/community/2014-February/005036.html
    
    """
```

niworkflows/workflows/epi/refmap.py
```diff
@@ -64,6 +66,12 @@ def init_epi_reference_wf(
         If ``True``, determines nonsteady states in the beginning of the timeseries
         and selects them for the averaging of each run.
         IMPORTANT: this option applies only to BOLD EPIs.
+    adaptive_bspline_grid : :obj:`bool`
+        If ``True``, determines the number of b-spline grid elements from data shape
+        and feeds them into N4BiasFieldCorrection, rather than setting an isotropic distance.
+    n4_iter : :obj:`int`
+        The number of b-spline iterations (default: 5). Fewer (e.g. 4) are recommended
```
oesteban, line 73:
```suggestion
        The number of B-Spline fitting iterations (default: 5). Fewer (e.g. 4) are recommended
```

niworkflows/workflows/epi/refmap.py
```diff
@@ -64,6 +66,12 @@ def init_epi_reference_wf(
         If ``True``, determines nonsteady states in the beginning of the timeseries
         and selects them for the averaging of each run.
         IMPORTANT: this option applies only to BOLD EPIs.
+    adaptive_bspline_grid : :obj:`bool`
+        If ``True``, determines the number of b-spline grid elements from data shape
```
oesteban, line 70:
```suggestion
        If ``True``, determines the number of B-Spline grid elements from data shape
```

niworkflows/workflows/epi/refmap.py
```diff
@@ -202,6 +210,21 @@ def _set_threads(in_list, maximum):
     else:
         wf.connect(inputnode, "t_masks", per_run_avgs, "t_mask")
 
+    # rodent-specific N4 settings
+    n4_avgs.inputs.n_iterations = [50] * n4_iter
```
oesteban, line 214: as above, why not setting on line 138?

niworkflows/anat/ants.py
```diff
@@ -536,6 +538,12 @@ def init_atropos_wf(
         ``(4,4,2,3)`` uses K=4, CSF=4, GM=2, WM=3.
     bspline_fitting_distance : float
         The size of the b-spline mesh grid elements, in mm (default: 200)
+    adaptive_bspline_grid : :obj:`bool`
+        If true, defines the number of b-spline mesh grid elements in each dimension rather
```
oesteban, line 542:
```suggestion
        If true, defines the number of B-Spline mesh grid elements in each dimension rather
```

niworkflows/anat/ants.py
```diff
@@ -536,6 +538,12 @@ def init_atropos_wf(
...
+        than using the isotropic distance given in ``bspline_fitting_distance``.
+    n4_iter : :obj:`int`
+        The number of b-spline iterations (default: 5). Fewer (e.g. 4) are recommended
```
oesteban, line 545:
```suggestion
        The number of B-Spline fitting iterations (default: 5). Fewer (e.g. 4) are recommended
```

oesteban (approved):
Looking good - left a few minor comments and nit picks

oesteban:
> `test_pytest` seems to be failing because of the pandas versioning rather than anything to do with the code, FYI

I just rebased your branch on `upstream/master` where the problem has been recently addressed.

oesteban:
I also have changed the target branch to `feat/infants-rodents` so that we don't mess up the release of fMRIPrep 21.0.x

eilidhmacnicol:
@oesteban seems like the issue that was causing the tests to fail was related to the nipype version. Pinning to master has bypassed this, so is this ready to merge or should we pin to a newer version of nipype first?

eilidhmacnicol:
More fun - it seems that whatever is causing the tests to fail has been resolved somewhere between the release of nipype 1.7.0 (https://github.com/nipy/nipype/tree/b38572030f3d968c945675efaccd053d123f158b; released 28 days ago at time of writing) and the version accessed today (https://github.com/nipy/nipype/tree/f756b71117493caea093bac79d576207d8f96482). 

My guess is that it is fixed by nipy/nipype#3395, so I guess we will have to pin to 1.7.1 when it is released.

eilidhmacnicol:
In the meantime, I think I will revert to pinning to master and push through so we are not waiting solely on a nipype release

PR #643, opened 2021-06-11
FIX: Improve mask contour visibility

mgxd:
Changes the mask outline from blue to red, for increased contrast.

coreg_t2w
Screen Shot 2021-06-11 at 4 53 23 PM

### DIFF
--- a/niworkflows/viz/utils.py
+++ b/niworkflows/viz/utils.py
@@ -367,7 +367,7 @@ def plot_registration(
             display.add_contours(white, colors="b", **kwargs)
             display.add_contours(pial, colors="r", **kwargs)
         elif contour is not None:
-            display.add_contours(contour, colors="b", levels=[0.5], linewidths=0.5)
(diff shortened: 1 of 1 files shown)


niworkflows/viz/utils.py
```diff
@@ -367,7 +367,7 @@ def plot_registration(
             display.add_contours(white, colors="b", **kwargs)
             display.add_contours(pial, colors="r", **kwargs)
         elif contour is not None:
-            display.add_contours(contour, colors="b", levels=[0.5], linewidths=0.5)
+            display.add_contours(contour, colors="r", levels=[0.5], linewidths=0.5)
```
oesteban, line 370: this will clobber the pial (line 368), can we find some better color?

mgxd: those should be mutually exclusive, since it's an `if elif` block no?

oesteban (approved):

PR #645, opened 2021-07-29
[FIX] `pandas` warning about use of keyword arguments

HippocampusGirl:
With the most recent `pandas` version, I am getting the following warning:
```
/usr/local/miniconda/lib/python3.7/site-packages/niworkflows/interfaces/utils.py:896: FutureWarning: In a future version of pandas all arguments of read_csv except for the argument 'filepath_or_buffer' will be keyword-only
```
I have attached the required change to avoid it.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -123,7 +123,7 @@ RUN conda install -y python=3.7.1 \
                      mkl-service \
                      numpy=1.19 \
                      scipy=1.5 \
-                     scikit-learn=0.19 \
+                     'scikit-learn>=0.21' \
                      matplotlib=2.2.2 \
                      pandas=0.23.4 \
                      libxml2=2.9.8 \
--- a/Dockerfile
+++ b/Dockerfile
@@ -123,9 +123,9 @@ RUN conda install -y python=3.7.1 \
                      mkl-service \
                      numpy=1.19 \
                      scipy=1.5 \
-                     'scikit-learn>=0.21' \
+                     scikit-learn=0.21 \
                      matplotlib=2.2.2 \
-                     pandas=0.23.4 \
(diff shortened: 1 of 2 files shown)


oesteban (approved):
Thanks!

effigies:
I think the keyword is delimiter? I can't open the failing test on my phone, though.

HippocampusGirl:
As far as I know, `pandas` supports both the `sep` and `delimiter` keyword arguments. 

I just had a look, the error is something else. 
```
E   ImportError: A sklearn version of at least 0.21 is required to use nilearn. 0.19.2 was found. Please upgrade sklearn
```

effigies:
Got it. So we need to update our Dockerfile around here: https://github.com/nipreps/niworkflows/blob/master/Dockerfile#L126

HippocampusGirl:
Should I submit a separate PR for the `Dockerfile` update?

oesteban:
I think we can use this PR :)

Dockerfile
```diff
@@ -123,7 +123,7 @@ RUN conda install -y python=3.7.1 \
                      mkl-service \
                      numpy=1.19 \
                      scipy=1.5 \
-                     scikit-learn=0.19 \
+                     'scikit-learn>=0.21' \
```
oesteban, line 126:
```suggestion
                     scikit-learn=0.21 \
```
We typically pin a particular minor in the Dockerfile for the trade-off between flexibility (of new patch releases fixing something) and reproducibility of the environment (expecting constant APIs through the same minor of the software)

HippocampusGirl: Got it! Will commit this together with the `pandas` upgrade that `nilearn` apparently also requires :-)

PR #649, opened 2021-08-30
FIX: Set slope and intercept to 1/0 if not otherwise provided

effigies:
Test fails prior to fix. Required updating checksums.

This will fix nipreps/fmriprep#2507 and should be considered high priority.


niworkflows/interfaces/bids.py
```diff
@@ -648,6 +648,10 @@ def _run_interface(self, runtime):
                 if new_data is None:
                     set_consumables(new_header, orig_img.dataobj)
                     new_data = orig_img.dataobj.get_unscaled()
+                else:
+                    # Without this, we would be writing nans
+                    # This is our punishment for hacking around nibabel defaults
```
mgxd, line 653: 😆

mgxd (approved):

PR #650, opened 2021-08-31
ENH: Allow plotting of image rotated to cardinal axes

mgxd:
Follows #544 

Allows `SimpleBeforeAfter` reports to rotate images with respect to the cardinal axes. This is motivated in the context of SDC correction, to give a clearer look at the distortions.

Screen Shot 2021-08-31 at 3 43 06 PM
Screen Shot 2021-08-31 at 3 43 15 PM

### DIFF
--- a/niworkflows/interfaces/reportlets/base.py
+++ b/niworkflows/interfaces/reportlets/base.py
@@ -52,6 +52,7 @@ class RegistrationRC(reporting.ReportCapableInterface):
     _fixed_image_label = "fixed"
     _moving_image_label = "moving"
     _contour = None
+    _dismiss_affine = False
 
     def _generate_report(self):
         """Generate the visual report."""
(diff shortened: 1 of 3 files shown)


niworkflows/interfaces/reportlets/base.py
```diff
@@ -52,6 +52,7 @@ class RegistrationRC(reporting.ReportCapableInterface):
     _fixed_image_label = "fixed"
     _moving_image_label = "moving"
     _contour = None
+    _rotate_to_canonical = False
```
oesteban, line 55:
```suggestion
    _dismiss_affine = False
```

I think I would be more explicit about not considering the affine.

oesteban (approved):
Just a nomenclature nit-pick, but looks good to me. Thanks for this.

PR #656, opened 2021-10-14
DOC: Overhaul - new skin, new deployment & multiversion archival

oesteban:
This PR corresponds to the manual update to the `gh-pages` I've done (and currently available at https://nipreps.org/niworkflows).

Inherently, I'm proposing the following changes:

- Only the tip of each MAJOR.MINOR version series is archived. If we roughly follow our own conventions for the patch releases, this should not be a problem. I think the gains in clarity (without uneven numbers of patch releases clouding the choice) and space (we only need to store one version per series) outweigh any caveat of not having access to older patch release docs.
- Drop the `sphinxcontrib-versioning` package - we just need to be sure we do not ever wipe `gh-pages` out. Now, the old versions are archived and only overwritten if a new patch release of that series is issued. The responsibility of these updates is moved to GH Actions (work in progress).
- For the version chooser widget to work with future versions that will be uploaded, [it has been modified to now be dynamic](https://github.com/nipreps/niworkflows/commit/2ab80ad74eb6ecdbc51094468d171c79a6520822) and read an index of versions from a JSON file ([`versions.json`](https://github.com/nipreps/niworkflows/blob/gh-pages/versions.json)).

I'm also changing the skin of the documentation with the furo theme. It felt much cleaner than the RTD skin, and turned out to be easy to customize to include the version switch right below the project name.

I'm tagging @rwblair as a reviewer because he had quite a bit of time struggling with the `sphinxcontrib-versioning` package and the overall setup - his feedback could be very relevant, and at the very least, he might be interested too in the solution.


.github/workflows/docs-build-update.yml
```diff
@@ -0,0 +1,77 @@
...
+
+        # Remove forward slashes
+        CURBRANCH=$( echo $CURBRANCH | sed 's+/+_+g' )
+        echo "Building branch/tag ${CURBRANCH:-<unkwown>}, from git ref <$GITHUB_REF>"
+        echo "CURBRANCH=${CURBRANCH}" >> ${GITHUB_ENV}
+
+    - name: Install GraphViz
+      run: |
+        sudo apt-get update -y
+        sudo apt-get install -y graphviz
```
effigies, line 36:
```suggestion
        sudo apt-get install -y --no-install-recommends graphviz
```

.github/workflows/docs-build-update.yml
```diff
@@ -0,0 +1,77 @@
...
+    branches: [ 'doc/*', 'docs/*', master, "maint/*" ]
+    tags: [ '*' ]
+
+jobs:
+  build:
+    if: "!contains(github.event.head_commit.message, '[skip ci]')"
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
```
effigies, line 15: Need to get other branches.

```suggestion
    - uses: actions/checkout@v2
      with:
        fetch-depth: 0
```

docs/Makefile
```diff
@@ -57,9 +57,9 @@ docs-coverage:
 	PYTHONPATH=$(PYTHONPATH) $(SPHINXBUILD) -b coverage $(ALLSPHINXOPTS) $(BUILDDIR)/coverage
 
 html:
-	PYTHONPATH=$(PYTHONPATH) $(SPHINXBUILD) -b html $(ALLSPHINXOPTS) $(BUILDDIR)/html
+	PYTHONPATH=$(PYTHONPATH) $(SPHINXBUILD) -b html $(ALLSPHINXOPTS) $(BUILDDIR)/$(CURBRANCH)
```
effigies, line 60: Posted this on sdcflows by accident:

> This is going to make it trickier to rerender older versions at need. Or rather, it will create a watershed where we do one thing to render before and another thing to render after. Even if we can't get a single "correct" result now, I would suggest that we leave this rule unchanged and create a new rule like `htmlbranch`.

oesteban:
Okay, I would need to unlock myself from this one - so if I didn't really address the open question about the `Makefile` @effigies, let's go back to it after merge. I really hope that's fine :)

PR #664, opened 2021-11-02
ENH: Avoid appending packagename to `Report` outputs

mgxd:
Now that YODA is the default in `fmriprep`/`nibabies`, lets remove the hacks:

https://github.com/nipreps/fmriprep/blob/4795349af994e3fa62331c1f6e89d7e8b829ba04/fmriprep/reports/core.py#L26-L47
https://github.com/nipreps/nibabies/blob/55143215eaebca4924cefdc74d8008a347efea16/nibabies/reports/core.py#L5-L22

### DIFF
--- a/niworkflows/reports/core.py
+++ b/niworkflows/reports/core.py
@@ -245,13 +245,13 @@ class Report:
     .. doctest::
 
     >>> robj = Report(testdir / 'out', 'madeoutuuid', subject_id='01', packagename='fmriprep',
-    ...               reportlets_dir=testdir / 'work' / 'reportlets')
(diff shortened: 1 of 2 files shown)


mgxd:
@oesteban has `dmriprep` moved on to YODA outputs?

is there another package I'm missing that this will affect?

oesteban:
> @oesteban has `dmriprep` moved on to YODA outputs?

No, it hasn't.

> is there another package I'm missing that this will affect?

fMRIPrep-rodents (still under `poldracklab/`). cc/ @eilidhmacnicol

oesteban:
I think this is already done in nireports - should we just update everything not to use niworkflows?

PR #666, opened 2021-11-04
MNT: Configure black to avoid excessive rewriting

effigies:
Any objection to these configurations? By using a higher line length and not changing all quotes to double-quote, we can keep the sheer amount of rewrites down.

Would like to bring this over to fMRIPrep and the other tools, too.


oesteban:
> Any objection to these configurations?

None at all.

PR #669, opened 2021-11-15
CI: Clear docker cache

effigies:
Attempting to fix regression test failure that can't be reproduced locally.


mgxd:
this probably can be closed now?

PR #674, opened 2021-12-08
TEST: Drop excessively long interface equivalence tests

effigies:
We know that our method runs faster and uses less memory than nilearn's, so let's take the fastest one and just ensure that we don't lose equivalence.


mgxd (commented):
How much time is this saving?

effigies:
A few minutes.

mgxd (approved):

PR #678, opened 2022-01-24
ENH: Add optional session filter when collecting data

mgxd:
A slight alteration to provide session level control.

### DIFF
--- a/niworkflows/utils/bids.py
+++ b/niworkflows/utils/bids.py
@@ -26,6 +26,7 @@ import json
 import re
 import warnings
 from bids import BIDSLayout
+from bids.layout import Query
 from packaging.version import Version
 
 
@@ -149,7 +150,7 @@ def collect_participants(
 def collect_data(
     bids_dir,
     participant_label,
-    session_id=False,
+    session_id=Query.ANY,
     task=None,
     echo=None,
     bids_validate=True,
@@ -168,8 +169,8 @@ def collect_data(
         The BIDS directory
     participant_label : :obj:`str`
         The participant identifier
-    session_id : :obj:`str`, None, or ``False``
-        The session identifier. If ``False``, all sessions will be used (default).
+    session_id : :obj:`str`, None, or :obj:`bids.layout.Query`
+        The session identifier. By default, all sessions will be used.
     task : :obj:`str` or None
         The task identifier (for BOLD queries)
     echo : :obj:`int` or None
@@ -235,15 +236,12 @@ def collect_data(
     if echo:
         queries["bold"]["echo"] = echo
 
-    if session_id is not False:
-        for acq in queries.keys():
-            queries[acq]["session"] = session_id
-
     subj_data = {
         dtype: sorted(
             layout.get(
                 return_type="file",
                 subject=participant_label,
+                session=session_id,
                 extension=[".nii", ".nii.gz"],
                 **query,
             )
--- a/niworkflows/utils/bids.py
+++ b/niworkflows/utils/bids.py
@@ -149,6 +149,7 @@ def collect_participants(
 def collect_data(
     bids_dir,
     participant_label,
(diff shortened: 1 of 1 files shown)


niworkflows/utils/bids.py
```diff
@@ -209,6 +227,10 @@ def collect_data(
     if echo:
         queries["bold"]["echo"] = echo
 
+    if session_id is not False:
+        for acq in queries.keys():
+            queries[acq]["session"] = session_id
```
mgxd, line 232: Since `bids_filters` also updating the default entities, I'm not sure which of these should have the final say..

niworkflows/utils/bids.py
```diff
@@ -145,6 +145,7 @@ def collect_participants(
 def collect_data(
     bids_dir,
     participant_label,
+    session_id=False,
```
mgxd, line 148: since 
```
session = None
```
is a valid way to differentiate within pybids, this defaults to `False`, which drops the `session` entity from the query entirely.

effigies: What about using PyBIDS query semantics?

```suggestion
    session_id=bids.layout.Query.ANY,
```

niworkflows/utils/bids.py
effigies: If we put `session` directly in the `layout.get` call like with `subject`, then we can just let `**query` raise an exception if `session` is multiply defined.

effigies (approved):

mgxd:
@effigies looks like using `Query.ANY` is filtering out the files...is this a pybids bug?

mgxd:
Actually, I think this is [expected](https://github.com/bids-standard/pybids/blob/23553339dda7537752ace92d6c99a3f523e06ee2/bids/layout/layout.py#L577-L579)

> * Query.NONE: The named entity must not be defined.
> * Query.ANY: the named entity must be defined, but can have any \
> value.

Perhaps there needs to be another value for Query that mimics ANY, but allows undefined entities as well.

effigies:
Oh damn.

         filters : dict
            Any optional key/values to filter the entities on.
            Keys are entity names, values are regexes to filter on. For
            example, passing filters={'subject': 'sub-[12]'} would return
            only files that match the first two subjects. In addition to
            ordinary data types, the following enums are defined (in the
            Query class):
                * Query.NONE: The named entity must not be defined.
                * Query.ANY: the named entity must be defined, but can have any
                    value.

I guess we don't currently have a query enum that would do what I was thinking. I guess we need to revert to what you had, but let's open an issue on PyBIDS to allow a query that means it can be present or absent (maybe `Query.OPTIONAL`?).

niworkflows/utils/bids.py
```diff
@@ -145,6 +146,7 @@ def collect_participants(
 def collect_data(
     bids_dir,
     participant_label,
+    session_id=Query.ANY,
```
oesteban, line 149:
```suggestion
    session_id=None,
```

niworkflows/utils/bids.py
```diff
@@ -214,6 +233,7 @@ def collect_data(
             layout.get(
                 return_type="file",
                 subject=participant_label,
+                session=session_id,
```
oesteban, line 236:
```suggestion
                session=Query.ANY if session_id is None else session_id,
```

oesteban (approved):
As per the new docstring, `session_id` can be `None`. If so, it would be better to make the value replacement the latest possible, rather than the default parameter.

oesteban:
> let's open an issue on PyBIDS to allow a query that means it can be present or absent (maybe `Query.OPTIONAL`?).

Perhaps `Query.EITHER` (somewhat closer to enthought traits)?

effigies:
https://github.com/bids-standard/pybids/pull/809

PR #679, opened 2022-02-03
FIX: Set pixdim[4] to match RepetitionTime

effigies:
Via https://neurostars.org/t/modifying-nifti-files-before-fmriprep/21340/4, a dataset with `RepetitionTime == 2`, `pixdim[4] == 2000` and `xyzt_units & NIFTI_UNITS_MSEC == NIFTI_UNITS_MSEC` has its outputs written out with `RepetiionTime == 2`, `pixdim[4] == 2000` and `xyzt_units & NIFTI_UNITS_SEC == NIFTI_UNITS_SEC`.

This fix ensures that `RepetitionTime` gets written to `pixdim[4]` for BOLD series. This should avoid any conflicts with other 4+D data that doesn't have second units.

@arokem @a3sha2 Are there any similar considerations for dMRI/ASL that we should fix while we're here?


a3sha2:
@effigies ASL is similar to fmri (or bold) this will work

PR #680, opened 2022-02-03
ENH: Additional nibabel interfaces

mgxd:
Porting over interfaces created within nibabies


mgxd:
I've separated adding each interface by commit, which should help review this. After this is in, I'd like to cut a new version to allow a new release of nibabies

PR #681, opened 2022-02-07
FIX: Scipy docs path

mgxd:


mgxd:
this should probably just wait for scipy docs to become stable again

PR #682, opened 2022-02-08
RF: Move BIDS skeleton generation into new `testing` module

mgxd:
`tests/` directories are not generally installed, so this functionality was unaccessible.


effigies:
Maybe this needs rebasing?

PR #651, opened 2021-09-30
ENH: Initiate a module of nipype interfaces for mathematical morphology

celprov:
This pull request augments the existing carpet plot by adding to it the signals from the voxels at the edge of the brain.
The new code contains both the computation of the crown mask and its insertion in the carpet plot.

Furthermore, an interface computing the crown mask has been added in the purpose of being used in fMRIPrep and MRIQC.

Last, a legend of the colorbar separating regions in the carpet plot has been adding therefore closing nipreps/mriqc#870.

IMPORTANT ! : This piece of code depends on the change in the fMRIprep repository commited under the pull request https://github.com/nipreps/fmriprep/pull/2621 and on the change in the MRIQC repository under the pull request nipreps/mriqc#968

PR #684, opened 2022-02-28
ENH : Clustering of carpet plot rows

celprov:
This PR implements the reordering of the rows such that voxels with similar BOLD dynamics are placed close to one another. 
One can choose between two clustering methods using the flag `plot_carpet(ward=False/True)`: Ward’s hierarchical clustering or hierarchical average linkage clustering. As Nilearn recommanded ward's algorithm for clustering, the latter is set as default.
The clustering is done on euclidean distances between pairs of z-scored time-series within each regional block (crown, cortical GM, subcortical GM ...). This reordering follows the method developed by Aquino et al. 2020 to reveal non-global structures present in the fMRI signal.

IMPORTANT : This code builds on the PR #651 and as such, depends on the change in the fMRIprep repository commited under the pull request nipreps/fmriprep#2621 and on the change in the MRIQC repository under the pull request nipreps/mriqc#968


oesteban:
> IMPORTANT : This code builds on the PR https://github.com/nipreps/niworkflows/pull/651 and as such, depends on the change in the fMRIprep repository commited under the pull request https://github.com/nipreps/fmriprep/pull/2621 and on the change in the MRIQC repository under the pull request https://github.com/nipreps/mriqc/pull/968

Why is such a long dependency chain necessary here? We should actually leverage that this ordering could be done independently of whether you have the crown mask or not. That would make this PR easier to check, and should generalize naturally to the crown when both PRs have been merged.

oesteban:
Please continue in the other PR, which is based off of master.

PR #685, opened 2022-03-01
ENH : separate shallow and deep WM+CSF in the carpetplot

celprov:
This PR implements the proposition of @effigies (https://github.com/nipreps/niworkflows/pull/651#issuecomment-961103549) to separate the WM+CSF compartement in the carpetplot into the shallow WM+CSF and the deep WM+CSF corresponding to the aCompCor mask.
This separation is possible in fMRIprep since the aCompCor mask is computed, however the latter is not calculated in MRIQC, thus the default behavior is `acompcor_mask=None`. This means that the separation is not implemented when niworkflows is run by MRIQC.

IMPORTANT : This piece of code depends on the change in the fMRIprep repository commited under the pull request nipreps/fmriprep#2726.


oesteban:
Once #687 is merged, there should not be any changes in niworkflows necessary to accomplish what the title of this PR states. Instead, the higher-level app (in this case *fMRIPrep* see nipreps/fmriprep#2726) is responsible for passing in a `segments` dictionary where the necessary regions (split deep and shallow WM and CSF) are indexed.

PR #686, opened 2022-03-01
ENH: Reorder within-segment rows of carpetplot with hierarchical clustering

oesteban:
Simple implementation of reordering of rows in the carpetplot using hierarchical clustering. Supersedes #684.


oesteban:
Seems to be working.

Before:

After:

PR #688, opened 2022-03-03
ENH: Miscellaneous improvements to carpetplot

oesteban:
Several minor improvements to carpets:
- [x] Separation between ROIs in the plot - introduces small separation bands to more clearly see the segments.
- [x] Simplify code with better use of matplotlib
- [x] Add ROI labels to the Y axis.

## Examples
NIfTI with segments:

NIfTI without segments:

CIFTI2:

### DIFF
--- a/niworkflows/viz/plots.py
+++ b/niworkflows/viz/plots.py
@@ -29,7 +29,7 @@ import pandas as pd
 import matplotlib.pyplot as plt
 from matplotlib import gridspec as mgs
 import matplotlib.cm as cm
-from matplotlib.colors import ListedColormap, Normalize
+from matplotlib.colors import Normalize
 from matplotlib.colorbar import ColorbarBase
 
 DINA4_LANDSCAPE = (11.69, 8.27)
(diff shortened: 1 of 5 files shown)


niworkflows/viz/plots.py
```diff
@@ -239,76 +241,75 @@ def plot_carpet(
...
+        ax.spines["left"].set_linewidth(3)
+        ax.spines["left"].set_color(colors[i])
+        ax.spines["left"].set_capstyle("butt")
+        ax.spines["left"].set_position(("outward", 2))
+
+        # Make all subplots have same xticks
+        xticks = np.linspace(0, data.shape[-1], endpoint=True, num=7)
+        ax.set_xticks(xticks)
+        ax.set_yticks([])
+        ax.set_ylabel(label)
```
celprov, line 288: Can you color code the labels with the same color as the colorbar ? It will be more easy to read.
```suggestion
        ax.set_ylabel(label, color = colors[i])
```

celprov (commented):
The carpetplot looks nice

niworkflows/viz/plots.py
oesteban: Are you sure?? I think that will actually make them harder to read.

oesteban: Let's go with them in black and you can try to convince me whenever you have time to generate some colored examples :)

Issue #694, opened 2022-03-15
TypeError: unlink() got an unexpected keyword argument 'missing_ok'


eilidhmacnicol:
`os.unlink()` was given a keyword argument, `missing_ok`, in Python 3.8 but `NiWorkflows` 1.5.x still supports Python 3.7.

Commit 518b360679780fc916cc1a3a180a2322430faf0f introduced this keyword, causing an error to be thrown in  nipreps/nirodents#55, whose Docker image is still in 3.7. Exemplary error trace from one of the three errors:
"""
Traceback (most recent call last):
  File "/usr/local/miniconda/lib/python3.7/site-packages/nipype/pipeline/plugins/multiproc.py", line 344, in _send_procs_to_workers
    self.procs[jobid].run(updatehash=updatehash)
  File "/usr/local/miniconda/lib/python3.7/site-packages/nipype/pipeline/engine/nodes.py", line 516, in run
    result = self._run_interface(execute=True)
  File "/usr/local/miniconda/lib/python3.7/site-packages/nipype/pipeline/engine/nodes.py", line 635, in _run_interface
    return self._run_command(execute)
  File "/usr/local/miniconda/lib/python3.7/site-packages/nipype/pipeline/engine/nodes.py", line 741, in _run_command
    result = self._interface.run(cwd=outdir)
  File "/usr/local/miniconda/lib/python3.7/site-packages/nipype/interfaces/base/core.py", line 419, in run
    runtime = self._run_interface(runtime)
  File "/usr/local/miniconda/lib/python3.7/site-packages/niworkflows/interfaces/bids.py", line 672, in _run_interface
    out_file.unlink(missing_ok=True)
TypeError: unlink() got an unexpected keyword argument 'missing_ok'
"""

Are there plans to stop supporting 3.7 (meaning I should bump NiRodents to 3.8) or is there a workaround?

effigies:
Ugh. I didn't realize that I was depending on new behavior. I thought we were testing all supported versions. I'll submit a PR.

That said, it's probably a good idea to bump up to 3.9 or even 3.10 when you get the opportunity. 3.7 will be end-of-life in another year or so.

mgxd:
nope - 3.7 isn't EOL til mid 2023 (https://endoflife.date/python)

we should probably replace those with:

```python
# 3.7 backport
try:
    out_file.unlink()
except FileNotFoundError:
    pass
```

eilidhmacnicol:
[...]

PR #697, opened 2022-03-17
CI: Install package across Python versions and run pytest 

effigies:
We have a kind of silly setup where we make sure that we can build the package across Python versions and get the correct version out, but we don't run even basic tests on these installations, meaning we miss things like #694.

This builds on #695, but I don't want to block that going in while I play with GitHub actions.


effigies:
@mgxd Any opposition to a merge?

mgxd (approved):
Nope, LGTM

PR #699, opened 2022-03-21
added transformation files to dwi datatype

### LINKED ISSUE #698: [ENH] Default path patterns for DWI preprocessing outputs
Hi,
I think the "default_path_patterns" that's used by the DerivativesDataSink should allow transformation files for DWI too (as it does for "func" datatype).
It should be an easy enough fix, as it only requires adding the "from" and "to" entities to "dwi" path patterns...

So it shouldn't raise the following error:
`
ValueError: Could not build path with entities {'subject': '1020', 'session': '202111280841', 'direction': 'FWD', 'suffix': 'xfm', 'datatype': 'dwi', 'extension': '.txt', 'to': 'T1w', 'from': 'dwi'}.
`

GalBenZvi:
fixes #698


effigies (approved):
Looks reasonable to me.

PR #703, opened 2022-03-30
FIX: Use copy function that does not preserve mtime when creating fsaverage directories

### LINKED ISSUE #702: Update atime when copying FreeSurfer
`shutil.copytree` uses `shutil.copy2`, which uses `shutil.copystat`. When we copy FreeSurfer `fsaverage*` directories, we are creating outputs that look as if they have not been accessed (atime) or modified (mtime) since whatever those are in the bundled FreeSurfer subjects. These may be what they are in the FreeSurfer distribution bundle.

The upshot is that on filesystems that do age-based sweeping (e.g. `/scratch` directories on many HPC clusters), we are basically setting users up to have critical files removed from under them.

This should hit fMRIPrep 20.2.x, 21.0.x and 22.0.x.

effigies:
Closes #702.

### DIFF
--- a/niworkflows/interfaces/bids.py
+++ b/niworkflows/interfaces/bids.py
@@ -4,7 +4,7 @@
 from collections import defaultdict
 from json import dumps, loads
 from pathlib import Path
-from shutil import copytree, rmtree
+import shutil
 from pkg_resources import resource_filename as _pkgres
 import re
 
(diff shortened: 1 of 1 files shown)


effigies:
@mgxd Would appreciate a review to make sure I'm not crazy. I intend to do a release cascade so we can be prepared for the next round of fMRIPrep releases.

mgxd (commented):
It doesn't seem to matter which copy function is used (at least from local testing on my Macbook), the directory mtime stays the same. In fact, the only timestamp that changes is ctime, and that happens regardless of the copy function.

If sweeps take that into account, we're probably fine - if not, we could just hack the mtime with something like

```
 os.utime(dest, (time.time(), time.time()))
```

mgxd:
I think we're looking at this in two different ways.

Your tests are comparing the files within `testdir`, which absolutely show how using `shutil.copy` over `shutil.copy2` modifies the mtime.

I was looking at the top level directory's mtime after using `shutil.copy` / `shutil.copy2`. My test was copying an existing `fsaverage*` directory, and then checking the copied directories' mtime. I'm not sure if the sweeps also care about directories, otherwise this is fine to merge.

effigies:
The relevant bit is actually the atime more than the mtime. What we think is happening is that the sweeps are seeing files that haven't been accessed in years/months and suddenly they're gone. Being in a not more recently accessed or modified directory does not seem to protect them.

Hope this is more clear. Can chat tomorrow if we're still not quite connecting.

PR #704, opened 2022-04-04
FIX: Alternate query for template brain mask

mgxd:
Catches cases where the brain mask is under the `label` entity (i.e. UNCInfant).

This should address the problem in nipreps/nibabies#187


mgxd:
I don't think this is too controversial, so merging it in

effigies:
I agree, seems reasonable.

PR #705, opened 2022-04-05
REL: 1.5.3

effigies:
Doing a PR first to verify that it's now safe to remove the `markupsafe` pin.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,18 @@
+1.5.3 (April 05, 2022)
+======================
+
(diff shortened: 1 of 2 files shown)


effigies:
Okay. Apparently we pin an old `jinja2` in https://github.com/nipreps-containers/miniconda. So we're stuck with this pin until we figure out when it's safe to update that.

mgxd:
i think niworkflows minor releases are probably the best time to upgrade that container - we should jump to jinja2 > 3 when moving to 1.6.0

PR #707, opened 2022-04-15
FIX: Address some reliability issues of the functional masking workflow

oesteban:
Addresses nipreps/fmriprep#2731

### DIFF
--- a/niworkflows/func/util.py
+++ b/niworkflows/func/util.py
@@ -364,7 +364,7 @@ def init_enhance_and_skullstrip_bold_wf(
 
     """
     from niworkflows.interfaces.nibabel import BinaryDilation
-    
+
     workflow = Workflow(name=name)
     inputnode = pe.Node(
         niu.IdentityInterface(fields=["in_file", "pre_mask"]), name="inputnode"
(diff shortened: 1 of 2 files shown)


niworkflows/func/util.py
```diff
@@ -403,16 +391,8 @@ def init_enhance_and_skullstrip_bold_wf(
...
-            operation="max",
-            kernel_shape="sphere",
-            kernel_size=6.0,
-            internal_datatype="char",
-        ),
-        name="skullstrip_first_dilate",
-    )
-    bet_mask = pe.Node(fsl.ApplyMask(), name="skullstrip_first_mask")
+    bet_dilate = pe.Node(BinaryDilation(radius=6), name="skullstrip_first_dilate")
+    bet_mask = pe.Node(ApplyMask(), name="skullstrip_first_mask")
```
mgxd, line 395: we should probably take the chance to rename these, since we're changing the interfaces altogether

mgxd (approved):
I think this looks good (at least from locally testing ds210) - would you mind rebasing to `maint/1.4.x` since we'll need to backport this fix to fMRIPrep 21.0.x

oesteban:
Okay, rebased. I'm going to generate a new branch `masks/` so that the mask tests are triggered.

BTW - any ideas of why get data is failing on circle?

effigies:
Just from a quick look, it seems this violates our bug-fix rules. Does it make sense to target 1.5.x instead and just go in 22.0.0?

effigies:
> BTW - any ideas of why get data is failing on circle?

https://github.com/datalad/datalad/pull/6631#issuecomment-1102682597

PR #712, opened 2022-04-20
FIX: Improve reliability of BOLD masking workflow

mgxd:
This is a backport of #711 to get in the 21.0.x series of fMRIPrep


mgxd:
I think this is good to go, tested in https://github.com/nipreps/fmriprep/pull/2731

PR #713, opened 2022-04-20
CI: Clear conda environment cache to fix get_data

effigies:


mgxd:
can we just restrict datalad 0.16.0/1?

effigies:
Yeah, I guess that's best. Just don't understand why we can't set up a conda environment without this happening.

effigies:
Guess so. Not clear why...

PR #715, opened 2022-04-21
REL: 1.4.9

mgxd:


effigies:
Nope, not fixed. See https://github.com/nipreps/niworkflows/tree/maint/1.4.x.

mgxd:
datalad just cut a new release, that's probably why.

effigies:
I don't think so. I updated the 0.16.1 package on conda; it's just not getting updated. Your pin might work in re-triggering things, though. I'd test it directly on `maint/1.4.x`. You're currently working off of the `master` cache instead of the `maint/1.4.x` cache, which is in a nasty state that I was trying (and failing) to correct.

effigies:
I agree that's a good idea. However, we need to bump the cache versions to make this take effect (since your `rel/1.4.9` cache has already been polluted by `master`). And then we're back to where we started. Shall I try again with a new branch and cache number?

PR #716, opened 2022-04-21
CI: Fallback to maintenance branch first

effigies:
New branch, so we don't need to bump the cache numbers yet.


effigies:
Actually: 

> No cache is found for key: env-v1-maint/1.4.x-
> Found a cache from build 4620 at env-v1-

Need to bump.

effigies:
Well, it made it through. Maybe this will be enough to push `maint/1.4.x` into a stable state. Will try again if not.

PR #726, opened 2022-05-25
REL: 1.6.0

effigies:
Prepare 1.6.0 release.

@mgxd Is there any reason we should include some of these fixes in 1.5.x? We could do a 1.5.5 release with the old miniconda image and a 1.6.0 with the new.

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,14 +1,8 @@
-networkx != 2.8.1
-furo ~= 2021.10.09
-matplotlib >= 2.2.0
-nibabel
+attrs
+furo ~= 2022.04.07
 nipype >= 1.5.1
-numpy
 packaging
-pydot >= 1.2.3
-pydotplus
 sphinx ~= 4.2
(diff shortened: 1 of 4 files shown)


mgxd:
I don't think there is any urgency to release 1.5.5 - I would punt that release until it's needed

effigies:
Mocking `etelemetry` and `templateflow.api.get()` turned a 26 minute doc build into 47s. Probably won't help with the full builds, but maybe we can set `NO_ET` at least to save some time.

docs/requirements.txt
```diff
@@ -1,14 +1,8 @@
-networkx != 2.8.1
-furo ~= 2021.10.09
-matplotlib >= 2.2.0
-nibabel
+attrs
+furo ~= 2022.04.07
```
mgxd, line 2:
```suggestion
furo ~= 2022.4.7
```

mgxd (approved):

PR #733, opened 2022-07-05
MAINT: Test if prereleases are compatible now

mgxd:
Since these are not first-class dependencies, we shouldn't be managing them.


mgxd:
Yeah, would be cleaner - I just wanted to appease the tests so I could get a release out.

mgxd:
oops forgot about this - opened #735

PR #741, opened 2022-09-05
FIX: Missing outputs in ``RobustAverage``

oesteban:
Trivial improvements to the existing interface:

- Re-enable the output of HMC matrices for AFNI's ``3dVolReg``
- Map the realigned file of references to an output of the interface for further use.

### DIFF
--- a/niworkflows/interfaces/images.py
+++ b/niworkflows/interfaces/images.py
@@ -225,6 +225,7 @@ class _RobustAverageInputSpec(BaseInterfaceInputSpec):
     nonnegative = traits.Bool(
         True, usedefault=True, desc="whether the output should be clipped below zero"
     )
+    num_threads = traits.Int(desc="number of threads")
 
 
 class _RobustAverageOutputSpec(TraitedSpec):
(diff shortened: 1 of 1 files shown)


oesteban:
> I would be careful of adding exists=True to output specs. Removing unnecessary outputs will cause these to get deleted, making resumes impossible.

I don't think `exists=True` has any effects for nipype, the output spec's trait's metadata are not ever checked.

The removal of unnecessary outputs is fully handled by the engine implementation, based on whether there is an outbound connection from the interface or not.

Happy to remove it though, if that is of any clarity. However, if a future version of nipype/pydra would look at these annotations, then the semantics would be adequate (as in, if the realignment worked well, a realigned output file must exist).

Sounds reasonable?

oesteban:
Just to clear out my way on the SDCFlows front, I'm going to merge this. If you feel we need to revise the `exists=True` issue, please open an issue and I'm happy to remove it down the line.

effigies:
@oesteban: https://github.com/nipy/nipype/issues/3493

PR #743, opened 2022-09-06
CI: Extract pip pre checks into separate, skippable job

mgxd:
Since github actions does not yet support early, successful job termination(!), this required splitting the test/install into separate stable and pre-release jobs.


effigies (approved):

mgxd:
seems like merge commits break it https://github.com/nipreps/niworkflows/runs/8219736026?check_suite_focus=true

effigies:
Fixed in #744

PR #749, opened 2022-09-26
FIX: Use nilearn plot_surf over surfplot

effigies:
Backport gh-745 to maint/1.4.x to fix builds.


effigies:
There will not be another release in 1.4.x.

PR #753, opened 2022-10-17
ENH: Add morphometric suffixes to pattern matching

mgxd:
Required for nipreps/smriprep#305


niworkflows/interfaces/surf.py
```diff
@@ -154,13 +154,19 @@ class Path2BIDS(SimpleInterface):
...
+    suffix = sulc
+    <BLANKLINE>
     """
 
     input_spec = _Path2BIDSInputSpec
     output_spec = _Path2BIDSOutputSpec
     _pattern = re.compile(
         r"(?P<hemi>[lr])h.(?P<suffix>(wm|smoothwm|pial|midthickness|"
-        r"inflated|vinflated|sphere|flat))[\w\d_-]*(?P<extprefix>\.\w+)?"
+        r"inflated|vinflated|sphere|flat|sulc|curv|thickness))[\w\d_-]*(?P<extprefix>\.\w+)?"
```
effigies, line 169: Do we want to add any other values while we're here?

mgxd: We could add any of the others in https://bids-specification.readthedocs.io/en/bep011/05-derivatives/04-structural-derivatives.html#surface-mapped-anatomical-scalar-derivatives

However, I'm fine with doing it as needed

PR #754, opened 2022-10-18
ENH: Add `Resolution` field to metadata if data has `res` entity

### LINKED ISSUE #752: Missing metadata for outputs that use the `res` entitiy.
Derivatives specification says if the `res` entitiy is used `Resolution` must be defined in its json file:
https://bids-specification.readthedocs.io/en/stable/05-derivatives/03-imaging.html#preprocessed-coregistered-andor-resampled-volumes

The actual content of the `Resolution` field is pretty wide open so boiler text explaining the number directly corresponds to millimeter resolution should be good enough.

This is an issue for both anatomical and functional outputs.

mgxd:
Closes #752


effigies:
We should do the same with `den`, which CIFTI and GIFTI files may have.

mgxd:
@effigies Do you have any suggestions on how to grab that information? Load up the `gii`/`dtseries.nii` and sum the vertices / grayordinates?

effigies:
I assumed there was metadata, as with res.

mgxd:
Nope, opened https://github.com/templateflow/templateflow/issues/145 though

PR #760, opened 2022-11-28
FIX: Use np.bool_ instead of deprecated np.bool

effigies:


effigies:
Fixes the pre-release build. Guess we'll deal with the Docker build separately.

PR #769, opened 2023-01-11
FIX: Coerce `MGHImage`s to `Nifti1Image`s when plotting

mgxd:
Supercedes https://github.com/nipreps/smriprep/pull/317

### DIFF
--- a/niworkflows/viz/utils.py
+++ b/niworkflows/viz/utils.py
@@ -348,6 +348,9 @@ def plot_registration(
         plot_params = robust_set_limits(anat_nii.get_fdata().reshape(-1), plot_params)
 
     # FreeSurfer ribbon.mgz
+    if contour:
+        contour = _ensure_nii(contour)
+
     ribbon = contour is not None and np.array_equal(
         np.unique(contour.get_fdata()), [0, 2, 3, 41, 42]
     )
--- a/niworkflows/viz/utils.py
+++ b/niworkflows/viz/utils.py
@@ -341,6 +341,8 @@ def plot_registration(
     if cuts is None:
         raise NotImplementedError  # TODO
 
+    anat_nii = _ensure_nii(anat_nii)
+
     out_files = []
     if estimate_brightness:
         plot_params = robust_set_limits(anat_nii.get_fdata().reshape(-1), plot_params)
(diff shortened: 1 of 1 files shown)


mgxd:
This fixed the errors previously encountered in the nibabies runs

effigies:
We should report this upstream to nilearn, as well. Workaround is fine because all we care about is data/affine, but they at least used to be able to handle `SpatialImage`s in general, so this seems like a regression.

mgxd:
@effigies - Looks like it needs to be a NiftiImage, so might as well coerce.

https://nilearn.github.io/stable/modules/generated/nilearn.plotting.plot_anat.html#nilearn.plotting.plot_anat
https://nilearn.github.io/stable/manipulating_images/input_output.html#inputing-data-file-names-or-image-objects

effigies:
https://nilearn.github.io/stable/manipulating_images/input_output.html#niimg-like-objects

> A string or pathlib.Path object with a file path to a Nifti or Analyse image
> An SpatialImage from nibabel, ie an object exposing get_fdata() method and affine attribute, typically a Nifti1Image from [nibabel](http://nipy.sourceforge.net/nibabel/).

niworkflows/viz/utils.py
```diff
@@ -341,11 +341,16 @@ def plot_registration(
...
 
+    anat_nii = _ensure_nii(anat_nii)
+
     out_files = []
     if estimate_brightness:
         plot_params = robust_set_limits(anat_nii.get_fdata().reshape(-1), plot_params)
 
     # FreeSurfer ribbon.mgz
+    if contour:
+        contour = _ensure_nii(contour)
```
effigies, line 352: This is just `nb.Nifti1Image.from_image()`. Let's also add a comment so we know why we did this.

```suggestion
    # nilearn 0.10.0 uses Nifti-specific methods
    anat_nii = nb.Nifti1Image.from_image(anat_nii)

    out_files = []
    if estimate_brightness:
        plot_params = robust_set_limits(anat_nii.get_fdata().reshape(-1), plot_params)

    # FreeSurfer ribbon.mgz
    if contour:
        contour = nb.Nifti1Image.from_image(anat_nii)
```

effigies:
I'm +1 for merging rather than relying on people not to have this specific version of nilearn.

niworkflows/viz/utils.py
```diff
@@ -714,3 +720,9 @@ def plot_melodic_components(
         pad_inches=0.01,
     )
     fig.clf()
+
+
+def _ensure_nii(img: nb.spatialimages.SpatialImage) -> nb.Nifti1Image:
+    if not isinstance(img, nb.Nifti1Image):
+        img = nb.Nifti1Image(img.dataobj, img.affine, img.header)
+    return img
```
mgxd, line 728:
```suggestion
```

effigies (approved):

PR #770, opened 2023-01-11
MNT: Use org-level context

mgxd:


effigies:
We need to fix the docker builds... We'll spend forever in conda trying to find a consistent environment.

effigies:
This needs doing, regardless of whether builds succeed.

PR #771, opened 2023-01-11
FIX: Remove unused ANTs parameter that was removed in 2.4.1

effigies:
`--use-estimate-learning-rate-once` was removed in https://github.com/ANTsX/ANTs/pull/1411 (released v2.4.1). It seems it has had no effect for some time (I think since v2.0.0).


effigies:
@mgxd @oesteban This is ready for review.

effigies:
Incidentally, I've been adding things ready for review to the "Needs review" column in https://github.com/orgs/nipreps/projects/7/views/8. If a label would be more prominent, I'm happy to do that instead/in addition.

oesteban (approved):

oesteban:
Thanks for this. Will look at your list and see whether labels seem more prominent or the list should be sufficient for me. Probably the latter.

PR #772, opened 2023-01-11
FIX: Remove unused ANTs parameter that was removed in 2.4.1

effigies:
Backporting #771 and #770 to the LTS branch.


mgxd:
@effigies the `resample_img()` issue was also present in nilearn 0.9.2. However, it was rarely hit as `plot_anat()` needs to have specific `cut_coords` to trigger resampling.

PR #773, opened 2023-01-11
MNT: Try bumping docker base to latest nipreps/miniconda

effigies:
Docker builds are broken due to conda environment solving timing out (>2 hours).


effigies:
Not good enough.

mgxd (approved):
LGTM, though perhaps we should be using a similar recipe to fmriprep...

effigies:
Yeah, I wouldn't mind that. Was going for a minimal diff, since we end up refactoring this thing to/from fmriprep-like about once a year.

PR #774, opened 2023-01-18
MAINT: Update ``FS_LICENSE`` (from context) and docker orb version

oesteban:
- Avoid having the license encoded string directly in the config file (e.g., if we are requested to stop using Chris G's)
- Update docker orb version (although I don't think we are using the orb at any moment -- not just niworkflows, btw).


effigies:
This will break every PR from a fork.

PR #777, opened 2023-01-24
Pass custom configuration into DerivativesDataSink entity parser

### LINKED ISSUE #776: DerivativesDataSink does not extract entities from custom configuration
As of 1.7.1, `DerivativesDataSink` only extracts entities present in the default pybids configurations from the `source_file`. This means that any custom entities defined in the `_config_entities` attribute of the `DerivativesDataSink` are ignored.

I think the relevant part of the code is:

https://github.com/nipreps/niworkflows/blob/1555d2b117d232654e6bd011813fb1767bd88ab6/niworkflows/interfaces/bids.py#L551-L555

Currently, `parse_file_entities` accepts `entities` and `config` options, so I think `DerivativesDataSink` could feed the custom configuration info into the `parse_file_entities` call.

I don't know if it's worth it to fix this with the current version of pybids, given the work being done in https://github.com/bids-standard/pybids/pull/926.

tsalo:
Closes #776.

Changes proposed:
- Convert `DerivativesDataSink._config_entities` and `DerivativesDataSink._file_patterns` into a `bids.layout.Config` object, then feed this Config into `parse_file_entities` (along with the default `bids` and `derivatives` ones) in order to extract custom entities from the `source_file`.

### DIFF
--- a/niworkflows/data/nipreps.json
+++ b/niworkflows/data/nipreps.json
@@ -34,7 +34,7 @@
     },
     {
       "name": "run",
-      "pattern": "[_/\\\\]+run-*(\\d+)",
+      "pattern": "[_/\\\\]+run-(\\d+)",
       "dtype": "int"
     },
     {
--- a/niworkflows/data/nipreps.json
+++ b/niworkflows/data/nipreps.json
@@ -34,7 +34,7 @@
     },
     {
       "name": "run",
-      "pattern": "[_/\\\\]+run-0*(\\d+)",
+      "pattern": "[_/\\\\]+run-*(\\d+)",
       "dtype": "int"
     },
     {
--- a/niworkflows/interfaces/bids.py
+++ b/niworkflows/interfaces/bids.py
@@ -55,7 +55,7 @@ from ..utils.misc import _copy_any, unlink
 
 regz = re.compile(r"\.gz$")
 _pybids_spec = loads(Path(_pkgres("niworkflows", "data/nipreps.json")).read_text())
-BIDS_DERIV_ENTITIES = frozenset({e["name"] for e in _pybids_spec["entities"]})
+BIDS_DERIV_ENTITIES = _pybids_spec["entities"]
 BIDS_DERIV_PATTERNS = tuple(_pybids_spec["default_path_patterns"])
 
 STANDARD_SPACES = tf.api.templates()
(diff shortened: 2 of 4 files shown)


effigies:
@tsalo Do you need help working through test failures?

tsalo:
I didn't even realize crucial tests were failing, but it's pretty clear what went wrong. I will definitely have questions about the cleanest version of the code, but I can figure out the failing tests at least.

tsalo:
The first set of failures came from feeding the pybids Config the entity names, rather than the entity dictionary. Once I fixed that, there was a second round of failures that I _think_ stemmed from a mismatch between how the run entity is defined in niworkflows and in pybids. I think the issue is that, because DerivativesDataSink never used the entity definitions from the custom config file (i.e., the niworkflows one), all of the tests are geared toward the entity definitions from pybids. I ended up changing the entity definition in the niworkflows config file, but I don't know if the change (a leading zero) was important. If it is, then I think the tests need to be updated instead.

I'm not sure what's causing the current round of failures though.

effigies:
Try this patch:

```Diff
diff --git a/niworkflows/reports/tests/test_core.py b/niworkflows/reports/tests/test_core.py
index 7d45b4fb..bf474e2c 100644
--- a/niworkflows/reports/tests/test_core.py
+++ b/niworkflows/reports/tests/test_core.py
@@ -32,6 +32,7 @@ from yaml import safe_load as load
 import matplotlib.pyplot as plt
 from bids.layout.writing import build_path
 from bids.layout import BIDSLayout
+from bids.layout.utils import PaddedInt
 
 import pytest
 
@@ -152,11 +153,11 @@ def test_report2(bids_sessions):
             [
                 (None, "faketask"),
                 (None, "faketask2"),
-                (1, "faketaskwithruns"),
-                (1, "mixedgamblestask"),
-                (2, "faketaskwithruns"),
-                (2, "mixedgamblestask"),
-                (3, "mixedgamblestask"),
+                (PaddedInt("01"), "faketaskwithruns"),
+                (PaddedInt("01"), "mixedgamblestask"),
+                (PaddedInt("02"), "faketaskwithruns"),
+                (PaddedInt("02"), "mixedgamblestask"),
+                (PaddedInt("03"), "mixedgamblestask"),
             ],
         ),
         ([""], [], []),
```

(If this isn't something you regularly do, you can do it just by running `patch` pasting the above, and finishing with `ctrl-d`.)

effigies:
Just to explain, to preserve zero-padding when generating filenames, we had to add a special class to pybids that would remember its zero-padding when `str(run)` was called. The function in the test is failing because it sorts lexically and `"02" < "1"`.

tsalo:
Thanks!

Zero-padding is so annoying. I say we get rid of it in BIDS 2.0.

effigies:
Lexical sorting is pretty popular. I think that's an uphill battle.

effigies:
@tsalo Could you have a look at the latest commit and see if it outrages you beyond reason? I think it should fix these tests once and for all.

tsalo:
It seems like a wonderful solution to me!

PR #779, opened 2023-01-27
FIX: Restore confound order in correlation plot

effigies:
Finally got irritated enough to figure out why the confound order went nuts in the correlation plot. Using set, there are no order guarantees. We can get deduplication with `dict.from_keys()` instead, which produces a set-like dict with values all `None`. We can then just `list()` on the dict to get the keys back in the order they were initially passed.


effigies:
## Before

Screenshot from 2023-01-27 18-27-23
## After

Screenshot from 2023-01-27 18-59-10

PR #784, opened 2023-03-01
FIX: Use a more reliable test of NIfTIness

### LINKED ISSUE #778: DerivativesDataSink datatype coercion does not skip non-dtseries CIFTIs
I have been trying to write out a `dseg`-suffixed CIFTI file with the `.dlabel.nii` extension with `DerivativesDataSink`, but it raises the following exception:

```
Traceback:
	  File "/usr/local/miniconda/lib/python3.8/site-packages/nipype/interfaces/base/core.py", line 398, in run
	    runtime = self._run_interface(runtime)
	  File "/usr/local/miniconda/lib/python3.8/site-packages/niworkflows/interfaces/bids.py", line 712, in _run_interface
	    new_header.set_data_dtype(data_dtype)
	AttributeError: 'Cifti2Header' object has no attribute 'set_data_dtype'
```

In the following section, DerivativesDataSink checks to see if an output file is a NIFTI by looking at the extension. However, it only counts `.dtseries.nii[.gz]` files as CIFTIs, and assumes any other files ending in `.nii[.gz]` are NIFTIs.

https://github.com/nipreps/niworkflows/blob/1555d2b117d232654e6bd011813fb1767bd88ab6/niworkflows/interfaces/bids.py#L651-L653

After that, DerivativesDataSink attempts to coerce the in_file's datatype to a target type for that suffix. However, this step only works for NIFTIs, so any CIFTIs mislabeled as NIFTIs above will cause an error.

I can't think of a great general-purpose solution, but two ideas I had were:

1. Hardcode the full list of CIFTI extensions in the check. The extensions might need to be updated from time to time though.
2. In the following section, check for a `nii.nifti_header` attribute, and set the datatype for that attribute instead of `nii.header` when it is present.

    https://github.com/nipreps/niworkflows/blob/1555d2b117d232654e6bd011813fb1767bd88ab6/niworkflows/interfaces/bids.py#L697-L712

effigies:
Closes #778.


tsalo (commented):
The change makes sense to me. Thanks for addressing this!

Is it worth including `Nifti2Image`s too? I know they're rare, and I don't know if the header check is meant to work with `Nifti2Image`s.

effigies:
`Nifti2Image`s will be caught by this check:

```Python
In [1]: import nibabel as nb

In [2]: nb.Nifti2Image.mro()
Out[2]: 
[nibabel.nifti2.Nifti2Image,
 nibabel.nifti1.Nifti1Image,
 nibabel.nifti1.Nifti1Pair,
 nibabel.analyze.AnalyzeImage,
 nibabel.spatialimages.SpatialImage,
 nibabel.dataobj_images.DataobjImage,
 nibabel.filebasedimages.SerializableImage,
 nibabel.filebasedimages.FileBasedImage,
 object]
```

tsalo:
Oh, awesome!

Issue #787, opened 2023-03-10
Finalize the migration of visual elements to NiReports


oesteban:
### What happened?

Replace current functionalities with shims that spit out deprecation warnings.

### What command did you use?

```shell
n/a
```

### What version of fMRIPrep are you running?

n/a

### How are you running this software?

Docker

### Is your data BIDS valid?

Yes

### Are you reusing any previously computed results?

No

PR #789, opened 2023-03-12
MAINT: Replace ported sources with shim imports and deprecation warnings

oesteban:
Resolves: #787.


effigies:
I would specifically not like to depend on nireports. I would just deprecate and not shim.

PR #792, opened 2023-03-28
MNT: Update build system and CI invocations

effigies:
Switch to `hatchling`+`hatch-vcs`. Move configuration to `pyproject.toml` where supported, away from `setup.cfg` in any case. Also generally use `pipx` instead of installing one-off tools in the base environment.


effigies:
@mgxd Would you mind having a look through?

.github/workflows/pythonpackage.yml
```diff
@@ -230,22 +239,17 @@ jobs:
...
-        echo "VERSION: \"${THISVERSION}\""
         echo "INSTALLED: \"${INSTALLED_VERSION}\""
-        test "${INSTALLED_VERSION}" = "${THISVERSION}"
+        test "${INSTALLED_VERSION}" = "${VERSION}"
+      env:
+        VERSION: ${{ needs.job_metadata.outputs.version }}
     - name: Install test dependencies
       run: python -m pip install $PIP_FLAGS "niworkflows[tests]"
     - name: Run tests
       uses: GabrielBB/xvfb-action@v1
```
mgxd, line 250: Can this be removed now that we dropped surfplot?

niworkflows/reports/tests/test_core.py
```diff
@@ -37,6 +35,11 @@
 
 from ..core import Report
 
+try:
+    import importlib_resources
+except ImportError:
+    import importlib.resources as importlib_resources
```
mgxd, line 41: and here

```suggestion
try:
    import importlib.resources as importlib_resources
except ImportError:
    import importlib_resources
```

mgxd (approved):
Looks good! Thanks for modernizing 😄

niworkflows/conftest.py
```diff
@@ -29,12 +29,23 @@
 import pandas as pd
 import pytest
 import tempfile
-import pkg_resources
+
+try:
+    import importlib_resources
+except ImportError:
+    import importlib.resources as importlib_resources
```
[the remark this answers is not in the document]
effigies: `importlib.resources` is present I think in all supported versions, so this would prevent us from using the backported features.

An alternative is to import the specific required functions:

```Python
try:
    from importlib.resources import files, as_file
except ImportError:
    from importlib_resources import files, as_file
```

But they're a little cryptic without the module name. We could do `import files as ...` but then we'd need to come up with a name.

.github/workflows/pythonpackage.yml
effigies: Let's try.

PR #795, opened 2023-04-10
FIX: Use len(segments) to avoid nsegments becoming wrong

effigies:
Another in the series of handling carpetplots in degenerate cases that pretty much only come up on CI. Here, we have a variable `nsegments = len(segments)`, but segments can be pared down if there are empty segments.

<100ns to call `len()` on a dict 4 times seems like a good trade over possibly getting de-synced again in the future.


effigies:
I assume no objections and will merge in the morning, but feel free to make any opinions known before then.

PR #798, opened 2023-04-13
MNT: Create a requirements.txt to keep the LTS branch buildable

effigies:
This PR adds a `requirements.txt` for use in building the docker image for testing the LTS branch of niworkflows.

To do this, I needed to create a `requirements.in` that incorporates `setup.cfg` with some additional constraints, as well as some pins from the [environment.yml](https://github.com/nipreps/fmriprep/blob/maint/20.2.x/docker/files/environment.yml). The goal is to maintain a requirements set that will continue to build on Python 3.7 without putting new `==` or `<` into `setup.cfg`.

One wrinkle in this was that setuptools 58 made it impossible to build some of the versions of packages we pin by removing Python 2/3 compatibility shims that tools were still using 3 years ago.

The instigating problem here was that newer seaborn requires newer matplotlib than we currently pin, but it's likely we'll have to address a few more of these before the LTS dies, and I wanted something that should hopefully make that next time less painful.


effigies:
@mgxd Any objections?

niworkflows/interfaces/utils.py
```diff
@@ -956,7 +956,7 @@ def less_breakable(a_string):
     drop_columns = drop_columns or []
     additional_metadata = additional_metadata or {}
     try:
-        tsv_data = pd.read_csv(in_tsv, "\t")
+        tsv_data = pd.read_csv(in_tsv, delimiter="\t")
```
mgxd, line 959: It looks like delimiter is just an alias anyways
```suggestion
        tsv_data = pd.read_csv(in_tsv, sep="\t")
```

mgxd (approved):
lgtm

PR #802, opened 2023-04-27
MAINT: Mark future versions as post releases on the current major/minor/micro

mgxd:
Right now we have released `1.7.9`, so future changes are being marked as `1.8.0.dev...`, breaking some of our pins.

This change avoids having to guess the next version and just marks as a post-release: `niworkflows-1.7.9.post53+g24f54712e`.

Alternatively, if there is a way to ensure `X.Y.9` -> `X.Y.10.dev...`, I'm fine with going that route.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -104,7 +104,7 @@ exclude = [
 
 [tool.hatch.version]
 source = "vcs"
-raw-options = {version_scheme = "release-branch-semver" }
+raw-options = {version_scheme = "post-release"}
 
 [tool.hatch.build.hooks.vcs]
 version-file = "niworkflows/_version.py"


effigies:
The 9 doesn't matter. The question is whether we're on a development or maintenance branch. Builds from `maint/1.7.x` will be `1.7.10.dev...` and builds from `master` will be `1.8.0.dev...`. If you have a bug-fix branch that you are specifically trying to target 1.7.x for, I would do something like `fix/1.7.x-description`

mgxd:
I see - this originally came up while playing around with the latest developmental versions of the preps on nibabies. Mainly, the problem is [smriprep's hard restriction of niworkflows](https://github.com/nipreps/smriprep/blob/master/setup.cfg#L28).

So perhaps the best course here is to just remove the compatiblity pin on smriprep's side.

effigies:
Yes. I would say the `maint/0.11.x` branch can pin `niworkflows ~= 1.7.x`, but `master` should not.

PR #803, opened 2023-05-05
FIX: Inspect if template has resolution instead of assuming

mgxd:
Addressing https://github.com/templateflow/templateflow/issues/149

### DIFF
--- a/niworkflows/utils/misc.py
+++ b/niworkflows/utils/misc.py
@@ -22,6 +22,8 @@
 #
 """Miscellaneous utilities."""
 import os
+from typing import Optional
+import warnings
 
 
 __all__ = [
@@ -35,7 +37,12 @@ __all__ = [
 ]
 
 
-def get_template_specs(in_template, template_spec=None, default_resolution=1):
+def get_template_specs(
+    in_template: str,
+    template_spec: Optional[dict] = None,
+    default_resolution: int = 1,
+    fallback: bool = False
+):
     """
     Parse template specifications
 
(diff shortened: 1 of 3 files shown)


niworkflows/utils/misc.py
```diff
@@ -72,11 +88,26 @@ def get_template_specs(in_template, template_spec=None, default_resolution=1):
         "res", template_spec.get("resolution", default_resolution)
     )
 
+    # Verify resolution is valid
+    if fallback:
+        res = template_spec['resolution']
+        if res and not isinstance(res, list):
```
effigies, line 94: Are you checking if `res` is falsey or specifically `None`? It could be an empty list, `0` or `''` and fail this.

```suggestion
        if res is None and not isinstance(res, list):
```

oesteban: I think your suggestion would actually break the logic.

That said, if you always listify, the intersection below (98) will still be empty with weird values. So it would be easier to just listify:

```suggestion
        if not isinstance(res, list):
```

niworkflows/utils/misc.py
```diff
@@ -72,11 +88,26 @@ def get_template_specs(in_template, template_spec=None, default_resolution=1):
...
+            res = [int(res)]
+
+        available_resolutions = tf.TF_LAYOUT.get_resolutions(template=in_template)
+        if res and not set(res) & set(available_resolutions):
```
oesteban, line 98:
```suggestion
        if not (set(res) & set(available_resolutions)):
```

parens for clarity?

effigies (approved):
This seems reasonable.

PR #806, opened 2023-05-18
FIX: Move from smoothwm to white

effigies:
We have some places where `wm` is used for FreeSurfer surfaces instead of `white`, breaking regexes/path building when using `white` surfaces.

This PR also adjusts the (unused by anybody) gifti_surface_wf to convert white surfaces instead of smoothwm (see https://github.com/nipreps/smriprep/issues/338), but that could be removed if we don't want to call that a bug fix.


mgxd (approved):
I also wondered why we were using smoothwm rather than white, but this looks good to me. I do think it is a big enough change to warrant a minor bump though.

effigies:
Sounds good. I don't think we'll need this for any maintenance releases of downstream tools.

I think the reason was that I got white and smoothwm mixed up as to which was most refined ages ago.

PR #816, opened 2023-07-10
ENH: Add data loader class, install at root and data modules

effigies:
Fixes current test failures from using a deprecated `importlib.resources` API.

Expanded on an `ExitStack`-based load function that @mgxd used in one of these projects into a class that provides flexible access to package data with readable, unzipped, and cached access methods.

We have the following loaders:

* `niworkflows.load_resource` - Generic, can get anything from anywhere
* `niworkflows.data.load` - Primary access point for package data
* `niworkflows.tests.data.load_test_data` - Convenience for tests
* `niworkflows.interfaces.tests.data.load_test_data` - Convenience for interface tests

I would like to figure out how to bundle the library in a zip to actually test that this works when installed that way, but for this PR the goal is to make it easy to think about and select whether we need the resource to be a `Path` or just `Traversable`.

This also finally purges `pkg_resources`, which we should avoid because we have no runtime dependency on setuptools, and the API is discouraged in third-party packages.

### DIFF
--- a/niworkflows/data/__init__.py
+++ b/niworkflows/data/__init__.py
@@ -10,6 +10,8 @@
 
 .. autoclass:: Loader
 """
+from __future__ import annotations
+
 import atexit
 import os
 from contextlib import AbstractContextManager, ExitStack
--- a/niworkflows/data/__init__.py
+++ b/niworkflows/data/__init__.py
@@ -1,0 +1,180 @@
+"""Niworkflows data files
+
+.. autofunction:: load
+
+.. automethod:: load.readable
+
+.. automethod:: load.as_path
+
+.. automethod:: load.cached
+
+.. autoclass:: Loader
+"""
+import atexit
+import os
+from contextlib import AbstractContextManager, ExitStack
+from functools import cached_property
+from pathlib import Path
+from types import ModuleType
+from typing import Union
+
+try:
+    from functools import cache
+except ImportError:  # PY38
+    from functools import lru_cache as cache
+
+try:  # Prefer backport to leave consistency to dependency spec
+    from importlib_resources import as_file, files
+except ImportError:
+    from importlib.resources import as_file, files  # type: ignore
+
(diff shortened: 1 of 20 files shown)


effigies:
Tests passing. Anybody have time for a review?

niworkflows/data/__init__.py
```diff
@@ -0,0 +1,182 @@
...
+            f"Load package files relative to ``{self._anchor}``.",
+            "",
+            "This package contains the following (top-level) files/directories:",
+            "",
+            *(f"* ``{path}``" for path in top_level),
+        ]
+
+        return "\n".join(doclines)
+
+    def readable(self, *segments) -> Traversable:
```
mgxd, line 144: Since it seems like this step is often, if not always, followed with a `read_text()` call, wdyt about adding another method (`read_text`?) to serve as some syntactic sugar?

mgxd (commented):
This is cool! If I'm understanding it right, this can be used as a dataloader across the *preps ecosystem?

niworkflows/data/__init__.py
effigies: I like the explicitness of `loader.how_to_load(what_to_load).do_something()`.

effigies:
> If I'm understanding it right, this can be used as a dataloader across the *preps ecosystem?

Yes. I would probably just vendor it in the local `data/__init__.py` for a few versions, rather than depend on it in niworkflows.

I think it's a handy util but not sure it's stable enough to really want to deal with it as a dependency. I did an [initial pass in BIDS](https://github.com/bids-standard/bids-specification/blob/master/tools/schemacode/bidsschematools/data/__init__.py), and then while working on this refactor found I wanted to explicitly avoid caching.

oesteban:
> Yes. I would probably just vendor it in the local `data/__init__.py` for a few versions, rather than depend on it in niworkflows.

Just to confirm - "vendor it in" means copying the infrastructure across packages (for a few versions)?

effigies:
Yes

PR #819, opened 2023-08-02
ENH: add gradunwarp base workflow for f/smriprep. 

bpinsard:
Working on https://github.com/nipreps/fmriprep/issues/1550. 
Relies on an updated version of the gradunwarp https://github.com/bpinsard/gradunwarp/tree/fix/mem_leaks.


niworkflows/workflows/gradunwarp.py
```diff
@@ -64,6 +64,29 @@ def init_gradunwarp_wf(
...
+        import os
+        import nitransforms.io
+        from nipype.utils.filemanip import fname_presuffix
+        fsl_warp = nitransforms.io.fsl.FSLDisplacementsField.from_filename(in_warp)
+        itk_warp_data = fsl_warp.get_fdata().reshape(fsl_warp.shape[:3]+(1,3))
+        itk_warp_data[...,(0,1)] *= -1
+        itk_warp = fsl_warp.__class__(itk_warp_data, fsl_warp.affine)
+        itk_warp.header.set_intent("vector")
+        out_fname = fname_presuffix(in_warp, suffix="_itk", newpath=os.getcwd())
+        itk_warp.to_filename(out_fname)
```
oesteban, line 77: nitransforms should do all the manipulations for you

bpinsard: I might have missed something, but from what I understood what needs to be provided to Ants ApplyTransform are ITK warpfields. In nitransforms I cannot see any function to output non-linear warp in ITK format from nitransforms object https://nitransforms.readthedocs.io/en/latest/_api/io.html#nitransforms.io.itk.ITKDisplacementsField, it seems to only deal with inputs. Thus, my attempt to invert that input step.

oesteban: Yeah, it looks like you are totally right (and I did the same here a while ago - https://github.com/nipreps/sdcflows/blob/375dd7a024abcd5d4d2b282241a037d76f9719e6/sdcflows/transform.py#L286-L292)

Would you like to become a proud contributor to nitransforms? :D

If not (or either way, actually), may I ask you to file an issue requesting this to be implemented there?

niworkflows/workflows/tests/test_gradunwarp.py
```diff
@@ -0,0 +1,45 @@
...
+    tmpdir.chdir()
+
+    wf = init_gradunwarp_wf()
+    if workdir:
+        wf.base_dir = str(workdir)
+
+    print(ds000030_dir)
+    wf.inputs.inputnode.input_file = [
+        str(f) for f in (ds000030_dir / "sub-10228" / "anat").glob("*_T1w.nii.gz")
+    ][0]
```
effigies, line 42: If we just need the first, don't bother building a list.

```suggestion
    wf.inputs.inputnode.input_file = str(next(ds000030_dir.glob("sub-10228/anat/*_T1w.nii.gz")))
```

niworkflows/workflows/tests/test_gradunwarp.py
```diff
@@ -0,0 +1,45 @@
...
+    wf.inputs.inputnode.grad_file = load_test_data('gradunwarp_coeffs.grad')
```
effigies, line 43: I think this needs to be a `str`, and you'll get a `Path` by default.

```suggestion
    wf.inputs.inputnode.grad_file = str(load_test_data('gradunwarp_coeffs.grad'))
```

effigies (commented):
Mostly comments on the nipype interface.

bpinsard:
I don't understand why I am getting errors on the T1w file when testing the gradunwarp pipeline. How does that ds0000030 differs from the openeuro one?

PR #831, opened 2023-10-13
STY: Assorted refurb suggestions

DimitriPapadopoulos:
* remove no-ops (`return` at end of function and `x = x`)
* use list comprehensions
* `utcnow()` → `now(tz=timezone.utc)`
   From https://docs.python.org/3/library/datetime.html#datetime.datetime.utcnow:
  > **Warning:** Because naive `datetime` objects are treated by many `datetime` methods as local times, it is preferred to use aware datetimes to represent times in UTC. As such, the recommended way to create an object representing the current time in UTC is by calling `datetime.now(timezone.utc)`.
  > 
  > _Deprecated since version 3.12:_ Use [`datetime.now()`](https://docs.python.org/3/library/datetime.html#datetime.datetime.now) with [`UTC`](https://docs.python.org/3/library/datetime.html#datetime.UTC) instead.

### DIFF
--- a/niworkflows/tests/conftest.py
+++ b/niworkflows/tests/conftest.py
@@ -23,7 +23,7 @@
 """ py.test configuration file """
 import os
 from pathlib import Path
-from datetime import datetime as dt
+import datetime as dt
 import pytest
 from templateflow.api import get as get_template
 from niworkflows.testing import test_data_env, data_env_canary
@@ -34,7 +34,7 @@ datadir = load_test_data()
 
 def _run_interface_mock(objekt, runtime):
     runtime.returncode = 0
-    runtime.endTime = dt.isoformat(dt.utcnow())
+    runtime.endTime = dt.datetime.isoformat(dt.datetime.now(dt.timezone.utc))
 
     objekt._out_report = os.path.abspath(objekt.inputs.out_report)
     objekt._post_run_hook(runtime)
--- a/niworkflows/viz/utils.py
+++ b/niworkflows/viz/utils.py
@@ -212,10 +212,7 @@ def _3d_in_file(in_file):
 
     in_file = filemanip.filename_to_list(in_file)[0]
 
-    try:
-        in_file = nb.load(in_file)
-    except AttributeError:
-        in_file = in_file
+    in_file = nb.load(in_file)
 
     if len(in_file.shape) == 3:
         return in_file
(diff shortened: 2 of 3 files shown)


niworkflows/tests/conftest.py
```diff
@@ -34,7 +34,7 @@
 
 def _run_interface_mock(objekt, runtime):
     runtime.returncode = 0
-    runtime.endTime = dt.isoformat(dt.utcnow())
+    runtime.endTime = dt.isoformat(dt.now(dt.timezone.utc))
```
effigies, line 37: I think you need to change L26 to:

```diff
-from datetime import datetime as dt
+import datetime as dt
```

And thiis would need to become:

```suggestion
    runtime.endTime = dt.datetime.isoformat(dt.datetime.now(dt.timezone.utc))
```

DimitriPapadopoulos: Ouch! Indeed, sorry about that. I keep mixing up these two levels of `datetime.datetime`.

niworkflows/viz/utils.py
```diff
@@ -215,7 +215,7 @@ def _3d_in_file(in_file):
     try:
         in_file = nb.load(in_file)
     except AttributeError:
-        in_file = in_file
+        pass
```
effigies, line 218: I have no idea how an `AttributeError` would arise here. I think we should just drop the try/except.

DimitriPapadopoulos: It was added in https://github.com/nipreps/niworkflows/commit/9ba1112845142b847b197f7d40a3a8193a347783 without any explanation.

Of the few places that raise an AttributeError in nibabel:

[**nibabel/nibabel/openers.py**](https://github.com/nipy/nibabel/blob/0e03b9403216af238296d5aa7e44f8f64242f896/nibabel/openers.py#L231-L235)
Lines 231 to 235 of [0e03b94](https://github.com/nipy/nibabel/blob/0e03b9403216af238296d5aa7e44f8f64242f896/)
```python
    def readinto(self, buffer: WriteableBuffer, /) -> int | None:
        # Check and raise our own error for type narrowing purposes
        if hasattr(self.fobj, 'readinto'):
            return self.fobj.readinto(buffer)
        raise AttributeError(f'{self.fobj.__class__.__name__} has no attribute "readinto"')
```

DimitriPapadopoulos: Removed the `AttributeError` check.

effigies (approved):

PR #833, opened 2023-11-14
FIX: Add DWI to the default queries of BIDS querying

oesteban:
This PR adds DWI to the queriable suffixes, and also opens a possibility to modify the default queries.

### DIFF
--- a/niworkflows/utils/bids.py
+++ b/niworkflows/utils/bids.py
@@ -263,7 +263,7 @@ def collect_data(
 
     # Special case: multi-echo BOLD, grouping echos
     if (
-        group_me
+        group_echos
         and "bold" in subj_data
         and any(["_echo-" in bold for bold in subj_data["bold"]])
     ):
--- a/niworkflows/utils/bids.py
+++ b/niworkflows/utils/bids.py
@@ -30,6 +30,19 @@ from bids.layout import Query
 from packaging.version import Version
 
 
+DEFAULT_BIDS_QUERIES = {
+    "bold": {"datatype": "func", "suffix": "bold", "part": ["mag", None]},
+    "dwi": {"suffix": "dwi"},
+    "flair": {"datatype": "anat", "suffix": "FLAIR", "part": ["mag", None]},
+    "fmap": {"datatype": "fmap"},
+    "pet": {"suffix": "pet"},
+    "roi": {"datatype": "anat", "suffix": "roi"},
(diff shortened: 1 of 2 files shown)


niworkflows/utils/bids.py
```diff
@@ -223,19 +237,10 @@ def collect_data(
         'return_type': 'file',
         'subject': participant_label,
         'extension': ['.nii', '.nii.gz'],
-        'session': session_id,
+        'session': session_id or Query.OPTIONAL,
```
effigies, line 240: This changes the meaning of None from Query.NONE, which diverges from pybids semantics. I think that's an okay change, but we should check that we never use None in the way now.

oesteban: You are right, didn't think of it. Considering that single session and multisession datasets cannot be mixed up in BIDS, I agree this should be okay.

niworkflows/utils/bids.py
```diff
@@ -151,11 +164,13 @@ def collect_participants(
 def collect_data(
     bids_dir,
     participant_label,
-    session_id=Query.OPTIONAL,
+    session_id=None,
     task=None,
     echo=None,
+    group_me=True,
```
effigies, line 170: Non-obvious parameter name. And please add an entry to the docstring.

```suggestion
    group_multiecho=True,
```

effigies (commented):
I've got I think all of the major nipreps checked out in one directory, and grepping for uses that include `session_id`, we only get nibabies. That one looks safe.

One small comment and this is good to go.

niworkflows/utils/bids.py
oesteban: I believe I did that because the name overloads the function defined below, but I can modify the function name.

effigies: Or `group_echos`?

effigies:
```suggestion
    group_echos=True,
```

niworkflows/utils/bids.py
```diff
@@ -256,7 +262,11 @@ def collect_data(
     }
 
     # Special case: multi-echo BOLD, grouping echos
-    if any(["_echo-" in bold for bold in subj_data["bold"]]):
+    if (
+        group_me
```
oesteban, line 266:
```suggestion
        group_echos
```

PR #835, opened 2023-11-17
RF: Replace Convert3d with nitransforms in MCFLIRT2ITK

effigies:
c3d_affine_tool is a pretty heavy dependency when we have `nitransforms`. This PR removes it, while adding a regression test to make sure that values stay unchanged.

I added a couple transform matrices and the header of an actual boldref to ensure that the test doesn't depend on trivial invariants, but it's not an exhaustive test.


oesteban (approved):
LGTM!

PR #839, opened 2023-11-29
Add ASL support

### LINKED ISSUE #838: Add ASL support

It would be nice to support ASL in the configuration files.

Yes

AFAICT, the locations that need to be updated are:

1. `NONSTANDARD_REFERENCES` in `niworkflows.utils.spaces`
2. `BIDSDataGrabber`
3. `DEFAULT_BIDS_QUERIES` in `niworkflows.utils.bids`
4. The filename patterns in `nipreps.json`

tsalo:
Closes #838.

Changes proposed:

- Add "asl" to list of non-standard spaces.
- Add "asl" field to data collection function (`collect_data`) and class (`BIDSDataGrabber`).
- Add ASL-related filename patterns to the BIDS config file (`nipreps.json`).

### DIFF
--- a/niworkflows/data/nipreps.json
+++ b/niworkflows/data/nipreps.json
@@ -83,7 +83,7 @@
     },
     {
       "name": "fmap",
-      "pattern": "(phasediff|magnitude[1-2]|phase[1-2]|fieldmap|epi|m0scan)\\.nii"
+      "pattern": "(phasediff|magnitude[1-2]|phase[1-2]|fieldmap|epi)\\.nii"
     },
     {
       "name": "datatype",
--- a/niworkflows/data/nipreps.json
+++ b/niworkflows/data/nipreps.json
@@ -83,11 +83,11 @@
     },
     {
       "name": "fmap",
-      "pattern": "(phasediff|magnitude[1-2]|phase[1-2]|fieldmap|epi)\\.nii"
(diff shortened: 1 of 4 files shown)


niworkflows/data/nipreps.json
```diff
@@ -83,11 +83,11 @@
     },
     {
       "name": "fmap",
-      "pattern": "(phasediff|magnitude[1-2]|phase[1-2]|fieldmap|epi)\\.nii"
+      "pattern": "(phasediff|magnitude[1-2]|phase[1-2]|fieldmap|epi|m0scan)\\.nii"
```
effigies, line 86: This creates a weird pseudo-entity `"fmap": "m0scan"`, which duplicates the suffix. This quirk is unlikely to be replicated in other APIs, and I think we're better off eliminating uses of it than adding use cases to it.

tsalo: Sorry, I'm not sure I understand 100%. Are you saying that you'd prefer to recommend to dataset curators that they place reverse phase-encoded M0 scans in the `perf` folder instead of the `fmap` folder, or that I should support this on ASLPrep's side rather than in niworkflows?

effigies: No, just that all this will do is populate an `fmap` entity in the BIDSLayout database. You already can retrieve it with `layout.get(suffix="m0scan", datatype="fmap")`.

Also, we should make it clearer that `m0scan` is a valid suffix for fieldmap files.

effigies (approved):

PR #841, opened 2023-12-04
ENH: Add MRIQC's custom multiproc runner that allows passing the pool

oesteban:
This is useful to minimize the workers' fingerprint.


oesteban:
> Is there a reason not to contribute this upstream to nipype?

None - just my laziness and reducing the maintenance effort of nipype.

There's a quasi-reason in the sense that I had the plan to write a plugin based on asyncio and test it with MRIQC, and this was like the first tiny test of a drop-in plugin implementation.

Can we merge this in here to speed up the deployment of sdcflows (which uses this, so we can test), and at the same time I'll send a PR to nipype?

PR #844, opened 2023-12-08
FIX: Copy header to ensure mask is really aligned

effigies:
For reasons I don't fully understand, the transformation of a template mask for N4 correction could sometimes end up with different qforms and sforms from the reference image. Our attempt to fix up the headers was insufficient, so I'm just copying the whole thing. This fixes the issue, and this image isn't going to be reused in any other context, so the preservation of other header fields seems irrelevant.


effigies:
@oesteban Any objection?

oesteban (approved):
No objections at all - thanks!

PR #843, opened 2023-12-06
Add `ignore_initial_volumes` param to `ConfoundsCorrelationPlot`

tsalo:
Related to https://github.com/nipreps/fmriprep/issues/2614. Once this is merged and there's a new release, I will open a PR to fMRIPrep.

Changes proposed:

- Add an `ignore_initial_volumes` parameter to the `ConfoundsCorrelationPlot` interface.
- Add an `ignore_initial_volumes` parameter to the `confounds_correlation_plot` function.
- Use the new parameter in related smoke tests.

PR #846, opened 2023-12-11
Correctly use `ignore_initial_volumes` in `ConfoundsCorrelationPlot`

tsalo:
Fixes a bug I missed in #843, where the new parameter wasn't actually used.


effigies:
Good catch. I was getting ready to release...

PR #847, opened 2023-12-12
REL: 1.10.0

effigies:
Release prep for 1.10.0.

Please make noise if you need to get something in first.


effigies:
@mgxd The docker builds were taking forever, so I ported over fMRIPrep's, with the necessary changes. Would you mind having a look?

PR #867, opened 2024-04-17
STY: Apply assorted ruff/refurb rules

DimitriPapadopoulos:


niworkflows/interfaces/confounds.py
```diff
@@ -295,7 +295,7 @@ def spike_regressors(
             mask[metric] = set(np.where(data[metric] < threshold)[0])
         elif criterion == ">":
             mask[metric] = set(np.where(data[metric] > threshold)[0])
-    mask = reduce((lambda x, y: x | y), mask.values())
+    mask = reduce(operator.or_, mask.values())
```
effigies, line 298: Import `operator`?

DimitriPapadopoulos: Yes, it might help :smile:

effigies (approved):

PR #876, opened 2024-06-20
FIX: Remove accidental MRIQC dependency, allow app config to be passed to workflow plugin

effigies:
Reverted some uses of `config` to the nipype config, which an app should generally be injecting into the nipype workflow config.

Some parts don't exist, so I'm allowing an app config to be passed as a plugin argument "app_config", and falling back to nipype defaults or trivial replacements.

Closes #875.

PR #880, opened 2024-07-18
TEST: Fix plugin invocation, use an initializer that can be verified

effigies:
Follow-up to #876, which didn't actually test what we wanted it to. We were manually calling `print()` inside the test and interpreting that as being run on process initialization. We need to pass the config to `MultiProcPlugin()`, not `workflow.run()`.

`capsys` does not capture stdout from subprocesses, it seems, so instead I'm telling the subprocesses to touch a file that I choose.

### DIFF
--- a/niworkflows/engine/tests/test_plugin.py
+++ b/niworkflows/engine/tests/test_plugin.py
@@ -75,7 +75,7 @@ def test_plugin_args_noconfig(workflow, caplog):
     workflow.run(plugin=MultiProcPlugin(plugin_args={"n_procs": 2, "memory_gb": 0.1}))
 
 
-def touch_file(file_path):
+def touch_file(file_path: str) -> None:
     """Module-level functions play more nicely with multiprocessing."""
     with open(file_path, "w") as f:
         f.write("flag")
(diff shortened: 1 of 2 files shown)


niworkflows/engine/tests/test_plugin.py
```diff
@@ -72,28 +72,28 @@ def test_plugin_defaults(workflow, caplog):
...
-        print("Custom init")
+    init_flag = tmp_path / "init_flag.txt"
 
     app_config = SimpleNamespace(
-        environment=SimpleNamespace(total_memory_gb=1),
-        _process_initializer=init_print(),
-        file_path='/does/not/need/to/exist/for/testing',
+        environment=SimpleNamespace(total_memory=1),
+        _process_initializer=touch_file,
+        file_path=init_flag,
```
mgxd, line 92: https://github.com/nipreps/niworkflows/blob/1dc0327aa10f156fde480ec0b23c61a026b8022a/niworkflows/engine/plugin.py#L449-L454

We should really document somewhere `_process_initializer`'s expected signature - I got burned by this.

mgxd (commented):
good catch

niworkflows/engine/tests/test_plugin.py
effigies: Please see the latest changes.

PR #883, opened 2024-07-23
ENH: Support combined surface-volume reference spaces

### LINKED ISSUE #881: Distinguish between surface and volume spaces in `Reference`

In order to parse surface-volume combinations for CIFTIs, I think `Reference` needs to be able to distinguish between the surface space and the volume space.

This is related to https://github.com/nipreps/fmriprep/issues/3330.

Yes

_No response_

tsalo:
Closes #881. This is a ways away from being usable, but it might be good to get folks' thoughts on it before I put too much effort into it.

Changes proposed:

- Add a delimiter between surface and volume space strings (`::`) to allow combinations of surfaces and volumes for CIFTIs.
- Add volspace and volcohort keys to the `spec` dictionary.
- Add a `cifti` attribute to the Reference object to denote if a space is CIFTI or not.

### DIFF
--- a/niworkflows/utils/spaces.py
+++ b/niworkflows/utils/spaces.py
@@ -148,6 +148,8 @@ class Reference:
     """The dictionary of specs."""
     standard = attr.ib(default=False, repr=False, type=bool)
     """Whether this space is standard or not."""
+    cifti = attr.ib(default=False, repr=False, type=bool)
+    """Whether this space is a CIFTI space or not."""
     dim = attr.ib(default=3, repr=False, type=int)
     """Dimensionality of the sampling manifold."""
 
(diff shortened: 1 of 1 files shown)


niworkflows/utils/spaces.py
```diff
@@ -165,12 +167,32 @@ def __attrs_post_init__(self):
                 spec["den"] = FSAVERAGE_DENSITY[space]
                 object.__setattr__(self, "spec", spec)
 
+        # XXX: This won't cover dhcpAsym, dhcpSym, or onavg
         if self.space.startswith("fs"):
             object.__setattr__(self, "dim", 2)
 
+        if "volspace" in self.spec:
```
tsalo, line 174: This check would hopefully only be necessary in the short-term, since mixed standard/non-standard CIFTIs should be allowed.

niworkflows/utils/spaces.py
```diff
@@ -148,6 +148,8 @@ class Reference:
     """The dictionary of specs."""
     standard = attr.ib(default=False, repr=False, type=bool)
     """Whether this space is standard or not."""
+    cifti = attr.ib(default=False, repr=False, type=bool)
+    """Whether this space is a CIFTI space or not."""
```
tsalo, line 152: Alternatively, I could use a new `dim` value, like 2.5, instead of adding this attribute.

PR #885, opened 2024-07-29
ENH: Add PrepareDerivative/SaveDerivative interfaces

effigies:
These are intended to break DerivativesDataSink functionality into separable components and avoid unnecessary writes that cause invalidation of downstream nodes.

Tests forthcoming.


niworkflows/interfaces/bids.py
```diff
@@ -288,6 +289,493 @@ def _run_interface(self, runtime):
...
+    out_path = OutputMultiObject(Str, desc="relative path in target directory")
+    fixed_hdr = traits.List(traits.Bool, desc="whether derivative header was fixed")
+
+
+class PrepareDerivative(SimpleInterface):
+    """
+    Prepare derivative files and metadata.
+
+    Collects entities from source files and inputs, filters them for allowed entities,
+    and passes to the out_entities output.
```
mgxd, line 328: `out_entities` is not an output field - should it be?

niworkflows/interfaces/bids.py
```diff
@@ -288,6 +289,493 @@ def _run_interface(self, runtime):
...
+                out_hash = hash_infile(out_file)
+                if in_hash != out_hash:
+                    out_file.unlink()
+
+            if not out_file.exists():
+                shutil.copyfile(in_file, out_file)
+
+            if metadata:
+                sidecar = out_file.parent / f"{out_file.name.split('.', 1)[0]}.json"
+                unlink(sidecar, missing_ok=True)
```
mgxd, line 771:
```suggestion
                sidecar.unlink(missing_ok=True)
```

niworkflows/interfaces/bids.py
```diff
@@ -288,6 +289,493 @@ def _run_interface(self, runtime):
...
+    metadata = InputMultiObject(
+        traits.DictStrAny, desc="metadata to be saved alongside the file"
+    )
+    relative_path = InputMultiObject(
+        traits.Str, desc="path to the file relative to the base directory"
+    )
+
+class _SaveDerivativeOutputSpec(TraitedSpec):
+    out_file = OutputMultiObject(File(exists=True), desc="written file path")
+    out_meta = OutputMultiObject(File(exists=True), desc="written JSON sidecar path")
```
mgxd, line 733: i remember removing `exists` from outputs because of potential caching issues - has that been confirmed / resolved?

mgxd (commented):
Big fan of deconstructing the monolith

niworkflows/interfaces/bids.py
effigies: Yeah, we should remove the exists check.

niworkflows/interfaces/tests/test_bids.py
```diff
@@ -291,24 +295,41 @@ def test_DerivativesDataSink_build_path(
 
         ds_inputs.append(str(fname))
 
-    dds = bintfs.DerivativesDataSink(
+    base_directory = tmp_path / "output"
+    work_dir = tmp_path / "work"
+    base_directory.mkdir()
+    work_dir.mkdir()
```
effigies, line 301: Unused, but I don't want to start CI over again. Apply before merging.

```suggestion
    base_directory.mkdir()
```

effigies:
@mgxd I think this is ready to go. I expanded the tests to check everything we were checking with `DerivativesDataSink`.

mgxd:
Yup, looks good to me!

PR #891, opened 2024-08-21
ENH: Add volspace and volcohort entities to IO config file

tsalo:
Closes none, but relates to #883, https://github.com/bids-standard/bids-specification/pull/1900, and https://github.com/nipreps/fmriprep/issues/3341.

Changes proposed:

- Add volspace and volcohort entities to the IO config file.
- Allow volspace, volcohort, and resolution in CIFTI filenames.

### DIFF
--- a/niworkflows/data/nipreps.json
+++ b/niworkflows/data/nipreps.json
@@ -74,6 +74,10 @@
       "pattern": "[_/\\\\]+space-([a-zA-Z0-9\\+]+)"
     },
     {
+      "name": "volspace",
+      "pattern": "[_/\\\\]+volspace-([a-zA-Z0-9\\+]+)"
+    },
+    {
       "name": "suffix",
       "pattern": "[._]*([a-zA-Z0-9]*?)\\.[^/\\\\]+$"
     },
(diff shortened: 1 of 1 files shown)


tsalo:
@effigies @mgxd would you mind taking a look at this at some point?

niworkflows/data/nipreps.json
```diff
@@ -149,13 +153,18 @@
       "name": "cohort",
       "pattern": "[_/\\\\]+cohort-0*(\\d+)",
       "dtype": "int"
+    },
+    {
+      "name": "volcohort",
+      "pattern": "[_/\\\\]+volcohort-0*(\\d+)",
+      "dtype": "int"
```
mgxd, line 160: never noticed this until now, but why are these 0 padded?

mgxd (approved):
LGTM

tsalo:
An alternative proposed by @effigies in https://github.com/bids-standard/bids-specification/pull/1900 involves using `+` in the `space` entity instead of a new entity, so that's a viable alternative.

niworkflows/data/nipreps.json
effigies: I believe an earlier version of pybids had something like that for run, so it parsed `int("1")` instead of `int("01")`. I don't think it was necessary or that it is still there.

tsalo:
I'm committing to just using `+`s, so I'm going to close this.

PR #892, opened 2024-08-24
STY: Use f-string within normalization interface

oesteban:


effigies (approved):
Not merging to reduce CI load. Let's merge batches of approved PRs at a time, only when there's a need to get something into the branch.

PR #898, opened 2024-11-08
Add optional dir entity to T2starmap pattern

tsalo:
I am currently processing multi-echo BOLD data with both AP and PA acquisitions through fMRIPrep, and I noticed that the T2starmap outputs didn't include the direction entity, so I was only getting one of the two directions for each acquisition.

I also made task optional for T2starmap, but I can revert that.


effigies (approved):
LGTM

PR #900, opened 2024-11-15
chore: Switch to tox for testing and CI

effigies:


effigies:
Good enough for now. I would like to update how we're tagging skip-on-missing-dependencies so we can limit our CircleCI usage to just the relevant bits, but that can be another PR.

PR #911, opened 2024-12-11
ENH: Allow adjusting histogram matching of configuration files

mgxd:


niworkflows/interfaces/norm.py
```diff
@@ -196,6 +197,15 @@ def _run_interface(self, runtime):
             NIWORKFLOWS_LOG.info('Loading settings from file %s.', ants_settings)
             # Configure an ANTs run based on these settings.
             self.norm = Registration(from_file=ants_settings, **ants_args)
+            if isdefined(self.inputs.use_histogram_matching):
```
effigies, line 200: To consider (and reformat if accepted): Is it worth logging only when the override changes something, or always?

```suggestion
            if isdefined(self.inputs.use_histogram_matching) and not self.norm.inputs.use_histogram_matching == self.inputs.use_histogram_matching:
```

PR #915, opened 2024-12-21
SYT: Apply ruff/pygrep-hooks rule PGH003

DimitriPapadopoulos:
PGH003 Use specific rule codes when ignoring type issues


effigies (approved):
Seems fine. It was probably copied from a codebase that does use type checking, but since we don't, we can omit this until we do.

PR #918, opened 2024-12-21
STY: Apply ruff/refurb rules (FURB)

DimitriPapadopoulos:


niworkflows/engine/tests/test_plugin.py
```diff
@@ -28,7 +25,7 @@ def workflow(tmp_path):
 
     # Generate many nodes and claim a lot of memory
     add_nd = pe.MapNode(
-        niu.Function(function=add, input_names=['x', 'y'], output_names=['z']),
+        niu.Function(function=operator.add, input_names=['x', 'y'], output_names=['z']),
```
effigies, line 28: The `Function` interface can't support builtin (C) functions. This file needs to be reverted.

PR #922, opened 2025-01-09
ENH: Parse kwargs to allow no T1w

mgxd:
This came up in https://github.com/nipreps/nibabies/issues/411

Also, adds a default to the expected kwarg `anat_only` to allow bare initialization

### DIFF
--- a/niworkflows/interfaces/bids.py
+++ b/niworkflows/interfaces/bids.py
@@ -253,18 +253,11 @@ class BIDSDataGrabber(SimpleInterface):
     >>> bids_src.inputs.subject_data = bids_collect_data(
     ...     str(datadir / 'ds114'), '01', bids_validate=False)[0]
     >>> bids_src.inputs.subject_id = '01'
-    >>> bids_src._require_t1w
-    True
-    >>> bids_src._require_funcs
-    True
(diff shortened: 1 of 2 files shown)


niworkflows/interfaces/bids.py
```diff
@@ -253,24 +253,32 @@ class BIDSDataGrabber(SimpleInterface):
...
+    >>> bids_src._require_funcs
+    True
     >>> res = bids_src.run()
     >>> res.outputs.t1w  # doctest: +ELLIPSIS +NORMALIZE_WHITESPACE
     ['.../ds114/sub-01/ses-retest/anat/sub-01_ses-retest_T1w.nii.gz',
      '.../ds114/sub-01/ses-test/anat/sub-01_ses-test_T1w.nii.gz']
 
+    >>> bids_src = BIDSDataGrabber(require_t1w=False)
+    >>> bids_src._require_t1w
+    False
```
effigies, line 267: Private variables aren't part of the public API so shouldn't be in doctests.

mgxd: moved

effigies (approved):

PR #927, opened 2025-02-25
Remove extra plus signs in regular expressions

tsalo:
Addresses https://github.com/nipreps/niworkflows/pull/926#discussion_r1970280359.


mgxd (commented):
im not sure why i included `//` in https://github.com/nipreps/niworkflows/pull/857 ...wdyt about removing that and just mirroring your changes to the other entities?

tsalo:
Sounds good. I'll do that now.

Issue #928, opened 2025-03-01
`brain_extraction_wf.inu_n4_final` can be affected by extreme values


effigies:
### What happened?

I've been debugging an Atropos crash that occurs for `brain_extraction_wf` in niworkflows 1.3.x (LTS) and 1.12.x (dev) but does not occur for `antsBrainExtraction.sh` in ANTs 2.3.4 (LTS) or 2.5.4 (latest).

Walking through step-by-step, here are the abbreviated inputs to Atropos:

#### `antsBrainExtraction.sh`:

```mermaid
graph TD;
    T1w --"TruncateImageIntensity"--> maskedN4Truncated0.nii.gz	--N4--> maskedN4Corrected0.nii.gz --> Atropos
    probseg --"resample and threshold"---> maskedBrainExtractionPriorWarped.nii.gz --> Atropos
```

#### `brain_extraction_wf`:

```mermaid
graph TD;
    T1w --"TruncateImageIntensity"--> trunc --N4--> inu_n4
    T1w --N4----> inu_n4_final --> Atropos
    thr_brainmask --N4--> inu_n4_final
    probseg --resample--> map_brainmask --threshold--> thr_brainmask --> Atropos
```

There is no equivalent to `inu_n4_final` in the `antsBrainExtraction.sh` workflow, so that seemed to be the most likely culprit. Re-running with verbose:

```
Running Atropos for 3-dimensional images.

Progress: 

itk::ExceptionObject (0x561d6f98af30)
Location: "unknown" 
File: /home/conda/feedstock_root/build_artifacts/ants_1732704629105/work/ImageSegmentation/antsGaussianListSampleFunction.hxx
Line: 100
Description: ITK ERROR: GaussianListSampleFunction(0x561d6f98c2b0): Covariance is singular (determinant = 8.91523e-07 < 1.0e-6)
```

If I switch the `inu_n4_final` image for `inu_n4` in the same shell command, the run succeeds. By taking the difference of the original image and the truncated image, and using the mask passed to atropos, we can see that an extreme value lies just inside the mask:

![Image](https://github.com/user-attachments/assets/dc0d5726-4526-4c53-85de-770ae8d4b64e)

None of the extreme values occur inside brain tissue, so it does not seem problematic to pass the truncated image to `inu_n4_final`, like:

```mermaid
graph TD;
    T1w --"TruncateImageIntensity"--> trunc --N4--> inu_n4
    trunc --N4---> inu_n4_final --> Atropos
[...]

PR #929, opened 2025-03-01
fix(brain_extraction_wf): Pass inu_n4 to atropos_wf

effigies:
antsBrainExtraction.sh passes the original N4 corrected image to Atropos, where we have been running a masked N4 before Atropos. We have examples where this difference leads to failures that do not occur in the original ANTs workflow. Further, if we do run Atropos, then this result is discarded and N4 is run using a white-matter mask.

This patch therefore simply passes the original N4 image to the Atropos workflow, but otherwise leaves the workflow unchanged.

Addresses #928. This targets the LTS branch `maint/1.3.x`. Will leave that issue open until this or another patch is merged into `master`.

### DIFF
--- a/niworkflows/anat/ants.py
+++ b/niworkflows/anat/ants.py
@@ -459,7 +459,7 @@ def init_brain_extraction_wf(
         ])
         wf.connect([
             (inputnode, atropos_wf, [("in_files", "inputnode.in_files")]),
-            (inu_n4_final, atropos_wf, [("output_image", "inputnode.in_corrected")]),
+            (inu_n4, atropos_wf, [("output_image", "inputnode.in_corrected")]),
             (thr_brainmask, atropos_wf, [("output_image", "inputnode.in_mask")]),
             (atropos_wf, outputnode, [
                 ("outputnode.out_file", "out_file"),


effigies:
@yohanchatelain I wonder if it would be possible for you to see if this change would have a significant impact on the metrics that you were calculating for the paper. You should be able to use the following Dockerfile:

```dockerfile
FROM nipreps/fmriprep:20.2.8

pip install --no-cache-dir git+https://github.com/nipreps/niworkflows.git@refs/pull/929/head
```

I believe it should be minimal, but if we could confirm/disconfirm that, it would help justify whether we can make a release in the LTS series or if I need to find some other way to make this patch available to users.

effigies:
@bpinsard @pbellec Are you still using fMRIPrep LTS? I don't know if you have any process for testing whether a new release is similar enough to an old release.

yohanchatelain:
Hi @effigies,
Sure, I can do it. Which version should we compare 20.2.8 to? The previous version, 20.2.7?

effigies:
Just to be clear, 20.2.8 was released last July, so this will be a patch on that. If you don't have stats for that, however, comparing to 20.2.7 would be fine.

yohanchatelain:
Ok I see, so we need to check this patch against the 20.2.8 release version.
I'll launch that.

effigies:
@yohanchatelain Just checking in. Do you have an estimate of how long it will take to analyze the changes?

yohanchatelain:
Hi @effigies, the longest subject takes approximately 5 days to run with fuzzy, so I should have the results next week.

effigies:
@yohanchatelain Just checking in today since I'll be out tomorrow.

yohanchatelain:
@effigies, I’ve finished the numerical stability test. [Here](https://github.com/yohanchatelain/fmriprep-reproducibility/blob/master/pull-929/plot.ipynb) is the notebook summarising the results.

The reference was built from 30 Random Rounding repetitions computed with fMRIPrep 20.2.8, and the test set consists of 1 IEEE execution with 20.2.8_PR-929 for each subject used in the paper.

- The 20.2.8 PR#929 does **not** pass the test when using version 20.2.8 as the reference.  
  image

- Additionally, I’d say the numerical quality is slightly better in version 20.2.8 than in 20.2.1.  
  image

Here are the results for version 20.2.1 from the paper:  
image

effigies:
Thanks @yohanchatelain. That was very helpful, and we declined to try to include this in an LTS release.

PR #931, opened 2025-03-03
fix(brain_extraction): Remove duplicated code, problematic deviation from antsBrainExtraction

effigies:
As discussed in #928, running a masked N4 before calling Atropos can lead to failures, although I don't fully understand why. Nonetheless, passing the output of the first N4 correction resolves the issue in the case I'm looking at as well as brings the workflow into line with `antsBrainExtraction.sh`, which does not do this step.

Furthermore, if `atropos_wf` is called, then N4 is called using the white matter mask at the end, which means this is duplicated effort. This PR rearranges the workflow so that the second top-level N4 is only called if `atropos_wf` isn't created. It also shifts from the pattern:

```py
wf.connect(A)
if condition:
  wf.disconnect(A)
  wf.connect(B)
```

to:

```py
if condition:
  wf.connect(B)
else:
  wf.connect(A)
```

I personally find this logic easier to follow and compare between the conditions.

One thing that I discovered was that the `lap_{tmpl,target}` and `mrg_{tmpl,target}` nodes were defined twice, but only connected if `use_laplacian`. Here we always define and connect `mrg_{tmpl,target}`, and only create and connect `lap_{tmpl,target}` if `use_laplacian`.

### DIFF
--- a/niworkflows/anat/ants.py
+++ b/niworkflows/anat/ants.py
@@ -334,37 +334,8 @@ def init_brain_extraction_wf(
         name='thr_brainmask',
     )
 
-    # Refine INU correction
-    inu_n4_final = pe.MapNode(
-        N4BiasFieldCorrection(
-            dimension=3,
-            save_bias=True,
-            copy_header=True,
-            n_iterations=[50] * 5,
-            convergence_threshold=1e-7,
-            shrink_factor=4,
-            bspline_fitting_distance=bspline_fitting_distance,
-        ),
-        n_procs=omp_nthreads,
-        name='inu_n4_final',
-        iterfield=['input_image'],
-    )
-    try:
-        inu_n4_final.inputs.rescale_intensities = True
-    except ValueError:
-        warn(
-            "N4BiasFieldCorrection's --rescale-intensities option was added in ANTS 2.1.0 "
-            f'({inu_n4_final.interface.version} found.) Please consider upgrading.',
-            UserWarning,
-            stacklevel=1,
-        )
-
-    # Apply mask
-    apply_mask = pe.MapNode(ApplyMask(), iterfield=['in_file'], name='apply_mask')
-
     wf.connect([
         (inputnode, trunc, [('in_files', 'op1')]),
-        (inputnode, inu_n4_final, [('in_files', 'input_image')]),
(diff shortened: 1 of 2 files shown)


effigies:
@mgxd I would appreciate a review here.

mgxd (commented):
Looks reasonable to me - tests inspire confidence.

> I personally find this logic easier to follow and compare between the conditions.
Agreed, though I noticed there are still 2 `disconnect`s remaining. Do you want to tackle that refactor in this PR as well?

effigies:
> Agreed, though I noticed there are still 2 disconnects remaining. Do you want to tackle that refactor in this PR as well?

Sure. I also changed `inu_n4_final` to `inu_n4` in the workflow where it's only done once.

niworkflows/anat/ants.py
```diff
@@ -486,14 +413,53 @@ def init_brain_extraction_wf(
...
+        # Check ANTs version
+        try:
+            inu_n4_final.inputs.rescale_intensities = True
+        except ValueError:
+            warn(
+                "N4BiasFieldCorrection's --rescale-intensities option was added in ANTS 2.1.0 "
+                f'({inu_n4.interface.version} found.) Please consider upgrading.',
+                UserWarning,
+                stacklevel=1,
+            )
```
effigies, line 445: Rebased to re-add this check, which was removed in passing. If we want to remove these, that can be a separate PR.

effigies:
@mgxd Good to go?

niworkflows/anat/ants.py
```diff
@@ -486,14 +413,53 @@ def init_brain_extraction_wf(
...
+            iterfield=['input_image'],
+        )
+
+        # Check ANTs version
+        try:
+            inu_n4_final.inputs.rescale_intensities = True
+        except ValueError:
+            warn(
+                "N4BiasFieldCorrection's --rescale-intensities option was added in ANTS 2.1.0 "
+                f'({inu_n4.interface.version} found.) Please consider upgrading.',
```
mgxd, line 442: not that it really matters but
```suggestion
                f'({inu_n4_final.interface.version} found.) Please consider upgrading.',
```

mgxd (approved):
very minor commment, LGTM

PR #938, opened 2025-05-01
ENH: Workflow splicer module

mgxd:
This PR lays some of the groundwork for fMRIPrep extensions.

`splice_workflow` splices a constructed workflow, allowing developers to swap in alternative processing workflows. This works by finding all tagged `Node`/`Workflow` in a graph, and substituting the matching `replacements` workflow. This will require some work in the NiPreps to encapsulate each processing step, likely with a heavy dosing of buffernodes for input and output fields.

Additionally, the `@tag` decorator can be used on any `init_...wf()` factory to facilitate adding the `_tag` attribute.

### DIFF
--- a/niworkflows/engine/splicer.py
+++ b/niworkflows/engine/splicer.py
@@ -1,0 +1,125 @@
+"""Workflow splicing operations."""
+
+import typing as ty
+
+from nipype.pipeline import Workflow
+from nipype.pipeline.engine.base import EngineBase
+
+
+def tag(tag: str) -> 'EngineBase':
+    """
+    Decorator to set a tag on an `init_...wf` function.
+
+    This is used to mark nodes or workflows for replacement in the splicing process.
+    """
+
+    def _decorator(func, *args, **kwargs) -> ty.Callable:
+        def _tag() -> 'EngineBase':
+            node = func(*args, **kwargs)
+            node._tag = tag
+            return node
+
+        return _tag
+
+    return _decorator
+
+
+def splice_workflow(
+    root_wf: Workflow,
+    replacements: dict[str, 'EngineBase'],
+    *,
+    write_graph: bool = False,
+    debug: bool = False,
+):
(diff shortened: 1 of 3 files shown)


niworkflows/engine/splicer.py
```diff
@@ -0,0 +1,123 @@
+"""Workflow splicing operations."""
+
+import typing as ty
+
+from nipype.pipeline import Workflow
```
oesteban, line 5: would this not conflict with the `LiterateWorkflow` defined in workflow? I think we want to change the import in `.workflow` so it imports from here instead of form nipype.

niworkflows/engine/splicer.py
```diff
@@ -0,0 +1,123 @@
...
+            _expanded_workflows.add(dst)
+            _splice_components(dst, substitutions, debug=debug)
+        elif isinstance(src, Workflow) and src not in _expanded_workflows:
+            _expanded_workflows.add(src)
+            _splice_components(src, substitutions, debug=debug)
+
+    if debug:
+        print(f'Workflow: {workflow}')
+        print(f'- Removing: {edge_removals}')
+        print(f'+ Adding: {edge_connects}')
```
effigies, line 118: `logger.debug()`...

oesteban (commented):
Trying to wrap my head around this. I left a couple of comments.

niworkflows/engine/splicer.py
```diff
@@ -0,0 +1,123 @@
...
+from nipype.pipeline.engine.base import EngineBase
+
+
+def tag(tag: str) -> 'EngineBase':
+    """
+    Decorator to set a tag on an `init_...wf` function.
+
+    This is used to mark nodes or workflows for replacement in the splicing process.
+    """
+
```
effigies, line 15: This function is unused/untested, and it looks kind of wrong to me, so a test seems critical.

PR #951, opened 2025-05-26
ENH: Toggle requirement for any image type BIDSDataGrabber, add task filtering when collecting PET data

mnoergaard:
Needed to update this part to be able to run PETPrep without the workflow looking for functional data, when not using the --anat-only flag.

### DIFF
--- a/niworkflows/utils/bids.py
+++ b/niworkflows/utils/bids.py
@@ -35,11 +35,9 @@ from packaging.version import Version
 
 DEFAULT_BIDS_QUERIES = {
     'bold': {
-        'datatype': 'func', 
-        'suffix': 'bold', 
+        'datatype': 'func',
+        'suffix': 'bold',
         'part': ['mag', None],
-        'task': Query.OPTIONAL,
-        'echo': Query.OPTIONAL,
     },
     'dwi': {'suffix': 'dwi'},
     'flair': {'datatype': 'anat', 'suffix': 'FLAIR', 'part': ['mag', None]},
@@ -257,6 +255,7 @@ def collect_data(
         'return_type': 'file',
         'subject': participant_label,
         'extension': ['.nii', '.nii.gz'],
+        'session': session_id or Query.OPTIONAL,
     }
 
     reserved_entities = [('subject', participant_label), ('session', session_id)]
@@ -277,7 +276,7 @@ def collect_data(
                 del layout_get_kwargs[entity]
 
     if task:
-        queries['bold']['task'] = task
+        queries['bold']['task'] = queries['pet']['task'] = task
 
     if echo:
         queries['bold']['echo'] = echo


niworkflows/utils/bids.py
```diff
@@ -41,7 +41,7 @@
     'pet': {'suffix': 'pet'},
     'roi': {'datatype': 'anat', 'suffix': 'roi'},
     'sbref': {'datatype': 'func', 'suffix': 'sbref', 'part': ['mag', None]},
-    't1w': {'datatype': 'anat', 'suffix': 'T1w', 'part': ['mag', None]},
+    't1w': {'suffix': 'T1w'},
```
mgxd, line 44: rather than change the default queries, you can pass in alternative queries by using `collect_data(..., queries=x)`

mgxd (commented):
This breaks compatibility with other tools depending on these - rather than this, we should add deviations where necessary.

mgxd:
@mnoergaard I cleaned up the BIDSDataGrabber logic in [42cb1a2](https://github.com/nipreps/niworkflows/pull/951/commits/42cb1a2424e4ce7a9d95bee39dca3d3e40ce649b) - mainly generalizing `require_*` to encompass all the image types we support. A few changes of note: 
- renamed `require_funcs` to `require_bold`, which is more in-line with the query and the other types.
- `require_pet` will not implicitly disable `require_bold` - they should be controlled separately

niworkflows/utils/bids.py
mgxd: This should be reverted

niworkflows/utils/bids.py
```diff
@@ -251,8 +251,9 @@ def collect_data(
         'return_type': 'file',
         'subject': participant_label,
         'extension': ['.nii', '.nii.gz'],
-        'session': session_id or Query.OPTIONAL,
     }
+    if session_id is not None:
+        layout_get_kwargs['session'] = session_id
```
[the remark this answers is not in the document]
mgxd: can we revert this?

niworkflows/utils/bids.py
```diff
@@ -271,20 +272,10 @@ def collect_data(
                 # avoid clobbering layout.get
                 del layout_get_kwargs[entity]
 
-    if task:
-        queries['bold']['task'] = task
-
-    if echo:
-        queries['bold']['echo'] = echo
-
```
mgxd, line 279: rather than remove this, the `task` entity should also be added to the bids query

niworkflows/utils/bids.py
```diff
@@ -271,20 +272,10 @@ def collect_data(
...
     subj_data = {
         dtype: sorted(layout.get(**layout_get_kwargs, **query)) for dtype, query in queries.items()
     }
 
-    # Special case: multi-echo BOLD, grouping echos
-    if group_echos and 'bold' in subj_data and any('_echo-' in bold for bold in subj_data['bold']):
-        subj_data['bold'] = group_multiecho(subj_data['bold'])
-
```
mgxd, line 287: This should be reverted

mgxd (dismissed):

mgxd (approved):
Looks good to me, but I'll let you test the petprep integration prior to merging 😃

mnoergaard:
Awesome - many thanks, @mgxd ! I will go ahead and test it and get back to you. Currently running things with the updated version from 4 hours ago. So good so far!

PR #1003, opened 2025-12-01
ENH: Add target resolution to ``GenerateSamplingReference``

### LINKED ISSUE #1002: Support target resolution input to GenerateSamplingReference

This would allow workflows to write out images in a target space, with any requested resolution.

Yes, but I would need guidance

I do not have a good handle on how GenerateSamplingReference/_gen_reference works, so I would need help with this.

tsalo:
This is a quick attempt at #1002. Closes #1002.

Changes proposed:

- Add a `target_resolution` parameter to `GenerateSamplingReference` and `_gen_reference`.
- Create a target shape and target affine from the `moving_image` and target resolution when requested. Then resample the moving image to that target resolution.


niworkflows/interfaces/nibabel.py
```diff
@@ -586,14 +679,24 @@ def _gen_reference(
...
+    if target_resolution is not None:
+        new_affine, new_shape = _calculate_target_affine(reoriented_moving_img, target_resolution)
+    else:
+        new_zooms = reoriented_moving_img.header.get_zooms()[:3]
+
+        # Avoid small differences in reported resolution to cause changes to
+        # FOV. See https://github.com/nipreps/fmriprep/issues/512
+        # A positive diagonal affine is RAS, hence the need to reorient above.
+        new_affine = np.diag(np.round(new_zooms, 3))
+        new_shape = fixed_image.shape[:3]
```
effigies: Well, now I'm looking at this again, and we were previously enforcing a diagonal matrix, and letting nilearn figure out the translations and bounding boxes. Can we just do that again with:

```suggestion
    new_zooms = target_resolution if target_resolution is not None else reoriented_moving_img.header.get_zooms()[:3]

    # Avoid small differences in reported resolution to cause changes to
    # FOV. See https://github.com/nipreps/fmriprep/issues/512
    # A positive diagonal affine is RAS, hence the need to reorient above.
    new_affine = np.diag(np.round(new_zooms, 3))

    resampled = nli.resample_img(fixed_image, target_affine=new_affine, interpolation='nearest')
```

Apologies for bouncing around. My head is all over the place right now, so please feel free to argue back...

niworkflows/interfaces/nibabel.py
```diff
@@ -595,6 +693,13 @@ def _gen_reference(
 
     resampled = nli.resample_img(fixed_image, target_affine=new_affine, interpolation='nearest')
 
+    resampled = nli.resample_img(
+        fixed_image,
+        target_affine=new_affine,
+        target_shape=new_shape,
+        interpolation='nearest',
+    )
+
```
effigies:
```suggestion
```

effigies (approved):

oesteban (approved):

Issue #1018, opened 2026-01-22
[BUG] Fix/Update Entity Order for DerivativesDataSink Object


bendhouseart:
### What happened?

Discovered [this issue](https://github.com/nipreps/petprep/issues/231) reviewing some niprep/petprep outputs for bep 023. It seems like the entities recorded in the [this nipreps.json](https://github.com/nipreps/niworkflows/blob/9e18356d7a7c8c6b81eedb1bcee8ea612b8fe579/niworkflows/data/nipreps.json) file need to be updated to fix.

```
sub-01_ses-baseline_desc-preproc_seg-gtm_tacs.json
sub-01_ses-baseline_desc-preproc_seg-gtm_tacs.tsv
```

Should be:

```
sub-01_ses-baseline_seg-gtm_desc-preproc_tacs.json
sub-01_ses-baseline_seg-gtm_desc-preproc_tacs.tsv
```   

### What command did you use?

```shell
ds_pet_tacs = pe.Node(
        DerivativesDataSink(
            base_directory=petprep_dir,
            suffix='tacs',
            seg=config.workflow.seg,
            desc='preproc',
            allowed_entities=('seg',),
            TaskName=all_metadata[0].get('TaskName'),
            **prepare_timing_parameters(all_metadata[0]),
        ),
        name='ds_pet_tacs',
        run_without_submitting=True,
        mem_gb=config.DEFAULT_MEMORY_MIN_GB,
    )

```

link to above [here](https://github.com/nipreps/petprep/blob/0797cfcb4401d7400e5034d43e473a699bfda130/petprep/workflows/pet/base.py#L728)

### What version of the software are you running?

>= 1.14.2

### How are you running this software?

Docker

### Is your data BIDS valid?

Yes

### Are you reusing any previously computed results?

No

### Please copy and paste any relevant log output.

```shell

```

effigies:
This would be an opportunity to switch to using the schema...

But the issue is that this line doesn't have `seg` in and the interface sticks new entities at the end.

https://github.com/nipreps/niworkflows/blob/9e18356d7a7c8c6b81eedb1bcee8ea612b8fe579/niworkflows/data/nipreps.json#L202C1-L202C62

https://github.com/nipreps/niworkflows/blob/9e18356d7a7c8c6b81eedb1bcee8ea612b8fe579/niworkflows/interfaces/bids.py#L595-L601

PR #1020, opened 2026-01-29
[FIX] Entity order for derivatives data sink

bendhouseart:
Closes issue #1018, this is the easier path to fix this. Only updates the entities touched by petprep. Please feel free to request/push changes to update other file name templates.


effigies:
hash should stay where it is, after session. @mgxd added it to allow for use in a BIDS-adjacent database context, so we should not mess with it.

PR #1057, opened 2026-09-02
fix: disable colorbar after default switch

mgxd:


effigies (approved):
Hmm. Is it time to delete this module?

PR #1058, opened 2026-09-08
Drop bids_filters entries with no matching query

### LINKED ISSUE #862: `niworkflows.utils.bids.collect_data` should drop unnecessary filters
### What happened?

Comes from nipreps/mriqc#1261.

When a filter is given (e.g.: `{'pet': {'session': '15'}}` but there's no query for `'pet'` as a suffix, then `collect_data` fails.

It would be easy to pick only the filters that apply to the queries.

### What command did you use?

```shell
mriqc
```

### What version of the software are you running?

master

### How are you running this software?

Local installation ("bare-metal")

### Is your data BIDS valid?

Yes

### Are you reusing any previously computed results?

No

### Please copy and paste any relevant log output.

_No response_

_No response_

asaraog:
Closes #862.

`collect_data` merges filters with `queries[acq].update(entities)` and never checks that `acq` is a key, so a filter naming a suffix absent from a caller-supplied `queries` raises instead of being ignored:

```python
collect_data(root, '01', bids_validate=False,
             queries={'t1w': {'datatype': 'anat', 'suffix': 'T1w'}},
             bids_filters={'pet': {'session': '15'}})
# KeyError: 'pet'
```

Two-line guard skips those filters. Test added using `generate_bids_skeleton`; it fails without the fix with the same `KeyError`. `pytest niworkflows/utils`: 73 passed, 5 skipped.

### DIFF
--- a/niworkflows/utils/bids.py
+++ b/niworkflows/utils/bids.py
@@ -271,6 +271,8 @@ def collect_data(
 
     bids_filters = bids_filters or {}
     for acq, entities in bids_filters.items():
+        if acq not in queries:  # filter with no matching query
+            continue
         # BIDS filters will not be able to override subject / session entities
         for entity, param in reserved_entities:
             if param == Query.OPTIONAL:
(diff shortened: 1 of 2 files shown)


effigies (approved):
Agreed. Thanks!

PR #1059, opened 2026-09-13
Space-separate workflow descriptions in visit_desc

### LINKED ISSUE #999: Space-separate descriptions from LiterateWorkflow.visit_desc()

Instead of joining the descriptions directly, space-separating them would be nice. I don't know how that would interact with multiline descriptions with newlines, but I think there's probably some way it could be handled (e.g., replacing " \n" with "\n).

Yes

_No response_

asaraog:
Closes #999. visit_desc joins descriptions with a space and strips them, so a description no longer needs a trailing space to stay separated.

### DIFF
--- a/niworkflows/utils/bids.py
+++ b/niworkflows/utils/bids.py
@@ -271,6 +271,8 @@ def collect_data(
 
     bids_filters = bids_filters or {}
     for acq, entities in bids_filters.items():
+        if acq not in queries:  # filter with no matching query
+            continue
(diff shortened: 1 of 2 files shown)


mgxd (commented):
Thanks! A few initial thoughts:

- `strip()` feels too aggressive, as devs may want to separate workflow descriptions by multiple lines, i.e. (`functional processing ... \n\n`)
- The collect data changes shouldn't be in this PR - please remove them

